PR #1, opened 2023-05-07
Implement Session Start (SS-54)

BrickPi:
Utilises GameManager's setup method to grab the init time for use as session ID for savegames, as well as setting up potential features in the future such as session time tracking for multiplayer.


Hop311 (approved):
Sim PR #1 ready to merge 👍

PR #9, opened 2023-07-21
Pops and other bits

Hop311:
Data structures and managers for:
- Pop, PopType
- Religion, ReligionGroup
- Culture, Culture Group, GraphicalCultureType

Provinces have an internal Pop list, currently filled with hardcoded test Pops. Provinces keep track of their total population, and the map keeps track of its total population and the highest individual province population (which will be used in the Population Density mapmode).

Requirements implemented:
- POP-18, POP-19, POP-20, POP-21, POP-26
- MAP-65

### DIFF
--- a/src/openvic/pop/Pop.cpp
+++ b/src/openvic/pop/Pop.cpp
@@ -1,0 +1,99 @@
+#include "Pop.hpp"
+
+#include <cassert>
+
+#include "../utility/Logger.hpp"
+
+using namespace OpenVic;
+
+Pop::Pop(PopType const& new_type, Religion const& new_religion)
+	: type { new_type }, religion { new_religion }, size { 1 } {
+	assert(size > 0);
+}
+
+PopType const& Pop::get_type() const {
+	return type;
+}
+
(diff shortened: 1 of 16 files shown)


src/openvic/pop/Pop.cpp
```diff
@@ -0,0 +1,107 @@
...
+	bool new_state_capital_only, bool new_demote_migrant, bool new_is_artisan)
+	: HasIdentifier { new_identifier },
+	  HasColour { new_colour, true },
+	  strata { new_strata },
+	  sprite { new_sprite },
+	  max_size { new_max_size },
+	  merge_max_size { new_merge_max_size },
+	  state_capital_only { new_state_capital_only },
+	  demote_migrant { new_demote_migrant },
+	  is_artisan { new_is_artisan } {
```
Juliusfrank11, line 45: If you are making a bool for artisans, it might be good to also make `is_slave` since they have a variety of unique (non) interactions with things like voting and income.

Spartan322 (approved):

PR #14, opened 2023-08-25
Dataloading scaffolding + basic culture and pop history loading

Hop311:
- Added [OpenVic-Dataloader](https://github.com/OpenVicProject/OpenVic-Dataloader) as a dependency (moved here from the main [OpenVic](https://github.com/OpenVicProject/OpenVic) repo where it was a dependency alongside this repo).
- Rewrote `SConstruct` based on `OpenVic-Dataloader`'s, which is itself invoked by the script to build the dataloader as a library and link it.
- Updated headless mode to execute as much of the game as is currently possible without the Godot layer (most defines loading apart from the map).
- Moved hardcoded defines from Godot layer to `GameManager` (mapmodes and building types).
- Added `Dataloader` manager class - accepts a list of paths (base install directory + mod directories, in order) with which to lookup files, checking the last added mod first and continuing back towards the base directory until it finds the required file. The `Dataloader` also contains defines loading functions that invoke the file lookup functions, pass the found filepaths to the `OpenVic-Dataloader` v2script parser, and pass the resulting AST onto the relevant manager class' specific data loading functions.
- Added `NodeTools` functions to assist with AST validation - in general, you pass a `NodeCPtr` to `NodeTools::expect_TYPE`, and if the node contains the correct data type/format it invokes a callback function `return_t(TYPE)`. List and dictionary functions can be more complex, e.g. running a list length callback first or using a key-callback map to direct dictionary parsing.
- Created `GoodCategory`, similar to `CultureGroup`, to replace the string previously used to represent good categories. Added `load_good_file` function stub, integrated into the `Dataloader` but currently only adding a single test good category and good (Typhion has started work on implementing this function).
- Moved `index_t = uint16_t` back into `Province`, as `uint16_t` isn't appropriate for a general global index type.
- Added religion distribution to provinces and a religion mapmode (currently doesn't do much as religion loading is unimplemented, but once that's done the mapmode will work automatically).
- Implemented `GraphicalCultureType`, `CultureGroup` and `Culture` loading, using the full range of `NodeTools` functions, fully integrated into the `Dataloader`.
- Added `load_pop_type_file` function stub, integrated into the `Dataloader` including passing the file name to the function (the `PopType` name comes from the file name, not the contents of the file). Currently only adding a single test pop type, task is available to any dev who asks me, although not all pop type defines data can be loaded at the moment due to reliance on other as-yet unimplemented dataloading stuff (goods loading required to load needs, military units needed for rebel army composition, condition parsing needed for migration, promotion, politics, etc.).
- Added basic pop history loading, currently loading pops' size and culture, which is then used to display the population density and (majority) culture mapmodes, as well as culture piecharts in the Province Overview Panel.
- Added `load_religion_file` function stub, integrated into the `Dataloader` but currently only adding a single test religion group and religion (Nyaf has started work on implementing this function).
- Broke up `Types.*pp` into smaller files: `Colour.hpp`, `IdentifierRegistry.*pp`, `Return.hpp`.
- Changed `IdentifierRegistry`'s `get_item_count` to `size` in order to match `std::vector`, and added `reserve` function, so `obj.reserve(obj.size() + val)` makes sense when `obj` is an `IdentifierRegistry` or `std::vector`.
- Added FakeByte's fixed point implementation, with string parsing functions rewritten to accept string start+end or start+length, instead of relying on C-style null-terminated strings. `Date`'s string parsing function has been rewritten similarly.
- Added `StringUtils.hpp` and `NumberUtils.hpp`, both used in FakeByte's fixed point code, with the `string_to_[u]int64` parsing functions rewritten to avoid relying on C-style null-terminated strings.


Spartan322 (dismissed):
I can only really comment on the dataloader stuff, I can probably look into refining this, but it should work for now. I honestly wouldn't recommend merging this just at the second though, probably should focus on updating things with the dataloader first.

CptAlanSmith:
return_t should be changed to bool - an acknowledgement that bool = true means a failure state can be communicated or commented but it's not an issue to the point of copying a bool and calling it something else.

Strong reservations about culture.cpp CultureManager::load_culture_file method, which should have its scope _heavily_ reduced to something sensible with a view to using more external method calls to perform calculations or other types of computation.
There are many issues which compound to create a perfect storm, here are some solutions which I would expect you to utilise:

1. Comment more on what is going on and why
2. Break out complex parameters for method calls into separate sections for calculation before the call - not inside a method call
3. Significantly reduce scope - nesting of 3 levels or less is recommended
4. Reduce the complexity of the code - multiple if statements taking up 100 lines that are easy to understand is superior to 3 lines of code that handles it all but can't be read

Whether you accept and integrate these comments or not, I expect a junior dev to be able to read and understand the method in less than 2 minutes or the method is not fit for purpose.

CptAlanSmith (changes requested):
https://github.com/OpenVicProject/OpenVic-Simulation/pull/14#issuecomment-1704440281

See my comment^

CptAlanSmith (approved):

CptAlanSmith:
Happy to see feedback taken into account, thanks.

src/openvic-simulation/GameManager.cpp
```diff
@@ -0,0 +1,132 @@
...
+	needs_update = true;
+}
+
+void GameManager::update_state() {
+	if (needs_update) {
+		Logger::info("Update: ", today);
+		map.update_state(today);
+		if (state_updated) state_updated();
+		needs_update = false;
+	}
```
Spartan322, line 21: Probably should prefer:
```suggestion
	if (needs_update) return;
	Logger::info("Update: ", today);
	map.update_state(today);
	if (state_updated) state_updated();
	needs_update = false;
```

src/openvic-simulation/GameManager.cpp
```diff
@@ -0,0 +1,132 @@
...
+	return province->expand_building(building_type_identifier);
+}
+
+bool GameManager::load_hardcoded_defines() {
+	bool ret = true;
+
+	static constexpr colour_t LOW_ALPHA_VALUE = float_to_alpha_value(0.4f);
+	static constexpr colour_t HIGH_ALPHA_VALUE = float_to_alpha_value(0.7f);
+	using mapmode_t = std::pair<std::string, Mapmode::colour_func_t>;
+	const std::vector<mapmode_t> mapmodes {
```
Spartan322, line 64: This isn't really an issue, but this is probably another good place to suggest the incorporating an ordered map of contiguous memory, specifically ordered on insertion order.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -0,0 +1,288 @@
...
+using namespace OpenVic::NodeTools;
+using namespace ovdl;
+
+bool Dataloader::set_roots(std::vector<std::filesystem::path> new_roots) {
+	if (!roots.empty()) {
+		Logger::error("Overriding existing dataloader roots!");
+		roots.clear();
+	}
+	bool ret = true;
+	for (std::reverse_iterator<std::vector<std::filesystem::path>::const_iterator> it = new_roots.crbegin(); it != new_roots.crend(); ++it) {
```
Spartan322, line 20: I do wonder if it might not be more efficient to give long STL types a shortened type alias in the namespace, so say `std::vector<std::filesystem::path>` becomes `path_vector` so this could be replaced with `std::reverse_iterator<path_vector::const_iterator>` if we'll avoid auto for this. (I'd probably foremost suggest path itself be made shortened to something like `fs_path` or `fs::path` anyway)

Spartan322 (dismissed):
Don't have much issue with this, might touch things up at a later date if I get around to it.

src/openvic-simulation/GameManager.cpp
Hop311: I assume you meant `if (!needs_update) return;`, and yeah your way is better, the overwhelming majority of calls to `update_state` will be while `needs_update` is false.

Thinking about similar cases throughout the code, is it worth using C++20's `[[likely]]` and `[[unlikely]]`, or should I just let the compiler handle it, writing the likely case first as a hint where possible?

src/openvic-simulation/GameManager.cpp
Hop311: These are "hardcoded defines", which ideally could be replaced with `txt` defines once we have a powerful enough scripting format. The vector is iterated over with the `std::string` and `Mapmode::colour_func_t` passed to `map.add_mapmode`, which validates the arguments, constructs a `Mapmode` (move-only, like its base-class `HasIdentifier`) and enters it into an `IdentifierRegistry<Mapmode>`. There, `Mapmode`s are stored in a `std::vector` to allow lookup by index, with a `std::map<std::string, size_t, std::less<void>>` used for lookup by identifier (the `std::less<void>` allows us to search by `std::string_view`).  Feel free to share any ideas for improvements,

src/openvic-simulation/dataloader/Dataloader.cpp
Hop311: Good ideas 👍

src/openvic-simulation/GameManager.cpp
Spartan322: `[[likely]]` and `[[unlikely]]` are optimizations for when the compiler can't see whats going on, the only reason to use it honestly is when you can know that the compiler doesn't actually know how to predict the branches well, so you try to tell it early how it should predict, iirc msvc may still completely ignore this because of course it does, it kinda requires benchmarks to know if this optimization is necessary, because if you're wrong you just costed the branch predictor immensely.

Spartan322:
Note any conversations you aren't doing now you can mark as resolved so this passes.

Hop311:
Changes since last reviews:
- incorporated feedback from Spartan on type and namespace aliases 
- added a `warning` logging function (alongside `info` and `error`) for invalid but recoverable input data (specifically mistakes in vanilla defines like duplicate province IDs in the list of water provinces, pops with size 0, random invalid pixels in the province and terrain images)
- Formatted hpp/cpp files and added `.clang-format` based on that of the main repo

Added req comments for:
- MAP-68, MAP-70, MAP-234
- POP-15, POP-16, POP-17, POP-22, POP-59 - POP-284, POP-286 - POP-299

Spartan322 (approved):
Looks fine enough, I don't think I'd call the fixed point `fixed_point_t`, especially when the header is `FixedPoint.hpp`, I'd assume that the header and main struct of the header would both be PascalCase, even in it being mostly a trivial integer proxy object.

Hop311:
I've been using `snake_case_t` for simple data storage structures, basically anything you could reasonably pass by value, so if I had to pick one I'd change the file name. There are a couple other types with different case from their file names, eg `Colour.hpp` vs `colour_t` and (until this PR) `Return.hpp` vs `return_t`, and I'm happy to have some group discussions about style and formatting, the `.clang-format` needs re-doing (at least for the main and sim repos) to have a column limit otherwise we're stuck with all-on-one-line or all-on-individual-lines for arguments and parameters.

PR #15, opened 2023-09-15
More dataloading

Hop311:
- Region and meta region loading
- Localisation loading
- First go at Modifiers (ModifierEffect, ModifierValue, Modifier, ModifierInstance and ModifierManager)


FarmingtonS9 (approved):
Approved

PR #22, opened 2023-09-22
Dataloading for military (Unit, LandUnity, NavalUnit)

zaaarf:
Should work, more or less. There's a couple TODOs, but I'm not sure we can implement them at this stage, so they'll probably have to stay there.

### DIFF
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -1,4 +1,5 @@
 #include "Dataloader.hpp"
+#include <filesystem>
 
 #include <openvic-dataloader/csv/Parser.hpp>
 #include <openvic-dataloader/detail/CallbackOStream.hpp>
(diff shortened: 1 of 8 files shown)


src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -150,13 +152,19 @@ bool Dataloader::_load_pop_types(PopManager& pop_manager, fs::path const& pop_ty
 			return pop_manager.load_pop_type_file(file.stem().string(), _parse_defines(file).get_file_node());
 		}
 	);
-	if (!ret) {
-		Logger::error("Failed to load pop types!");
-	}
```
CptAlanSmith, line 155: Why remove this in a commit about military stuff?

CptAlanSmith (approved):
Successfully compiled without issues.
Minor comments on small stuff - no code changes asked for.

src/openvic-simulation/dataloader/Dataloader.cpp
zaaarf: Just something I noticed - that was being printed twice. And I've been told to avoid making commits for these small things, so I just added it since I was editing that file anyway.

Hop311:
Although the code compiles, when trying to run it I get a segfault. I can help fix this tomorrow and get it merged.

zaaarf:
Fixed all the errors, now runs and parses everything correctly!

Hop311 (approved):
Working on my machine 👍

PR #31, opened 2023-09-24
Fixing compile-time warnings

zaaarf:
Only warnings left for me are the `#pragma once` things (which joe should already have fixed in his branch). When we merge Testing, we'll finally have a warn-free compilation. The built binary seems to work fine. 

Note: this applies to linux with `gcc`, haven't tested it on other systems/compilers.

### DIFF
--- a/src/openvic-simulation/Modifier.cpp
+++ b/src/openvic-simulation/Modifier.cpp
@@ -71,19 +71,14 @@ ModifierValue ModifierValue::operator-(ModifierValue const& right) const {
 	return ret -= right;
 }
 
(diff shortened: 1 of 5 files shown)


Spartan322:
If this is already fixed in joe's branch, won't this change just introduce needless merge conflicts?

zaaarf:
> If this is already fixed in joe's branch, won't this change just introduce needless merge conflicts?

Joe only fixed the "`#pragma once` in main file" warnings, which are *not* fixed by this. This one fixes all the other warnings, sorry if that was phrased poorly.

Spartan322 (approved):

PR #33, opened 2023-09-25
Terrain types

Hop311:
- Moved terrain bmp loading to this repo
- `terrain.txt` loading + parsing
- Terrain type modifier calculation based on number of pixels of each type in the province
- Added simple terrain mapmode

### DIFF
--- a/src/openvic-simulation/GameManager.cpp
+++ b/src/openvic-simulation/GameManager.cpp
@@ -132,16 +132,23 @@ bool GameManager::expand_building(Province::index_t province_index, const std::s
 	return province->expand_building(building_type_identifier);
 }
 
+static constexpr colour_t LOW_ALPHA_VALUE = float_to_alpha_value(0.4f);
+static constexpr colour_t HIGH_ALPHA_VALUE = float_to_alpha_value(0.7f);
+static constexpr colour_t LAND_COLOUR = 0x0D7017;
+static constexpr colour_t WATER_COLOUR = 0x4287F5;
+
+static colour_t default_colour(Province const& province) {
+	return LOW_ALPHA_VALUE | (province.get_water() ? WATER_COLOUR : LAND_COLOUR);
+}
+
 bool GameManager::load_hardcoded_defines() {
 	bool ret = true;
 
-	static constexpr colour_t LOW_ALPHA_VALUE = float_to_alpha_value(0.4f);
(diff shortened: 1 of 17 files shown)


src/openvic-simulation/GameManager.cpp
```diff
@@ -124,16 +132,23 @@ bool GameManager::expand_building(Province::index_t province_index, const std::s
 	return province->expand_building(building_type_identifier);
 }
 
+static constexpr colour_t LOW_ALPHA_VALUE = float_to_alpha_value(0.4f);
+static constexpr colour_t HIGH_ALPHA_VALUE = float_to_alpha_value(0.7f);
+static constexpr colour_t LAND_COLOUR = 0x0D7017;
+static constexpr colour_t WATER_COLOUR = 0x4287F5;
```
CptAlanSmith, line 138: Can you comment why these values please? I did google and found nothing obvious.

src/openvic-simulation/Modifier.cpp
```diff
@@ -138,4 +161,4 @@ namespace OpenVic { //so the compiler shuts up
 		}
 		return stream;
 	}
-}
\ No newline at end of file
+}
```
CptAlanSmith, line 164: Advise minor reduction in nesting if possible.

src/openvic-simulation/map/Map.cpp
```diff
@@ -504,6 +403,134 @@ bool Map::load_region_file(ast::NodeCPtr root) {
 	return ret;
 }
 
+static constexpr colour_t colour_at(uint8_t const* colour_data, int32_t idx) {
+	idx *= 3;
+	return (colour_data[idx + 2] << 16) | (colour_data[idx + 1] << 8) | colour_data[idx];
+}
+
```
CptAlanSmith, line 410: Please comment these numbers in use. Particularly idx *= 3;
It's not clear of the why, though I can see an 8 and 16 division of data. 24 bit color? I don't know.

CptAlanSmith (approved):

src/openvic-simulation/Modifier.cpp
Hop311: This function is going to be rewritten to use a better system than `default_callback`.

CptAlanSmith (approved):

PR #44, opened 2023-10-05
Implement Trait Loading

BrickPi:
Loads leader traits, tests passing.

### DIFF
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -587,6 +587,7 @@ bool Dataloader::load_defines(GameManager& game_manager) const {
 	static const fs::path units_directory = "units";
 
 	static const fs::path buildings_file = "common/buildings.txt";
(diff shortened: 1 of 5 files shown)


Spartan322 (changes requested):
For consistency with every other file name, LeaderTraits should probably be renamed to LeaderTrait.

Spartan322 (approved):

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -587,6 +587,7 @@ bool Dataloader::load_defines(GameManager& game_manager) const {
 	static const fs::path units_directory = "units";
 
 	static const fs::path buildings_file = "common/buildings.txt";
+	static const fs::path bookmark_file = "common/bookmarks.txt";
```
Spartan322, line 590: I feel like rearranging this is kinda pointless in this PR, like if we need to rearrange this stuff, we should prefer a PR focused on that, but whatever I guess

Spartan322 (approved):

PR #45, opened 2023-10-05
Skeleton Event Loading

BrickPi:
NodeTools based event loader. Completion depends on scripting support.


zaaarf (approved):
Tested and read through it. Looks good to me!

PR #46, opened 2023-10-09
Add Bookmark Loading

Spartan322:
Testing passes, bookmark identifiers are just the string of the date, might be better to use the name however I was unsure with it being a localization value, only other alternative is storing it sequentially.

### DIFF
--- a/src/openvic-simulation/history/Bookmark.cpp
+++ b/src/openvic-simulation/history/Bookmark.cpp
@@ -1,0 +1,95 @@
+#include "Bookmark.hpp"
+
+#include <openvic-dataloader/v2script/AbstractSyntaxTree.hpp>
+
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+
+#include "types/Date.hpp"
+#include "types/IdentifierRegistry.hpp"
+#include "utility/Logger.hpp"
+
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
+
+Bookmark::Bookmark(
+	std::string_view new_identifier,
+	std::string_view new_name,
+	std::string_view new_description,
+	Date new_date,
+	uint32_t new_initial_camera_x,
+	uint32_t new_initial_camera_y
+)
+	: HasIdentifier { new_identifier },
+	  name { new_name },
+	  description { new_description },
+	  date { new_date },
+	  initial_camera_x { new_initial_camera_x },
+	  initial_camera_y { new_initial_camera_y } {
+}
+
+std::string_view Bookmark::get_name() const {
+	return name;
+}
+
+std::string_view Bookmark::get_description() const {
+	return description;
+}
+
+Date const& Bookmark::get_date() const {
+	return date;
+}
+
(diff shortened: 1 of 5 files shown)


src/openvic-simulation/history/Bookmark.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+	struct Bookmark : HasIdentifier {
+		friend struct BookmarkManager;
+
+	private:
+		std::string name;
+		std::string description;
+		Date date;
+		uint32_t initial_camera_x;
+		uint32_t initial_camera_y;
```
Hop311, line 20: Can't these be `const`?

src/openvic-simulation/history/Bookmark.cpp
```diff
@@ -0,0 +1,95 @@
...
+	  initial_camera_y { new_initial_camera_y } {
+}
+
+std::string_view Bookmark::get_name() const {
+	return name;
+}
+
+std::string_view Bookmark::get_description() const {
+	return description;
+}
```
Hop311, line 36: Where possible, I've been returning `std::string const&` rather than `std::string_view`s - if a `std::string_view` is needed then it'll be converted automatically, we don't need to force that conversion now, especially if we ever need these values as proper `std::string`s or C strings.

src/openvic-simulation/history/Bookmark.cpp
```diff
@@ -0,0 +1,95 @@
...
+			std::string_view name, description;
+			Date date;
+			uint32_t initial_camera_x, initial_camera_y;
+
+			bool ret = expect_dictionary_keys(
+				"name", ONE_EXACTLY, expect_string(assign_variable_callback(name)),
+				"desc", ONE_EXACTLY, expect_string(assign_variable_callback(description)),
+				"date", ONE_EXACTLY, expect_date(assign_variable_callback(date)),
+				"cameraX", ONE_EXACTLY, expect_int(assign_variable_callback(initial_camera_x)),
+				"cameraY", ONE_EXACTLY, expect_int(assign_variable_callback(initial_camera_y))
```
Hop311, line 85:
```suggestion
				"cameraX", ONE_EXACTLY, expect_uint(assign_variable_callback_uint(initial_camera_x)),
				"cameraY", ONE_EXACTLY, expect_uint(assign_variable_callback_uint(initial_camera_y))
```
Change to unsigned versions + add the `_uint` to the callbacks to use versions that check the value's size before downcasting. These won't be necessary soon as my big PR includes changes that move the check+cast to the `expect_[u]int` functions, but for now they should still be used.

src/openvic-simulation/history/Bookmark.cpp
Spartan322: A const string reference can potentially allocate still, whereas string_views are guaranteed never to allocate, in every case where you know the string will outlive a call, you should be using string_view as it will never allocate.

BrickPi (approved):
Good work.

PR #50, opened 2023-10-11
Lots of accumulated changes

Hop311:
- Replaced `ALLOW_OTHER_KEYS` (and equivalent bool) arguments with default key-value callback arguments. This allowed me to change the "add size the decrement for every reserved key" with "increment for every non-reserved/default key" for finding the amount to reserve.
- Renamed `expect_[u]int` to `expect_[u]int64`, and added new `expect_[u]int` functions which check the value so that casts down from 64-bits are safe. Removed `assign_variable_callback_[u]int` because they're no longer needed.
- Added `expect_(years|months|days)` functions taking`Timespan` callbacks.
- Added `string_map_t` template for `std::string`-keyed map supporting heterogeneous lookup
- Added `expect_mapped_string` for simple string to value lookups, it returns a `std::string_view` callback so you can choose whether to `expect_identifier` or `expect_string`. Used for enum lookups.
- Simplified `IDENTIFIER_REGISTRY_ACCESSORS` so it doesn't need the type as an argument.
- Moved water province locking to inside `set_water_province_list`.
- Changed government `election_duration` to `term_duration`.
- Renamed `IdentifierRegistry` to `UniqueKeyRegistry` and added storage type and string and pointer getter functions as template parameters, in preparation for storing `std::unique_ptr`s instead of normal values for GUI registries which will store different derived classes alongside each other. Created template aliases `ValueRegistry` and `InstanceRegistry` for the standard value and `std::unique_ptr` storage cases, and `IdentifierRegistry` and `IdentifierInstanceRegistry` for `HasIdentifier` specialisations of those cases.
- Commented out where `Logger` outputs `function_name` to make output more readable.


BrickPi:
The changes to timespan months & years and respective nodetools are very appreciated.

zaaarf (approved):
Compiled with no warnings on Linux with `gcc`. Test was succesful.

PR #51, opened 2023-10-13
Dataloading tech

zaaarf:


zaaarf:
Finally got everything to load. There's still some things that need to be tweaked, but we're very close.

The most important (read: global) change in this PR is to the `ModifierManager`: `expect_modifier_value_and_keys` (and all similar methods) will now "flatten": `value = { subvalue = 0.5 }` will now be read the same as `value_subvalue = 0.5`. There should be no relevant drawbacks to this approach, but let me know if you figure out something else.

zaaarf:
Absolutely no clue of why it fails to build on Mac. Any ideas?

src/openvic-simulation/misc/Modifier.cpp
```diff
@@ -291,23 +303,35 @@ bool ModifierManager::load_triggered_modifiers(ast::NodeCPtr root) {
...
+			return add_modifier_cb(effect, value);
+		} else if (complex_modifiers.contains(key) && value->cast_to<ast::AbstractListNode>() != nullptr) {
+			return expect_dictionary([this, &key, &add_modifier_cb, &default_callback](std::string_view identifier, ast::NodeCPtr node) -> bool {
+				std::string flat_identifier = std::string(key);
+				flat_identifier += "_";
+				flat_identifier += identifier;
+				ModifierEffect const* effect = get_modifier_effect_by_identifier(flat_identifier);
+				if(effect != nullptr) {
+					return add_modifier_cb(effect, node);
+				} else return false;
```
Hop311, line 332: Could you add an error message here including the `flat_identifier` which couldn't be found?

src/openvic-simulation/misc/Modifier.cpp
```diff
@@ -291,23 +303,35 @@ bool ModifierManager::load_triggered_modifiers(ast::NodeCPtr root) {
...
+		} else {
+			Logger::error("Failed to validate modifier effect: ", effect->get_identifier());
+			return false;
 		}
-		return default_callback(key, value);
+	};
+
+	return [this, &modifier, default_callback, effect_validator, add_modifier_cb](std::string_view key, ast::NodeCPtr value) -> bool {
+		ModifierEffect const* effect = get_modifier_effect_by_identifier(key);
+		if (effect != nullptr && value->cast_to<ast::IdentifierNode>() != nullptr) {
```
Hop311, line 322:
```suggestion
		if (effect != nullptr && value->is_type<ast::IdentifierNode>()) {
```
Well done on finding such effective solutions to some very nasty problems, namely that some modifier effect keys have different uses in certain contexts. Here the problem is the list of `colonial_points` provided by a `naval_base` at each level, which causes an error since the list is a different node type from the normal modifier effect value node (Identifier with a decimal value). What's going to be trickier to deal with are the potential cases where the value node type is the same but there's still a special use, as those won't trigger an error but they will load incorrectly.

src/openvic-simulation/misc/Modifier.cpp
```diff
@@ -291,23 +303,35 @@ bool ModifierManager::load_triggered_modifiers(ast::NodeCPtr root) {
...
+			return add_modifier_cb(effect, value);
+		} else if (complex_modifiers.contains(key) && value->cast_to<ast::AbstractListNode>() != nullptr) {
```
Hop311, line 324:
```suggestion
		} else if (complex_modifiers.contains(key) && value->is_derived_from<ast::AbstractListNode>()) {
```
Again, clever solution to a nasty problem, this time `artillery` being used as a good to be encouraged by a national focus, and not as the key for a list of modifiers to the military unit of the same name.

Eventually (in a future PR, this one is massive enough as is 😁), we're going to need to design a proper system for handling these multiple use keys. Some ideas:
- similar to `default_callback`, have a function that's called at the beginning and only parse modifier effects if that function doesn't accept the key-value pair (the function would need to return 3 values, success/failure for keys it accepts and a third value for keys it rejects that need to be parsed as modifier effects).
- change how effect validators are used so that modifier keys they reject go to `default_callback`, instead of being thrown away as "failed keys". This requires less code change but more work to actually use, as it requires listing every single allowed modifier effect while the first idea just moves some of the already existing default parsing code to the beginning of the process.

Again, let's leave this for now, it needs some serious brainstorming and research into which modifier effect keys also have special uses.

Hop311 (commented):
I've only had time to review up to the Technology files, I'll do them tomorrow after work.


Massive congratualtions for the amazing work you've done with modifiers in this PR, I'm delighted to see how confident you are with C++ and the OpenVic codebase and I look forward to working through the rest of this project with you 🙂

src/openvic-simulation/military/Unit.cpp
```diff
@@ -292,64 +292,114 @@ bool UnitManager::load_unit_file(GoodManager const& good_manager, ast::NodeCPtr
...
+				STAT_MODIFIER("defence", true, RAW_DECIMAL);
+				STAT_MODIFIER("discipline", true, RAW_DECIMAL);
+				STAT_MODIFIER("support", true, RAW_DECIMAL);
+				STAT_MODIFIER("maneuver", true, RAW_DECIMAL);
+				STAT_MODIFIER("siege", true, RAW_DECIMAL);
+				break;
+			}
+            case Unit::type_t::NAVAL: {
+				STAT_MODIFIER("colonial_points", true, INT);
+				STAT_MODIFIER("supply_consumption_score", false, PROPORTION_DECIMAL); //idk
```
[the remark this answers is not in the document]
zaaarf: Thanks so much for that, I honestly had no idea of what this was even supposed to do.

zaaarf:
Thank you for the kind words and the effort in reviewing this. I've applied all the stuff you pointed out in this commit., but I've left the requested change where @schombert intervened open so you don't miss it.

BrickPi (changes requested):
Should merge the first three commits into a single technology loading one, merge the last three into a single modifier commit. Will review content when I get a chance.

src/openvic-simulation/misc/Modifier.hpp
```diff
@@ -107,6 +109,7 @@ namespace OpenVic {
 	private:
 		IdentifierInstanceRegistry<ModifierEffect> modifier_effects;
 		IdentifierRegistry<Modifier> event_modifiers;
+		std::unordered_set<std::string_view> complex_modifiers;
```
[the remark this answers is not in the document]
Hop311: Oh nice, it's got an example of how to wrap up `std::hash<std::string_view>` might include this in a separate PR, although I've also been encouraged to use a separate set/map library.

src/openvic-simulation/tech/Technology.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+bool TechnologyManager::load_technologies_file(ModifierManager const& modifier_manager, UnitManager const& unit_manager, BuildingManager const& building_manager, ast::NodeCPtr root) {
+	size_t expected_technologies = 0;
+	bool ret = expect_dictionary_reserve_length(technologies, [this, &expected_technologies](std::string_view key, ast::NodeCPtr value) -> bool {
+		return expect_length(add_variable_callback(expected_technologies))(value);
+	})(root);
+	technologies.reserve(technologies.size() + expected_technologies);
+
+	ret &= expect_dictionary([this, &modifier_manager, &unit_manager, &building_manager](std::string_view tech_key, ast::NodeCPtr tech_value) -> bool {
```
Hop311, line 107:
```suggestion
	return expect_dictionary_reserve_length(technologies, [this, &modifier_manager, &unit_manager, &building_manager](std::string_view tech_key, ast::NodeCPtr tech_value) -> bool {
```
The `expected_*` counting pattern is used to count things spread across multiple parent dictionaries, e.g. count cultures from every culture group. In this case, all the techs are in the top level dictionary of the file (`root`), so we can just reserve its length normally in order to reserve enough space for every tech in the file.
(Also remove `return ret;` at the bottom of this function, now that we only have a single result to return.)

src/openvic-simulation/tech/Technology.cpp
```diff
@@ -0,0 +1,154 @@
...
+	bool ret = true;
+
+	ret &= modifier_manager.add_modifier_effect("unciv_military_modifier", true, ModifierEffect::format_t::PROPORTION_DECIMAL);
+	ret &= modifier_manager.add_modifier_effect("unciv_economic_modifier", true, ModifierEffect::format_t::PROPORTION_DECIMAL);
+
+	for (TechnologyFolder const& folder : get_technology_folders()) {
+		std::string folder_name = std::string(folder.get_identifier()) + "_research_bonus";
+		ret &= modifier_manager.add_modifier_effect(folder_name, true, ModifierEffect::format_t::PROPORTION_DECIMAL);
+	}
+	return true;
```
Hop311, line 153:
```suggestion
	return ret;
```

PR #56, opened 2023-10-16
Change some NodeTools functions to return Callback concepts

Spartan322:


src/openvic-simulation/dataloader/NodeTools.hpp
```diff
@@ -18,6 +22,28 @@ namespace OpenVic {
...
+		concept Callback = Functor<Fn, bool, Args...>;
+
+		template<typename Fn, typename ...Args>
+		concept NodeCallback = Callback<Fn, ast::NodeCPtr>;
+
+		template<typename Fn, typename ...Args>
+		concept KeyValueCallback = Callback<Fn, std::string_view, ast::NodeCPtr>;
+
+		template<typename Fn, typename ...Args>
+		concept LengthCallback = Functor<Fn, std::size_t, std::size_t>;
```
Hop311, line 45: Do these still need the `Args` template parameter?

src/openvic-simulation/dataloader/NodeTools.hpp
```diff
@@ -197,23 +223,39 @@ namespace OpenVic {
...
+		template<std::integral T>
+		callback_t<T> assign_variable_callback_cast(auto& var) {
+			return [&var](T val) -> bool {
+				var = val;
+				return true;
+			};
+		}
+
+		template<std::signed_integral T>
+		callback_t<T> assign_variable_callback_cast(auto& var) {
```
Hop311, line 243: What are these specialisations for?

Hop311 (approved):

src/openvic-simulation/dataloader/NodeTools.hpp
Spartan322: Because of how these work, without this expec_int and expect_uint can't deduce the type you pass in, unfortunately since bool is an integral, this is actually a bit greedy.

src/openvic-simulation/dataloader/NodeTools.hpp
Spartan322: Yeah it really doesn't

Hop311 (approved):

PR #60, opened 2023-10-19
Country History & Order of Battle Loading

BrickPi:
Loads country histories and Order of Battle files, tests passing.

### DIFF
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -22,6 +22,7 @@
 
 #ifdef _WIN32
 #include <Windows.h>
+
 #include "Dataloader_Windows.hpp"
 #endif
 
@@ -90,8 +91,8 @@ static fs::path _search_for_game_path(fs::path hint_path = {}) {
 		if (!registry_path.empty())
 			return registry_path;
 
-		#pragma warning(push)
-		#pragma warning(disable: 4996)
+#pragma warning(push)
+#pragma warning(disable : 4996)
 		static const fs::path prog_files = std::string(std::getenv("ProgramFiles"));
 		hint_path = prog_files / "Steam";
 		if (!fs::is_directory(hint_path, error_code)) {
@@ -102,7 +103,7 @@ static fs::path _search_for_game_path(fs::path hint_path = {}) {
 				return "";
 			}
 		}
-		#pragma warning(pop)
+#pragma warning(pop)
 // Cannot support Android
 // Only FreeBSD currently unofficially supports emulating Linux
 #elif (defined(__linux__) && !defined(__ANDROID__)) || defined(__FreeBSD__)
@@ -278,7 +279,7 @@ static fs::path _search_for_game_path(fs::path hint_path = {}) {
 		auto parser = lexy_vdf::Parser::from_file(vic2_steam_lib_directory);
 		if (!parser.parse()) {
 			// Could not find or load appmanifest_42960.acf, report error as warning
-			for(auto& error : parser.get_errors()) {
+			for (auto& error : parser.get_errors()) {
 				Logger::warning(error.message);
 			}
 			return "";
@@ -458,30 +459,56 @@ csv::Windows1252Parser Dataloader::parse_csv(fs::path const& path) {
 }
 
 bool Dataloader::_load_pop_types(PopManager& pop_manager, UnitManager const& unit_manager, GoodManager const& good_manager, fs::path const& pop_type_directory) const {
-	const bool ret = apply_to_files_in_dir(pop_type_directory, ".txt",
-		[&pop_manager, &unit_manager, &good_manager](fs::path const& file) -> bool {
-			return pop_manager.load_pop_type_file(file.stem().string(), unit_manager, good_manager, parse_defines(file).get_file_node());
-		}
-	);
+	const bool ret = apply_to_files_in_dir(pop_type_directory, ".txt", [&pop_manager, &unit_manager, &good_manager](fs::path const& file) -> bool {
+		return pop_manager.load_pop_type_file(file.stem().string(), unit_manager, good_manager, parse_defines(file).get_file_node());
+	});
 	pop_manager.lock_pop_types();
 	return ret;
 }
 
 bool Dataloader::_load_units(UnitManager& unit_manager, GoodManager const& good_manager, fs::path const& units_directory) const {
-	const bool ret = apply_to_files_in_dir(units_directory, ".txt",
-		[&unit_manager, &good_manager](fs::path const& file) -> bool {
-			return unit_manager.load_unit_file(good_manager, parse_defines(file).get_file_node());
-		}
-	);
+	const bool ret = apply_to_files_in_dir(units_directory, ".txt", [&unit_manager, &good_manager](fs::path const& file) -> bool {
+		return unit_manager.load_unit_file(good_manager, parse_defines(file).get_file_node());
+	});
 	unit_manager.lock_units();
 	return ret;
 }
 
-bool Dataloader::_load_countries(GameManager& game_manager, fs::path const& countries_file, ast::NodeCPtr root) const {
+bool Dataloader::_load_oobs(GameManager& game_manager) const {
+	static const fs::path oob_directory = "history/units";
+
+	/* used for countries with no defined initial OOB */
+	game_manager.get_military_manager().get_deployment_manager().add_deployment("NULL", std::vector<Army>(), std::vector<Navy>(), std::vector<Leader>());
+
+	bool ret = apply_to_files_in_dir(oob_directory, ".txt", [&game_manager](fs::path const& file) -> bool {
+		if (file.filename() == "v2dd2.txt") return true; /* dev diary just stuck in there for no reason whatsoever */
+		return game_manager.get_military_manager().get_deployment_manager().load_oob_file(game_manager, file.filename().string(), parse_defines(file).get_file_node());
+	});
+
+	/* we also load OOBs in top level subdirectories, for other start dates etc */
+	for (auto root : roots) {
+		const fs::path path = root / oob_directory;
+		std::error_code ec;
+		for (fs::directory_entry const& entry : fs::directory_iterator { path, ec }) {
+			if (entry.is_directory()) {
+				ret &= apply_to_files_in_dir(entry, ".txt", [&entry, &game_manager](fs::path const& file) -> bool {
+					return game_manager.get_military_manager().get_deployment_manager().load_oob_file(game_manager, (entry.path().filename() / file.filename()).string(), parse_defines(file).get_file_node());
+				});
+			}
+		}
+	}
+
+	game_manager.get_military_manager().get_deployment_manager().lock_deployments();
+	return ret;
+}
+
+bool Dataloader::_load_countries(GameManager& game_manager) const {
+	static const fs::path countries_file = "common/countries.txt";
+
 	bool is_dynamic = false;
 
 	bool ret = expect_dictionary(
-		[this, &game_manager, &is_dynamic, &countries_file](std::string_view key, ast::NodeCPtr value) -> bool {
+		[this, &game_manager, &is_dynamic](std::string_view key, ast::NodeCPtr value) -> bool {
 			if (key == "dynamic_tags") {
 				return expect_bool(assign_variable_callback(is_dynamic))(value);
 			}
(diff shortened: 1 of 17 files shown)


src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+const std::vector<Reform const*>& CountryHistory::get_reforms() const {
+	return reforms;
+}
+
+OOB const* CountryHistory::get_inital_oob() const {
+	return inital_oob;
+}
+
+#define EXISTING_ENTRY country_histories.at(country).at(date)
```
Spartan322, line 81: Should really undef this so its limiting its scope, also since its only ever used for `add_country_history_entry`, should really keep it inside that function's block, especially since it uses the arguments for it, it makes it easier to see what its relevant to.

Though I have to wonder why not just save the iterator once and reference the iterator instead?

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -458,30 +459,56 @@ csv::Windows1252Parser Dataloader::parse_csv(fs::path const& path) {
...
 	unit_manager.lock_units();
 	return ret;
 }
 
-bool Dataloader::_load_countries(GameManager& game_manager, fs::path const& countries_file, ast::NodeCPtr root) const {
+bool Dataloader::_load_oobs(GameManager& game_manager) const {
+	static const fs::path oob_directory = "history/units";
+
+	/* used for countries with no defined initial OOB */
+	game_manager.get_military_manager().get_order_of_battle_manager().add_oob("NULL", std::move(std::vector<Army>()), std::move(std::vector<Navy>()), std::move(std::vector<Leader>()));
```
Spartan322, line 481: std::move here is completely superfluous here, construction for rvalues automatically makes them an rvalue, move should only really be useful for moving an lvalue, usually an already constructed temporary, into an rvalue.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -521,9 +566,11 @@ bool Dataloader::_load_map_dir(GameManager& game_manager, fs::path const& map_di
...
 
 #define APPLY_TO_MAP_PATHS(F) \
-	F(definitions) F(provinces) F(positions) F(terrain) F(rivers) \
-	F(terrain_definition) F(tree_definition) F(continent) F(adjacencies) \
-	F(region) F(region_sea) F(province_flag_sprite)
+	F(definitions) \
+	F(provinces) \
+	F(positions) F(terrain) F(rivers) \
+		F(terrain_definition) F(tree_definition) F(continent) F(adjacencies) \
+			F(region) F(region_sea) F(province_flag_sprite)
```
Spartan322, line 573: If we're gonna format this like that, might as well format them on each line if clang-format will respect that.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -546,13 +593,14 @@ bool Dataloader::_load_map_dir(GameManager& game_manager, fs::path const& map_di
...
-		#X, ONE_EXACTLY, expect_string(assign_variable_callback(X)),
+	#X, ONE_EXACTLY, expect_string(assign_variable_callback(X)),
 		APPLY_TO_MAP_PATHS(MAP_PATH_DICT_ENTRY)
 #undef MAP_PATH_DICT_ENTRY
 
 #undef APPLY_TO_MAP_PATHS
 
-		"border_heights", ZERO_OR_ONE, success_callback,
+			"border_heights",
+		ZERO_OR_ONE, success_callback,
```
Spartan322, line 603: Why is this being moved to two lines by the formatter?

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -0,0 +1,308 @@
...
+			plurality == -1 ?				EXISTING_ENTRY.plurality :			plurality,
+			national_value == nullptr ?		EXISTING_ENTRY.national_value :		national_value,
+			civilised == false ?			EXISTING_ENTRY.civilised :			civilised,
+			prestige == -1 ?				EXISTING_ENTRY.prestige :			prestige,
+			!updated_reforms ?				EXISTING_ENTRY.reforms :			std::move(reforms),
+			initial_oob == nullptr ?		EXISTING_ENTRY.inital_oob :			initial_oob
+		};
+
+		country_histories.at(country).erase(date);
+		country_histories.at(country).emplace(date, entry);
```
Spartan322, line 132: Why not: (I'm using `auto` to shorthand this, use the actual iterator types, even if you must use `decltype`)
```suggestion
	auto country_history = country_histories[country];
	auto date_history = country_history.find(date);

	/* combine duplicate histories, priority to current (defined later) */
	if (date_history != country_history.end()) {
		CountryHistory entry = CountryHistory {
			primary_culture == nullptr ? 	date_history->primary_culture : 	primary_culture,
			!updated_accepted_cultures ? 	date_history->accepted_cultures : 	std::move(accepted_cultures),
			religion == nullptr ? 			date_history->religion : 			religion,
			ruling_party == nullptr ? 		date_history->ruling_party : 		ruling_party,
			last_election == Date(0) ? 		date_history->last_election : 		last_election,
			!updated_upper_house ? 			date_history->upper_house : 		std::move(upper_house),
			capital == nullptr ? 			date_history->capital : 			capital,
			government_type == nullptr ?	date_history->government_type : 	government_type,
			plurality == -1 ?				date_history->plurality :			plurality,
			national_value == nullptr ?		date_history->national_value :		national_value,
			civilised == false ?			date_history->civilised :			civilised,
			prestige == -1 ?				date_history->prestige :			prestige,
			!updated_reforms ?				date_history->reforms :			std::move(reforms),
			initial_oob == nullptr ?		date_history->inital_oob :			initial_oob
		};

		country_history.erase(date);
		country_history.emplace(date, entry);
```

Also this is a really valid case why const'ing everything isn't always a great idea, if we're gonna be modifying an existing object, all this does is create useless construction, in which case it be more efficient not to const these values, which given we already make them private variables anyway never really made sense to me.

Spartan322 (changes requested):
There's a lot of cases where you access the country of the history multiple times when the key doesn't change, (least not without a race condition I guess, but nothing in the loader would stand up to a race condition anyway) I would seriously recommend storing the key as a variable so you don't continuously access the map for no reason.

src/openvic-simulation/dataloader/Dataloader.cpp
BrickPi: Unsure, fixed it in the file. Not too familiar with clang-format.

Spartan322 (changes requested):
Honestly would also prefer OOB be renamed to OrderOfBattle, I don't really get the acronym. If only we had a better name tbh.

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -0,0 +1,300 @@
...
+		if (capital != nullptr) existing_entry->second.capital = capital;
+		if (government_type != nullptr) existing_entry->second.government_type = government_type;
+		if (plurality >= 0) existing_entry->second.plurality = plurality;
+		if (national_value != nullptr) existing_entry->second.national_value = national_value;
+		if (civilised) existing_entry->second.civilised = true;
+		if (prestige >= 0) existing_entry->second.prestige = prestige;
+		if (updated_reforms) existing_entry->second.reforms = reforms;
+		if (initial_oob != nullptr) existing_entry->second.inital_oob = initial_oob;
+	} else {
+		country_histories.at(country).emplace(date, CountryHistory {
```
Spartan322, line 126: Need to replace at here too.

Spartan322 (changes requested):
Exceptions are a pain to deal with for plenty of reasons and we pretty much already disable them on build, so they'll just crash the program without much info, we should avoid them wherever possible.

src/openvic-simulation/military/Unit.cpp
```diff
@@ -263,90 +263,41 @@ bool UnitManager::load_unit_file(GoodManager const& good_manager, ast::NodeCPtr
...
-					"maneuver", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(maneuver)),
-					"siege", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(siege))
-				);
+				ret &= add_key_map_entries(key_map, "primary_culture", ZERO_OR_ONE, expect_bool(assign_variable_callback(primary_culture)), "sprite_override", ZERO_OR_ONE, expect_identifier(assign_variable_callback(sprite_override)), "sprite_mount", ZERO_OR_ONE, expect_identifier(assign_variable_callback(sprite_mount)), "sprite_mount_attach_node", ZERO_OR_ONE, expect_identifier(assign_variable_callback(sprite_mount_attach_node)), "reconnaissance", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(reconnaissance)), "attack", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(attack)), "defence", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(defence)), "discipline", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(discipline)), "support", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(support)), "maneuver", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(maneuver)), "siege", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(siege)));
```
Spartan322, line 275: Could this be reverted please, this being one line is absolutely a nightmare to read.

src/openvic-simulation/military/Unit.cpp
```diff
@@ -263,90 +263,41 @@ bool UnitManager::load_unit_file(GoodManager const& good_manager, ast::NodeCPtr
...
...ariable_callback(active)), "unit_type", ONE_EXACTLY, expect_identifier(assign_variable_callback(unit_type)), "floating_flag", ONE_EXACTLY, expect_bool(assign_variable_callback(floating_flag)), "priority", ONE_EXACTLY, expect_uint(assign_variable_callback(priority)), "max_strength", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(max_strength)), "default_organisation", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(default_organisation)), "maximum_speed", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(maximum_speed)), "weighted_value", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(weighted_value)), "move_sound", ZERO_OR_ONE, expect_identifier(assign_variable_callback(move_sound)), "select_sound", ZERO_OR_ONE, expect_identifier(assign_variable_callback(select_sound)), "build_time", ONE_EXACTLY, expect_days(assign_variable_callback(build_time)), "build_cost", ONE_EXACTLY, good_manager.expect_good_decimal_map(move_variable_callback(build_cost)), "supply_consumption", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(supply_consumption)), "supply_cost", ONE_EXACTLY, good_manager.expect_good_decimal_map(move_variable_callback(supply_cost)));
```
Spartan322, line 267: Could this be reverted please, this being one line is absolutely a nightmare to read.

src/openvic-simulation/military/Unit.cpp
```diff
@@ -263,90 +263,41 @@ bool UnitManager::load_unit_file(GoodManager const& good_manager, ast::NodeCPtr
...
..._key_map_entries(key_map, "naval_icon", ONE_EXACTLY, expect_uint(assign_variable_callback(naval_icon)), "sail", ZERO_OR_ONE, expect_bool(assign_variable_callback(sail)), "transport", ZERO_OR_ONE, expect_bool(assign_variable_callback(transport)), "capital", ZERO_OR_ONE, expect_bool(assign_variable_callback(capital)), "colonial_points", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(colonial_points)), "can_build_overseas", ZERO_OR_ONE, expect_bool(assign_variable_callback(build_overseas)), "min_port_level", ONE_EXACTLY, expect_uint(assign_variable_callback(min_port_level)), "limit_per_port", ONE_EXACTLY, expect_int(assign_variable_callback(limit_per_port)), "supply_consumption_score", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(supply_consumption_score)), "hull", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(hull)), "gun_power", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(gun_power)), "fire_range", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(fire_range)), "evasion", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(evasion)), "torpedo_attack", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(torpedo_attack)));
```
Spartan322, line 290: Could this be reverted please, this being one line is absolutely a nightmare to read.

Spartan322 (changes requested):

Spartan322 (dismissed):

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -0,0 +1,315 @@
...
+					return false;
+				}
+
+				reforms.push_back(reform);
+				return ret;
+			}
+			// TODO: technologies & inventions
+			return true;
+		},
+		/* we have to use a lambda, assign_variable_callback_pointer apparently doesn't play nice with const & non-const accessors */
```
Hop311, line 224: Ah ok yeah `Map::expect_province_identifier` is the only instance of that type of function with a non-const version as well as the usual const one, so the compiler was having difficulty telling which to use, weird. I'll look into it, maybe change the non-const one's name.

Hop311 (approved):
One small typo in a function that's not currently used, but still better to fix it before merging.
Country history code is great, really nice to see how easy it is to parse so many of the loaded data types, e.g. in `CountryHistory.cpp` there's `Province`, `Culture`, `Religion`, `GovernmentType`, and `NationalValue` loading all in a row.
Various lambda functions need some re-formatting because `.clang-format` mangled them but let's merge the changes and use a fixed `.clang-format` to unpack them again automatically.

Hop311 (approved):

PR #61, opened 2023-10-19
Add Defines Loading

Spartan322:
Caches start_date and end_date in DefineManager

Add static `get_property` and HASID_PROPERTY macro to HasIdentifier

### DIFF
--- a/src/openvic-simulation/misc/Define.cpp
+++ b/src/openvic-simulation/misc/Define.cpp
@@ -1,0 +1,121 @@
+
+#include "Define.hpp"
+
+#include <cassert>
+#include <cstdlib>
+#include <memory>
+
+#include <openvic-dataloader/v2script/AbstractSyntaxTree.hpp>
+
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "openvic-simulation/types/Date.hpp"
+#include "openvic-simulation/types/IdentifierRegistry.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
+
+Define::Define(
+	std::string_view new_identifier,
+	std::string&& new_value,
+	Type new_type
+) : HasIdentifier { new_identifier },
+	value { std::move(new_value) },
+	type { new_type } {
+}
+
+fixed_point_t Define::get_value_as_fp() const {
+	return fixed_point_t::parse(value);
+}
+
+int64_t Define::get_value_as_int() const {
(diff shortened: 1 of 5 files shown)


src/openvic-simulation/dataloader/NodeTools.cpp
```diff
@@ -142,8 +142,8 @@ node_callback_t NodeTools::expect_colour(callback_t<colour_t> callback) {
 	};
 }
 
-node_callback_t NodeTools::expect_date(callback_t<Date> callback) {
-	return expect_identifier(
+node_callback_t NodeTools::expect_date(callback_t<Date> callback, string_conversion_callback_t parse_callback) {
```
BrickPi, line 145: Solved this problem differently with OOBs, just changed `expect_date` to use `expect_identifier_or_string` which feels more concise to use. This method works too but we need to decide on one to use.

src/openvic-simulation/misc/Define.cpp
```diff
@@ -0,0 +1,120 @@
...
+						case 'p': // pops
+							type = Pops;
+							break;
+						case 'a': // ai
+							type = Ai;
+							break;
+						case 'g': // graphics
+							type = Graphics;
+							break;
+						default: return false;
```
BrickPi, line 100: Would log an error message with the invalid identifier here.

BrickPi (changes requested):
Looks good generally, `HASID_PROPERTY` is a big clarity improvement & start_date loading is appreciated.

src/openvic-simulation/dataloader/NodeTools.cpp
Spartan322: Could change the default to be `expect identifier_or_string` and then introduce the parse_callback for any additional behavior instead, that's probably the best solution.

BrickPi: Sounds good to me

Hop311: The approach I've used for things like this is to return ‘std::string_view‘ callback, so you can then choose to do ‘expect_identifier(expect_date(...))‘ or ‘expect_string(expect_date(...))‘, or any other function which accepts a string_view callback. This also removes the need for a Node as the root source of the date

Hop311: Thinking about it some more, every expect function which is basically just # a specialisation of ‘expect_identifier‘ should return a string_view callback, eg expect_fixed_point.

BrickPi (approved):

PR #64, opened 2023-10-22
ModifierEffects stored as instances

Hop311:
- Changed `ModifierEffect`s to be stored in `unique_ptr`s, so that pointers and references to them are stable without the need for locking. This allows us to add effects mid-load, even after having already parsed other modifiers.
- Added `PROPORTION_DECIMAL` as an effect format representing unscaled fractions, e.g. 0.01 or -0.2 (corresponding to +1% and -20% respectively). This used to be represented by `PERCENTAGE_DECIMAL`, which now instead represents fractions multiplied by 100 to be actual percentages, e.g. 15 or -40 (corresponding to +15% and -40% respectively).
- Removed hardcoded `max_***` and `min_build_***` building modifier effects, instead generating them for each building loaded into the `BuildingManager`.

- Fixed some TGC compatibility issues in country loading, one failure remains while I work out how best to address it (a TGC placeholder policy `no_position_set` is defined as a `social_policy` but used for other types in `countries/Zombie Wasteland.txt`).

- Changed the return type form of ‘Date‘ from a const ref to a copied value, since it's really just an ‘int64_t‘ under the hood.


src/openvic-simulation/Modifier.cpp
```diff
@@ -103,7 +103,7 @@ bool ModifierManager::add_modifier_effect(std::string_view identifier, bool posi
 		Logger::error("Invalid modifier effect identifier - empty!");
 		return false;
 	}
-	return modifier_effects.add_item({ identifier, positive_good, format });
+	return modifier_effects.add_item(std::unique_ptr<ModifierEffect> { new ModifierEffect { identifier, positive_good, format } });
```
schombert, line 106: it is a bit cleaner to use `std::make_unique` for this

src/openvic-simulation/Modifier.cpp
```diff
@@ -191,12 +193,13 @@ bool ModifierManager::setup_modifier_effects() {
...
 	ret &= add_modifier_effect("social_reform_desire", false);
 	ret &= add_modifier_effect("supply_consumption", false);
+	ret &= add_modifier_effect("tax_efficiency", true);
 	ret &= add_modifier_effect("army_tech_research_bonus", true);
 	ret &= add_modifier_effect("commerce_tech_research_bonus", true);
 	ret &= add_modifier_effect("culture_tech_research_bonus", true);
 	ret &= add_modifier_effect("industry_tech_research_bonus", true);
-	ret &= add_modifier_effect("naval_tech_research_bonus", true);
-	ret &= add_modifier_effect("unit_start_experience", true); // weird, naval_unit_start_experience = 15 would give a 15% boost
+	ret &= add_modifier_effect("navy_tech_research_bonus", true);
```
schombert, line 201: If you don't want to hard-code the building max modifiers, you may not want to hard-code these either as they are derived from the names of the tech categories. Does any mod add a tech category? That I don't know.

src/openvic-simulation/Modifier.cpp
Hop311: I couldn't because `ModifierEffect`'s constructor is `private`, accessible by `ModifierManager` because it's a friend. Thinking about it again, I should be able to make `std::make_unique` a friend instead. That will publicly expose `ModifierEffect` construction though, at which point I might as well just make the constructor public and not need a friend declaration.

src/openvic-simulation/Modifier.cpp
Hop311: Thanks for letting me know, dunno about mods changing categories but since we're making it possible to change them then the changes should propagate all the way through.

src/openvic-simulation/Modifier.cpp
Spartan322: Honestly making the constructor private makes little sense if the registry is locked, the member variables don't even have to be private (or const) necessarily if we disable the const functions of the registry, should however rename the non-const overload functions to something like `<name>_nc` or `<name>_mutable to clarify non-const/mutable return value.

BrickPi (approved):
Looks good generally, primarily reviewed history changes, probably want to wait for Spartan to check off before merge.

Spartan322 (changes requested):
Could this be squashed a little bit at least, at minimum enough to remove the commits responsible for only fixing things that target this PR? (like especially if a commit was not building, it might as well be ignored, but also like fix and tweak commits should be squashed down to keep the PR's commits at least more relevant, making regression testing easier)

Hop311:
> Could this be squashed a little bit at least, at minimum enough to remove the commits responsible for only fixing things that target this PR? (like especially if a commit was not building, its existence might as well be ignored, but also like fix and tweak commits should be squashed down to keep the PR's commits at least more relevant, making regression testing easier)

I agree most commits could do with being renamed, and as for squashing I'm happy to where it makes sense, did you have any particular commits in mind (going off the actual contents of the commits, rather than just their names)?

Spartan322 (approved):

PR #66, opened 2023-10-26
Implement Province History Loading

BrickPi:
Loads province history, tests passing.

### DIFF
--- a/src/openvic-simulation/history/ProvinceHistory.cpp
+++ b/src/openvic-simulation/history/ProvinceHistory.cpp
@@ -1,0 +1,263 @@
+#include "ProvinceHistory.hpp"
+
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "openvic-simulation/GameManager.hpp"
+
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
+
+ProvinceHistory::ProvinceHistory(
+    Country const* new_owner,
+    Country const* new_controller,
+    std::vector<Country const*>&& new_cores,
+    Good const* new_rgo,
+    uint8_t new_life_rating,
+    TerrainType const* new_terrain_type,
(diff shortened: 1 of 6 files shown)


src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -0,0 +1,260 @@
...
+        if (current.first > entry) continue;
+        if (current.first > closest_entry && current.first < entry) closest_entry = current.first;
+    }
+
+    auto entry_registry = province_registry->second.find(closest_entry);
+    if (entry_registry != province_registry->second.end()) {
+        return &entry_registry->second;
+    }
+	/* warned about lack of entries earlier, return nullptr */
+	return nullptr;
```
Spartan322, line 145: Indenting is a little off.

Spartan322 (approved):
Looks fine enough I suppose, should wait for @Hop311 before merge.

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -0,0 +1,260 @@
...
+                uint8_t level;
+
+                bool ret = game_manager.get_economy_manager().get_building_manager().expect_building_str(assign_variable_callback_pointer(building))(key);
+                ret &= expect_uint(assign_variable_callback(level))(value);
+
+                buildings.emplace(building, level);
+                updated_buildings = true;
+                return ret;
+            }
+            return true;
```
Hop311, line 180:
```suggestion
            return key_value_invalid_callback(key, value);
```
Currently invalid keys are ignored - a simple solution would be to return `false` instead, but that wouldn't print an error message. Using `key_value_invalid_callback` can save you the trouble of writing that error message out yourself.

Hop311 (approved):

Hop311 (approved):

PR #67, opened 2023-10-29
Formating

Hop311:
- Add `.astylesrc` format file and script, in general it works well with the only flaws being an occasional missing/extra space either side of an operator or pointer/(move) reference, and an occasional missing/extra indent level (consistently across a block, so it's easy to find and fix).
- Updated `.clang-format` format file and added a script, it supports some features `astyle` doesn't and fixes most of its mistakes, but `clang-format`'s own mistakes are far worse, ranging from absolutely mangling indentation (very uneven and hard to fix) to not bothering at all and just moving everything onto one massive line before giving up.

By applying both of these tools (and doing a LOT of manual cleaning up) I've got the code in a mostly consistent style, following various rules including a maximum column width of 128 on each line (NOT 128 characters, e.g. a single tab contributes 4 column width).
In future PRs, authors should use `astyle` and `clang-format` to format the files they change, taking care to clean up any messy parts, especially in sections of code they haven't changed (and so the code not matching the formatted version is likely deliberate)

### DIFF
--- a/.astylesrc
+++ b/.astylesrc
@@ -1,0 +1,217 @@
+# @file Astyle code automatic formatting settings
+# @see http://astyle.sourceforge.net/astyle.html#_General_Information
+
+#
+# Java style uses attached braces.
+#
+# int Foo(bool isBar) {
+#     if (isBar) {
+#         bar();
+#         return 1;
+#     } else
+#         return 0;
+# }
+#
+--style=attach
+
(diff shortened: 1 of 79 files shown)


BrickPi (dismissed):
astyle & clang-format look good

BrickPi (approved):

PR #68, opened 2023-10-29
Improve Province Histories

BrickPi:
Fix a few variations of keys, add slave state status for split states, use `std::bitset` over individual update booleans.


Hop311 (approved):
LGTM :+1:

PR #70, opened 2023-11-07
Diplomacy History

BrickPi:
Loads `cb_types.txt`, as well as alliance, subject, & war history.

Tests passing.


src/openvic-simulation/history/DiplomaticHistory.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
+
+WarHistory::added_wargoal_t::added_wargoal_t(
+    Date new_added,
+    Country const* new_actor,
+    Country const* new_receiver,
+    WargoalType const* new_wargoal,
+    std::optional<Country const*> new_third_party,
```
schombert, line 14: Keep in mind that wargoals can also liberate dead countries. Certainly you can make a country object for each tag (or even combine them, such that there is a single object representing both the country and the tag) and instantiate countries for dead tags, but whatever design you end up with, it will also have to accommodate a country changing its tag, and this may cause issues if there is a "dead" country that is already mapped to that tag

src/openvic-simulation/military/Wargoal.hpp
```diff
@@ -0,0 +1,127 @@
...
+        PO_REPAY_DEBT =            0b0000001000000000,
+        PO_REMOVE_PRESTIGE =       0b0000000100000000,
+        PO_MAKE_PUPPET =           0b0000000010000000,
+        PO_RELEASE_PUPPET =        0b0000000001000000,
+        PO_STATUS_QUO =            0b0000000000100000,
+        PO_INSTALL_COMMUNISM =     0b0000000000010000,
+        PO_REMOVE_COMMUNISM =      0b0000000000001000,
+        PO_REMOVE_CORES =          0b0000000000000100, // only usable with ANNEX, DEMAND_STATE, or TRANSFER_PROVINCES
+        PO_TRANSFER_PROVINCES =    0b0000000000000010,
+        PO_CLEAR_UNION_SPHERE =    0b0000000000000001
```
schombert, line 26: looks like you are missing dismantle forts from this list

schombert: also, technically po_destroy_naval_bases, although I don't know if a mod has ever used that without also destroy forts

src/openvic-simulation/history/DiplomaticHistory.cpp
BrickPi: These structs are used exclusively for history, I believe there will be different structures for general gameplay that will have to eventually take into consideration OOTB mode.

src/openvic-simulation/military/Wargoal.hpp
BrickPi: Yeah that's where PO_DISARMAMENT comes from, I'll test sometime if they can be used discretely

schombert: po_disarmament is its own thing. It appears in the cut down to size cb in the base game. I am talking about
```
	po_destroy_forts = yes
	po_destroy_naval_bases = yes
```
which can be found in the base game's dismantle forts cb

src/openvic-simulation/history/DiplomaticHistory.cpp
Spartan322: Optimally it would be best for us to run through the history data for necessary loading and then delete them asap since they contain a massive amount of wasted memory on things that will never change, pretty much all of the history that isn't being immediately used actually could be cache on a storage drive and we'd only get advantages out of doing so.

Spartan322 (changes requested):
Change the space indents in Wargoal and DiplomaticHistory to tabs.

src/openvic-simulation/military/Wargoal.hpp
```diff
@@ -0,0 +1,129 @@
...
+	struct WargoalTypeManager;
+
+	enum class peace_options_t : uint32_t {
+		PO_ANNEX =                 0b1000000000000000,
+		PO_DEMAND_STATE =          0b0100000000000000,
+		PO_COLONY =                0b0010000000000000,
+		PO_ADD_TO_SPHERE =         0b0001000000000000,
+		PO_DISARMAMENT =           0b0000100000000000,
+		PO_REMOVE_FORTS =           0b0000100000000000,
+		PO_REMOVE_NAVAL_BASES =           0b0000100000000000,
```
Spartan322, line 17: Formatting for these two is broken.

Spartan322 (approved):

Spartan322 (approved):

PR #71, opened 2023-11-07
GUI and GFX file loading

Hop311:


src/openvic-simulation/interface/GUI.hpp
```diff
@@ -0,0 +1,230 @@
...
+
+#include "openvic-simulation/interface/GFX.hpp"
+
+namespace OpenVic {
+	class UIManager;
+}
+namespace OpenVic::GUI {
+	class Scene;
+
+	class Element : virtual public Named<UIManager const&> {
```
Spartan322, line 11: macOS Clang still does not support something with a virtual base class containing a constexpr member function.

Hop311: Nah not really, since writing that code I've realised that windows and the top level of GUI files are less similar than I thought, so it shouldn't be an issue to restructure them.

Spartan322 (approved):
I kinda wonder if it wouldn't be more efficient just to stick the GUI stuff in the extension and directly generate nodes that way so we don't have to write a conversion, but whatever, this works for now.

Spartan322 (changes requested):
Actually could you rebase the virtual changes into the first commit? There's no reason for them to be a separate commit.

src/openvic-simulation/dataloader/Dataloader.hpp
```diff
@@ -70,6 +77,8 @@ namespace OpenVic {
 		 * DAT-24
 		 */
 		fs::path lookup_file(std::string_view path, bool print_error = true) const;
+		/* Checks alternate file endings, e.g. if "*.tga" doesn't exist then try "*.dds" */
```
BrickPi, line 80: This should potentially be renamed/extended, being able to pass multiple allowed file extensions seems useful generally.

src/openvic-simulation/interface/GUI.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+ListBox::ListBox() : size {} {}
+
+bool ListBox::_fill_key_map(NodeTools::key_map_t& key_map, UIManager const& ui_manager) {
+	bool ret = Element::_fill_key_map(key_map, ui_manager);
+	ret &= add_key_map_entries(key_map,
+		"backGround", ZERO_OR_ONE, success_callback,
+		"size", ONE_EXACTLY, expect_ivec2(assign_variable_callback(size)),
+		"spacing", ZERO_OR_ONE, success_callback,
+		"scrollbartype", ZERO_OR_ONE, success_callback,
```
BrickPi, line 178:
```suggestion
		"scrollbartype", ZERO_OR_ONE, success_callback, // TODO: Implement modable listbox scrollbars
```

src/openvic-simulation/interface/GFX.cpp
```diff
@@ -0,0 +1,98 @@
...
+		"spriteType", ZERO_OR_MORE, _expect_instance<Sprite, TextureSprite>(callback),
+		"progressbartype", ZERO_OR_MORE, _expect_instance<Sprite, ProgressBar>(callback),
+		"PieChartType", ZERO_OR_MORE, _expect_instance<Sprite, PieChart>(callback),
+		"LineChartType", ZERO_OR_MORE, _expect_instance<Sprite, LineChart>(callback),
+		"textSpriteType", ZERO_OR_MORE, _expect_instance<Sprite, TextureSprite>(callback),
+		"maskedShieldType", ZERO_OR_MORE, _expect_instance<Sprite, MaskedFlag>(callback),
+		"corneredTileSpriteType", ZERO_OR_MORE, success_callback,
+		"tileSpriteType", ZERO_OR_MORE, success_callback,
+		"BarChartType", ZERO_OR_MORE, success_callback,
+		"scrollingSprite", ZERO_OR_MORE, success_callback
```
BrickPi, line 21:
```suggestion
		"scrollingSprite", ZERO_OR_MORE, success_callback // TODO: Implement modable scrollbar sprites
```

src/openvic-simulation/interface/GFX.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+bool TextureSprite::_fill_key_map(key_map_t& key_map) {
+	bool ret = Sprite::_fill_key_map(key_map);
+	ret &= add_key_map_entries(key_map,
+		"texturefile", ZERO_OR_ONE, expect_string(assign_variable_callback_string(texture_file)),
+		"textureFile", ZERO_OR_ONE, expect_string(assign_variable_callback_string(texture_file)),
+		"noOfFrames", ZERO_OR_ONE, expect_uint(assign_variable_callback(no_of_frames)),
+
+		"norefcount", ZERO_OR_ONE, success_callback,
+		"effectFile", ZERO_OR_ONE, success_callback,
```
BrickPi, line 35:
```suggestion
		"effectFile", ZERO_OR_ONE, success_callback, // TODO: Implement GFX Effects
```
I can image implementing effects will be a lot of fun.

BrickPi (approved):
Great work on GUIs & thanks for going through and `PROPERTY`ising WargoalTypes

src/openvic-simulation/interface/GFX.cpp
Hop311: Yeah I really hope the shader conversion is feasible, gotta get Spartan on the case

src/openvic-simulation/dataloader/Dataloader.hpp
Hop311: I've renamed it to `lookup_image_file_or_dds`, originally it used a list of potential endings but I simplified it once I realised the only case that needs changing (in vanilla at least) is `*.tga` -> `*.dds`. The function doesn't actually check if the input is `*.tga`, if you pass it `*.dds` it'll attempt to look it up twice.
I'll leave the implementation as is for now, we can decide whether to treat this as a special case to account for mistakes in the defines, or whether we want to support fuzzy image format for GFX asset loading more broadly (`*.dds` -> `*.tga` as well as other image formats) so that UI modders don't have to worry about preserving the (often archaic) image formats or have to modify the GFX files if all they're changing is the contents of an asset.

Spartan322 (approved):

PR #73, opened 2023-11-12
Apply Province Histories

BrickPi:
- Adds a function to apply province history to provinces
- Adds missing properties to Province structure
- Adds setter to `BuildingInstance` level
Tests passing.


Hop311 (approved):
Really great work, exactly what we need for the starting political mapmode + fleshing out the province overview panel

PR #74, opened 2023-11-12
Country History Improvements

BrickPi:
Adds simple `PROPERTY_RW` macro for getters & setters.
Adds `CountryInstance` class to represent in-game countries.
Adds function to apply country history to country instances.

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.hpp
+++ b/src/openvic-simulation/country/CountryInstance.hpp
@@ -1,0 +1,39 @@
+#pragma once
+
+#include "openvic-simulation/country/Country.hpp"
+#include "openvic-simulation/history/CountryHistory.hpp"
+#include "openvic-simulation/map/Province.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+namespace OpenVic {
+    /* Representation of an existing country that is currently in-game. */
+    struct CountryInstance {
+    private:
+        Country const* PROPERTY_RW(base_country);
+        Culture const* PROPERTY_RW(primary_culture);
+        std::vector<Culture const*> PROPERTY(accepted_cultures);
+        Religion const* PROPERTY_RW(religion);
+        CountryParty const* PROPERTY_RW(ruling_party);
+        Date PROPERTY_RW(last_election);
+        decimal_map_t<Ideology const*> PROPERTY(upper_house);
+        Province const* PROPERTY_RW(capital);
+        GovernmentType const* PROPERTY_RW(government_type);
+        fixed_point_t PROPERTY_RW(plurality);
+        NationalValue const* PROPERTY_RW(national_value);
+        bool PROPERTY_RW(civilised);
+        fixed_point_t PROPERTY_RW(prestige);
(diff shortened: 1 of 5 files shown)


src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -0,0 +1,39 @@
...
+        CountryParty const* PROPERTY_RW(ruling_party)
+        Date PROPERTY_RW(last_election)
+        decimal_map_t<Ideology const*> PROPERTY(upper_house)
+        Province const* PROPERTY_RW(capital)
+        GovernmentType const* PROPERTY_RW(government_type)
+        fixed_point_t PROPERTY_RW(plurality)
+        NationalValue const* PROPERTY_RW(national_value)
+        bool PROPERTY_RW(civilised)
+        fixed_point_t PROPERTY_RW(prestige)
+        std::vector<Reform const*> PROPERTY(reforms)
```
Hop311, line 25: This works for now as a place to put the country history reforms, eventually we'll need a map from reform groups to single active reforms from that group.

src/openvic-simulation/utility/Getters.hpp
```diff
@@ -63,4 +63,13 @@ public: \
...
 private:
 
+#define PROPERTY_RW(NAME) \
+	PROPERTY(NAME) \
+public: \
+	void set_##NAME(decltype(NAME) new_##NAME) { \
+		NAME = new_##NAME; \
+	} \
+\
+private:
```
Hop311, line 73: Good job with how you've used this macro, only with things which we're happy to pass by value (so primitive types and basic primitive type wrappers like `Date` and `fixed_point_t`). As for special logic to choose the argument type and control how it's assigned, let's see what Spartan can come up with (but that can be a future PR).

Hop311 (commented):
Good work with the macro and `CountryInstance`, it's gonna be one of the most important structs in the game so I'm glad to see it finally added.
*As a note for the future*, I notice you're often having to check to avoid suplicates in vectors - the solution would be `std::set` except that doesn't preserve the order elements get added in so we'll need something else, could just be a wrapper around `std::vector` with a special insert function but I'm sure Spartan knows some useful libraries we could take a look at.

Hop311:
Also you used `PROPERTY_RW` in the first commit and only added the macro itself in the second one, then the third is a tiny change - might as well squash it all into a single commit.

Hop311:
Also thinking again, some things in CountryInstance should be moved to Country, e.g. cores are tied to the tag, when s

BrickPi:
> Also thinking again, some things in CountryInstance should be moved to Country, e.g. cores are tied to the tag, when switching tag, eg via a decision, you have to replace the old tag's cores with the new ones yourself as part of the effect script. Others I'm not sure about and may need testing, it's possible there are even things that stay with the tag _and_ instance upon tag switch.

I was thinking we should keep `Country` just static values. Cores are currently stored by province which is where I think they fit best.

Hop311 (approved):
If we can keep Country as a static definition that would be ideal, especially for being able to reset and apply a different bookmark without needing to reload the starting state. We'll need to do some testing, eg switch to a new tag then release the original tag again and see what data it retains.

PR #75, opened 2023-11-14
Accumulated changes

Hop311:
- Changed mapmodes to have 2 colours (base and stripe).
- Removed default land/sea colour tinting from mapmodes,  now terrain mapmode has no province tinting at all.
- Renamed `decimal_map_t` to `fixed_point_map_t`, moved it to `FixedPointMap.hpp` and added more helper functions (for finding the second largest item in a distribution and checking if it's at least a third of the total, for culture and religion mapmodes).
- Organised various map reset and setup code into a chain of `reset` functions and a chain of `apply_history_...` functions, with its tip (at least in this repo) being `GameManager::load_bookmark`, which resets and then loads province history up to the date of the bookmark.
- Moved `Modifier.*pp` to the `misc` folder.
- Changed various things to use the `PROPERTY` macro.
- Made `expect_assign` print an error when the callback fails, helping with errors inside lists and dictionaries.

### DIFF
--- a/src/openvic-simulation/economy/Building.hpp
+++ b/src/openvic-simulation/economy/Building.hpp
@@ -18,7 +18,16 @@
 namespace OpenVic {
 
 	struct BuildingManager;
-	struct BuildingType;
+
+	struct BuildingType : HasIdentifier {
+		friend struct BuildingManager;
+
+	private:
+		BuildingType(std::string_view new_identifier);
+
+	public:
+		BuildingType(BuildingType&&) = default;
+	};
 
 	/* REQUIREMENTS:
 	 * MAP-11, MAP-72, MAP-73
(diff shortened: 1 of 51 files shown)


src/openvic-simulation/map/Map.hpp
```diff
@@ -14,7 +14,10 @@ namespace OpenVic {
 	struct Mapmode : HasIdentifier {
 		friend struct Map;
 
-		using colour_func_t = std::function<colour_t(Map const&, Province const&)>;
+		/* Bottom 32 bits are the base colour, top 32 are the stripe colour, both in ARGB format with the alpha channels
+		 * controlling interpolation with the terrain colour (0 = all terrain, 255 = all corresponding RGB) */
+		using base_stripe_t = uint64_t;
```
BrickPi, line 19: I'm unsure about the name `base_stripe_t` however it'll have to work for now, I don't have a better name on hand, perhaps it should just be `colour_t`

src/openvic-simulation/economy/Building.hpp
```diff
@@ -18,7 +18,16 @@
 namespace OpenVic {
 
 	struct BuildingManager;
-	struct BuildingType;
+
+	struct BuildingType : HasIdentifier {
```
BrickPi, line 22: Building.*pp should probably be separated into BuildingType & BuildingInstance files.

BrickPi (changes requested):
Lots of good work :)

src/openvic-simulation/map/Map.hpp
Hop311: Yeah it was the best I could come up with, IMO colour_t would cause confusion with the single colour type, especially since it's in the OpenVic namespace so inside Mapmode it would be ambiguous which one it refers to.

src/openvic-simulation/economy/Building.hpp
Hop311: I guess yeah, BuildingManager doesn't rely on BuildingInstance, this PR didn't introduce that arrangement but since I'm already rearranging Building.hpp won't hurt to force push another change

BrickPi (approved):

PR #76, opened 2023-11-15
Implement National Focus Loading

BrickPi:
Loads `common/national_focus.txt`
Adds launch.json for MSVC debugging
Fixes bug in `expect_modifier_value_and_keys_and_default()`

Needs testing.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "(Windows) Launch",
+            "type": "cppvsdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/bin/openvic-simulation.headless.windows.template_debug.dev.x86_64.exe",
+            "args": [],
+            "stopAtEntry": false,
+            "cwd": "${fileDirname}",
+            "environment": [],
+            "console": "integratedTerminal"
+        }
+
+    ]
+}
--- a/src/openvic-simulation/politics/NationalFocus.cpp
+++ b/src/openvic-simulation/politics/NationalFocus.cpp
@@ -1,0 +1,118 @@
+#include "NationalFocus.hpp"
+
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
+
+NationalFocusGroup::NationalFocusGroup(std::string_view new_identifier) : HasIdentifier { new_identifier } {}
+
+NationalFocus::NationalFocus(
+	std::string_view new_identifier,
+	uint8_t new_icon,
+	NationalFocusGroup const& new_group,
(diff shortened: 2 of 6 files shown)


src/openvic-simulation/politics/NationalFocus.cpp
```diff
@@ -0,0 +1,107 @@
...
+			NationalFocus::pop_promotion_map_t promotions;
+			NationalFocus::party_loyalty_map_t loyalties;
+			NationalFocus::production_map_t production;
+
+			Ideology const* last_specified_ideology = nullptr; // weird, I know
+
+			bool ret = modifier_manager.expect_modifier_value_and_keys_and_default(
+				move_variable_callback(modifiers),
+				[&promotions, &pop_manager, &production, &good_manager, &modifiers, &modifier_manager](std::string_view key, ast::NodeCPtr value) -> bool {
+					if (pop_manager.has_pop_type_identifier(key)) {
```
Hop311, line 66: Under the hood this searches for the pop type pointer and checks it's not null, if you get the pointer and do the check yourself you can avoid an unnecessary second search. (Same thing for the goods check beneath this)

.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
Hop311, line 1: Did you mean to include this file?

BrickPi: Yeah, it's a generic VSCode debugging setup for windows.

src/openvic-simulation/politics/NationalFocus.hpp
```diff
@@ -0,0 +1,76 @@
...
+	struct NationalFocus : HasIdentifier {
+		friend struct NationalFocusManager;
+
+	public:
+		using pop_promotion_map_t = std::map<PopType const*, fixed_point_t>;
+		using party_loyalty_map_t = std::map<Ideology const*, fixed_point_t>;
+		using production_map_t = std::map<Good const*, fixed_point_t>;
+
+	private:
+		std::string PROPERTY(icon);
```
Hop311, line 31: This should be a uint*_t (width is up to you), and check in add_national_focus should be that it's non-zero (the 0-th icon is the "no/unset national focus" one)

Hop311 (commented):
Needs one more small change I just noticed, also it's failing when I test it but with no error message beyond the one in Dataloader.cpp, investigating

src/openvic-simulation/politics/NationalFocus.cpp
```diff
@@ -0,0 +1,118 @@
...
+						return ret;
+					}
+					Good const* good = good_manager.get_good_by_identifier(key);
+					if (good != nullptr) {
+						fixed_point_t boost;
+						bool ret = expect_fixed_point(assign_variable_callback(boost))(value);
+						production[good] = boost;
+						return ret;
+					}
+					return false;
```
Hop311, line 89:
```suggestion
					return key_value_invalid_callback(key, value);
```
OK this revealed the error, invalid key `railroads` (modifier affecting capitalist chance to build railroads). Add `ret &= add_modifier_effect("railroads", true);` to `Modifier.cpp` and everything loads successfully.
In general everywhere you `return false;` should have an error message so we know where the failure occurred, here `key_value_invalid_callback` is basically a prepackaged error message.

Hop311 (approved):
Great work Brick 👍

PR #78, opened 2023-11-22
Refactoring (*mostly* related to the property macro)

zaaarf:
TL;DR:
- Expanded a bit the `PROPERTY` macro (the one to generate getters/setters) to support some more flexibility
- Used it and its variants whenever possible
- Renamed some things that were in lowerCamelCase to snake_case, to match the project's overall style
  - If this was a mistake, let me know, it should be as simple as rolling back one commit

Compiles and runs successfully on linux x64 with gcc, but test it for yourself.


Hop311 (approved):
Great work, thanks for being so thorough 👍 Glad I've managed to dodge doing this myself 😁

PR #79, opened 2023-11-28
States First Pass

BrickPi:
First pass, need some eyes on it. Overall philosophy is that the big `regions` vector should be kept up to date by functions that modify it with validation for states that need to be split or merged, but that might be better to eventually work into the `State` class itself.

Adds `State` struct.
Adds state member to Provinces
Adds logic for generating states, called on bookmark load.

### DIFF
--- a/src/openvic-simulation/GameManager.cpp
+++ b/src/openvic-simulation/GameManager.cpp
@@ -59,6 +59,7 @@ bool GameManager::load_bookmark(Bookmark const* new_bookmark) {
 	}
 	today = bookmark->get_date();
 	ret &= map.apply_history_to_provinces(history_manager.get_province_manager(), today);
+	map.get_state_manager().generate_states(map);
 	// TODO - apply country history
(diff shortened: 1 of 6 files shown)


schombert:
In Victoria 2 it is possible for the same "region" to become 2 states within a single nation if some of the provinces within it are colonial and some are not. If I am reading it correctly, your logic will result in all of the provinces in the region becoming a single state regardless of whether their colonial status is the same or not.

BrickPi:
> If I am reading it correctly, your logic will result in all of the provinces in the region becoming a single state regardless of whether their colonial status is the same or not.

```C++
inline bool operator==(const State& lhs, const State& rhs) {
    	return (lhs.get_owner() == rhs.get_owner() && lhs.get_colony_status() == rhs.get_colony_status());
}
```
It checks for colony status as well as owner, thank you for taking the time to look though!

Spartan322 (approved):

PR #82, opened 2023-12-02
Accumulated changes from Dev Diary GUI focus period

Hop311:
- Crime, event, static and triggered modifier loading (apart from triggers which await condition loading).
- Image lookup changed to try `.dds` first when given a `.tga` path (thank you paradox, very cool).
- First rgo owner, `is_slave` and `administrative_efficiency` pop types' icons are cached for use in the UI, so we don't need to hardcode them.
- Added a bunch of `PopType` variables + parsing code for them, and added TODO comments for all the rest.
- Added month names for use in the UI.
- Fixed `UniqueKeyRegistry::get_item_by_index` to work for `_Storage != _Type` type varieties (i.e. "Instance" versions with `_Storage = std::unique_ptr<_Type>`).
- Added `get_item_by_index` to `IDENTIFIER_REGISTRY[_NON_CONST]_ACCESSORS*` macros with an optional argument to shift indices down by, used for provinces where the first province in the registry has index 1 and 0 is the null/unset province.
- Added option for registry variable name to be different from the plural item name for `IDENTIFIER_REGISTRY[_NON_CONST]_ACCESSORS*`.


src/openvic-simulation/misc/Modifier.hpp
```diff
@@ -80,6 +80,32 @@ namespace OpenVic {
...
+	public:
+		TriggeredModifier(TriggeredModifier&&) = default;
+	};
+
+	struct Crime final : TriggeredModifier {
+		friend struct ModifierManager;
+
+	private:
+		const bool PROPERTY(default_active);
+		bool PROPERTY_RW(active);
```
BrickPi, line 101: Would prefer Crime being its own files, even if its small, for organisation.

src/openvic-simulation/tech/Technology.cpp
```diff
@@ -100,47 +113,52 @@ bool TechnologyManager::load_technology_file_schools(ModifierManager const& modi
 bool TechnologyManager::load_technologies_file(ModifierManager const& modifier_manager, UnitManager const& unit_manager, BuildingManager const& building_manager, ast::NodeCPtr root) {
 	return expect_dictionary_reserve_length(technologies, [this, &modifier_manager, &unit_manager, &building_manager](std::string_view tech_key, ast::NodeCPtr tech_value) -> bool {
 		ModifierValue modifiers;
-		std::string_view area_identifier;
-		Technology::year_t year;
-		fixed_point_t cost;
+		TechnologyArea const* area = nullptr;
+		Technology::year_t year = 0;
```
BrickPi, line 117: Why are these not just `Date::year_t`?

BrickPi (commented):
Nice job

src/openvic-simulation/tech/Technology.cpp
zaaarf: This one is on me, sorry! I made my own alias in Technology because I forgot we had a centralised one.

zaaarf (approved):

PR #87, opened 2023-12-09
Dataloading Country History - Part 2 

zaaarf:
Added:
- `Technology` loading
- `TechnologySchool` loading
- `Invention` loading
- Foreign investment loading
- A bunch of other keys that were missing, it only *seemed* to work because we had `success_callback` as default

### DIFF
--- a/src/openvic-simulation/history/CountryHistory.cpp
+++ b/src/openvic-simulation/history/CountryHistory.cpp
@@ -21,10 +21,13 @@ bool CountryHistoryMap::_load_history_entry(
 	PoliticsManager const& politics_manager = game_manager.get_politics_manager();
 	IssueManager const& issue_manager = politics_manager.get_issue_manager();
 	CultureManager const& culture_manager = game_manager.get_pop_manager().get_culture_manager();
+	CountryManager const& country_manager = game_manager.get_country_manager();
+	TechnologyManager const& technology_manager = game_manager.get_research_manager().get_technology_manager();
+	InventionManager const& invention_manager = game_manager.get_research_manager().get_invention_manager();
 
 	return expect_dictionary_keys_and_default(
-		[this, &game_manager, &dataloader, &deployment_manager, &issue_manager, &entry](
-			std::string_view key, ast::NodeCPtr value) -> bool {
+		[this, &game_manager, &dataloader, &deployment_manager, &issue_manager, 
+			&technology_manager, &invention_manager, &country_manager, &entry](std::string_view key, ast::NodeCPtr value) -> bool {
 			ReformGroup const* reform_group = issue_manager.get_reform_group_by_identifier(key);
 			if (reform_group != nullptr) {
 				return issue_manager.expect_reform_identifier([&entry, reform_group](Reform const& reform) -> bool {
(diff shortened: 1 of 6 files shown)


src/openvic-simulation/history/CountryHistory.hpp
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <map>
+#include <optional>
+#include <unordered_map>
```
Hop311, line 5: Unused header

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -1,12 +1,13 @@
...
 
 #include "openvic-simulation/GameManager.hpp"
+#include <string_view>
 
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
 CountryHistoryEntry::CountryHistoryEntry(Country const& new_country, Date new_date)
-	: HistoryEntry { new_date }, country { new_country } {}
+	: HistoryEntry { new_date }, country { new_country }, releasable_vassal { true } {}
```
Hop311, line 10:
```suggestion
	: HistoryEntry { new_date }, country { new_country } {}
```
`releasable_vassal` is a `std::optional<bool>`, so we only want it to be set to anything if it shows up in the history entry. There'll be an equivalent variable in either `Country` or `CountryInstance` which defaults to true, but here it should default to `std::nullopt_t` (which it will do if you don't explicitly initialise it).

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -46,9 +50,30 @@ bool CountryHistoryMap::_load_history_entry(
...
+			if (technology != nullptr) {
+				uint8_t flag = -1;
+				if (expect_uint(assign_variable_callback(flag))(value)) {
+					if (flag == 1) return entry.technologies.emplace(technology, true).second;
+					else if (flag == 0) return entry.technologies.emplace(technology, false).second;
+					else {
+						Logger::warning("Refusing to load country history technology with non-boolean value: ", key, ", ", flag);
+						return true;
+					}
+				} else return false;
```
Hop311, line 64: You can do this similarly to how you've done inventions by using `expect_int_bool`, which converts 0->false, 1->true.

src/openvic-simulation/history/CountryHistory.hpp
```diff
@@ -36,12 +40,25 @@ namespace OpenVic {
...
+		std::map<Invention const*, bool> PROPERTY(inventions);
+		fixed_point_map_t<Country const*> PROPERTY(foreign_investment);
+		std::optional<fixed_point_t> PROPERTY(consciousness);
+		std::optional<fixed_point_t> PROPERTY(nonstate_consciousness);
+		std::optional<fixed_point_t> PROPERTY(literacy);
+		std::optional<fixed_point_t> PROPERTY(nonstate_culture_literacy);
+		std::optional<bool> PROPERTY_CUSTOM_PREFIX(releasable_vassal, is);
+		std::optional<fixed_point_t> PROPERTY(colonial_points);
+		std::unordered_set<std::string> PROPERTY(country_flags);
+		std::unordered_set<std::string> PROPERTY(global_flags);
```
Hop311, line 58:
```suggestion
		string_set_t PROPERTY(country_flags);
		string_set_t PROPERTY(global_flags);
```
Although the order here probably doesn't matter, I'd prefer to stay on the safe side and use `string_set_t` that guaruntees the same order across all machines. Could you also change the `std::unordered_set`s in `Technology.hpp` and `Invention.hpp` to `std::set`s?.

Hop311 (commented):
Great work, the history loading code is perfect, thanks for going beyond tech & inventions to finish loading most of the rest of the file's contents 👍

src/openvic-simulation/history/CountryHistory.cpp
zaaarf: oh, right, didn't think of that

src/openvic-simulation/history/CountryHistory.hpp
zaaarf: sure thing!

src/openvic-simulation/history/CountryHistory.cpp
zaaarf: Oh, i just assumed it would only work with actual bool types

Hop311 (approved):
Another great PR, thank you

PR #88, opened 2023-12-09
Make vec2_t header only (friend functions weren't being found)

Hop311:


zaaarf (approved):

zaaarf:
To be thorough, I even built and tested it locally, and it works fine. I don't see anything that went wrong (or, really, that *could* go wrong, you're basically just moving stuff to the header)

PR #89, opened 2023-12-10
Pop History Loading (with the new system)

zaaarf:
This one was annoying. The code itself should work fine (I tested it), but I don't rule out typos or small mistakes - it is pretty late, after all.

Also, I had to change the access level of a few things to make a sensible loading flow. I expect there to be better ways I didn't think of to go about this.

### DIFF
--- a/src/openvic-simulation/history/ProvinceHistory.cpp
+++ b/src/openvic-simulation/history/ProvinceHistory.cpp
@@ -2,7 +2,6 @@
 
 #include "openvic-simulation/GameManager.hpp"
 #include "openvic-simulation/dataloader/NodeTools.hpp"
-#include <memory>
 
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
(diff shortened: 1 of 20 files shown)


src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -15,6 +16,37 @@ std::unique_ptr<ProvinceHistoryEntry> ProvinceHistoryMap::_make_entry(Date date)
 	return std::unique_ptr<ProvinceHistoryEntry> { new ProvinceHistoryEntry { province, date } };
 }
 
+std::unique_ptr<ProvinceHistoryEntry> ProvinceHistoryMap::_get_entry(Date date) {
+	return std::move(entries.at(date));
+}
```
Spartan322, line 21: Why is this returning a unique_ptr and not a pointer?

zaaarf: Because `entries` is a map of `Date` to `unique_ptr`, and I had no idea of how to unwrap it.

Spartan322: [`std::unique_ptr::operator*`](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator*) and [`std::unique_ptr::get`](https://en.cppreference.com/w/cpp/memory/unique_ptr/get)

Returning a unique_ptr here is bad because moving a unique_ptr swaps the data with a nullptr, so the pointer entry in `entries` will become null. Every case for which you move a unique_ptr does this, never move a unique_ptr unless you are moving the ownership of the data, as in the responsibility for freeing the pointer data. This is what dereferencing the unique_ptr is for.

zaaarf: I understand, I'll fix it right now.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -157,26 +161,29 @@ bool PopManager::load_pop_type_file(
...
+	fixed_point_t militancy = 0, consciousness = 0;
+	RebelType const* rebel_type = nullptr;
+
 	bool ret = expect_dictionary_keys(
 		"culture", ONE_EXACTLY, culture_manager.expect_culture_identifier(assign_variable_callback_pointer(culture)),
 		"religion", ONE_EXACTLY, religion_manager.expect_religion_identifier(assign_variable_callback_pointer(religion)),
-		"size", ONE_EXACTLY, expect_uint(assign_variable_callback(size)),
-		"militancy", ZERO_OR_ONE, success_callback,
-		"rebel_type", ZERO_OR_ONE, success_callback
+		"size", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(size)),
```
wvpm, line 176: Decimal sizes are accepted when loading data but are truncated in game.
Should it log a warning when size % 1 > 0 ?

Hop311: Doing this for each pop individually would cause too much warning spam, since hundreds if not thousands of pops have non-integer sizes in 1861. We could easily reduce it to one warning per file, but ideally it'd be a single warning if any pops in the start date have non-integer sizes. I'll add this later today.

Hop311 (dismissed):

Hop311 (approved):

PR #92, opened 2023-12-11
on_action dataloading

zaaarf:
This one was pretty straightforward. The one thing I'm not sure on is that conversion from string to int - is there a better way?

### DIFF
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -668,6 +668,7 @@ bool Dataloader::load_defines(GameManager& game_manager) const {
 	static constexpr std::string_view event_modifiers_file = "common/event_modifiers.txt";
 	static constexpr std::string_view static_modifiers_file = "common/static_modifiers.txt";
 	static constexpr std::string_view triggered_modifiers_file = "common/triggered_modifiers.txt";
+	static constexpr std::string_view on_actions_file = "common/on_actions.txt";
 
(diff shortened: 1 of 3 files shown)


src/openvic-simulation/misc/OnActions.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+bool OnActionManager::load_on_action_file(EventManager const& event_manager, ast::NodeCPtr root) {
+	bool ret = expect_dictionary([this, &event_manager](std::string_view identifier, ast::NodeCPtr node) -> bool {
+		OnAction::weight_map_t weighted_events;
+		bool ret = expect_dictionary(
+			[&identifier, &weighted_events, &event_manager](std::string_view weight_str, ast::NodeCPtr event_node) -> bool {
+				std::string_view event_id;
+				uint64_t weight = StringUtils::string_to_uint64(std::string(weight_str));
+				bool ret = expect_identifier_or_string(assign_variable_callback(event_id))(event_node);
```
Hop311, line 25:
```suggestion
				Event const* event = nullptr;
				bool ret = event_manager.expect_event_identifier(assign_variable_callback_pointer(event))(event_node);
```
The `Event const*` can be got directly from the node.

Hop311 (commented):
`OnAction`s could be stored and loaded by the `EventManager`, IMO they're not complex or distinct enough to need separate files and their own manager.

Hop311 (approved):

PR #95, opened 2023-12-11
Calculate province center

wvpm:
Calculate the province center and store it in positions.
Center  = sum x / pixel count , sum y / pixel count

### DIFF
--- a/src/openvic-simulation/map/Map.cpp
+++ b/src/openvic-simulation/map/Map.cpp
@@ -493,34 +493,41 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
 	std::vector<fixed_point_map_t<TerrainType const*>> terrain_type_pixels_list(provinces.size());
 	bool ret = true;
 	std::unordered_set<colour_t> unrecognised_province_colours;
+	
+	std::vector<fixed_point_t> pixels_per_province(provinces.size());
+	std::vector<fixed_point_t> x_sum_per_province(provinces.size());
+	std::vector<fixed_point_t> y_sum_per_province(provinces.size());
 
 	for (size_t y = 0; y < height; ++y) {
 		for (size_t x = 0; x < width; ++x) {
-			const size_t idx = x + y * width;
+			const size_t pixel_index = x + y * width;
+			const colour_t province_colour = colour_at(province_data, pixel_index);
 
-			const colour_t province_colour = colour_at(province_data, idx);
 			if (x > 0) {
-				const size_t jdx = idx - 1;
+				const size_t jdx = pixel_index - 1;
 				if (colour_at(province_data, jdx) == province_colour) {
-					province_shape_image[idx].index = province_shape_image[jdx].index;
+					province_shape_image[pixel_index].index = province_shape_image[jdx].index;
 					goto set_terrain;
 				}
 			}
+
 			if (y > 0) {
-				const size_t jdx = idx - width;
+				const size_t jdx = pixel_index - width;
 				if (colour_at(province_data, jdx) == province_colour) {
-					province_shape_image[idx].index = province_shape_image[jdx].index;
+					province_shape_image[pixel_index].index = province_shape_image[jdx].index;
 					goto set_terrain;
 				}
 			}
+
 			{
-				const Province::index_t index = get_index_from_colour(province_colour);
-				if (index != Province::NULL_INDEX) {
-					province_checklist[index - 1] = true;
-					province_shape_image[idx].index = index;
+				const Province::index_t province_index = get_index_from_colour(province_colour);
+				if (province_index != Province::NULL_INDEX) {
+					province_checklist[province_index - 1] = true;
+					province_shape_image[pixel_index].index = province_index;
 					goto set_terrain;
 				}
 			}
+
 			if (!unrecognised_province_colours.contains(province_colour)) {
 				unrecognised_province_colours.insert(province_colour);
 				if (detailed_errors) {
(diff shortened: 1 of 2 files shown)


src/openvic-simulation/map/Map.cpp
```diff
@@ -529,21 +536,28 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
...
+
 			province_shape_image[idx].index = Province::NULL_INDEX;
 
-		set_terrain:
-			const TerrainTypeMapping::index_t terrain = terrain_data[idx];
-			TerrainTypeMapping const* mapping = terrain_type_manager.get_terrain_type_mapping_for(terrain);
-			if (mapping != nullptr) {
-				if (province_shape_image[idx].index != Province::NULL_INDEX) {
-					terrain_type_pixels_list[province_shape_image[idx].index - 1][&mapping->get_type()]++;
+			set_terrain:
```
Hop311, line 542: `set_terrain:` is a label, which is meant to be one indent level lower than the surrounding code. Can you move everything form line 542 to line 560 back one indent level?

src/openvic-simulation/map/Map.cpp
```diff
@@ -529,21 +536,28 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
...
+				const Province::index_t province_index = get_index_from_colour(province_colour);
```
Hop311, line 543: You already have the province index `province_shape_image[idx].index`, no need to look up by colour again.

src/openvic-simulation/map/Map.cpp
```diff
@@ -529,21 +536,28 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
...
+				const uint16_t array_index = province_index-1;
```
Hop311, line 544: You should check the province index isn't 0 before doing this, otherwise you'll wrap around and get a value of 65535. If the province index is 0, there's no centre to calculate so the code can just continue on to the terrain part below.

src/openvic-simulation/map/Map.cpp
```diff
@@ -554,6 +568,17 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
 	size_t missing = 0;
 	for (size_t idx = 0; idx < province_checklist.size(); ++idx) {
 		Province* province = provinces.get_item_by_index(idx);
+
+		const uint16_t province_index = province->index;
+		if(province_index == Province::NULL_INDEX) {
+			continue;
+		}
+		const uint16_t array_index = province_index-1;
```
Hop311, line 576: This whole section isn't needed, `idx` is already the value you need (no subtraction necessary either).

src/openvic-simulation/map/Map.cpp
```diff
@@ -577,6 +602,7 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
 bool Map::_generate_province_adjacencies() {
 	bool changed = false;
 
+
```
Hop311, line 605: Could you remove the changes to this function, as it's not actually being changed in this PR (this newline + the `province_index` variable in the loop)

Hop311 (commented):
Great work! Some small comments, mostly about style or simplifying code, the only actual "bug" is not checking if the index is 0 before subtracting 1 from it, but other than that this code should all work 👍

src/openvic-simulation/map/Map.cpp
wvpm: I thought the code above checks for null index before entering set_terrain.

src/openvic-simulation/map/Map.cpp
wvpm: Yes. Not sure how this ended up changed.

src/openvic-simulation/map/Map.cpp
wvpm: You're right. Thanks

src/openvic-simulation/map/Map.cpp
wvpm: I changed it to use idx as you suggested in another comment. So no more -1.

Hop311: If checks for null index before jumping ahead to set_terrain, but if no index can be found for the colour it continues on to set the index to NULL_INDEX just above set_terrain, and then continues on to the code under set_terrain.

src/openvic-simulation/map/Map.cpp
wvpm: I highly doubt that. It was indented as if it was inside the for y loop, it's in the for x loop. It uses variables from the for x loop. So I indented it that way.

src/openvic-simulation/map/Map.cpp
wvpm: idx was all I needed, so I use that now.

src/openvic-simulation/map/Map.cpp
Hop311: The line containing `set_terrain` is a label, which goes one indent level lower than the code around it. All the code below stays at the same indent level as before.
With your changes, `set_terrain` is at the level of the x loop, and all the code below it is indented one level *more* than the x loop.

src/openvic-simulation/map/Map.cpp
```diff
@@ -529,21 +536,26 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
...
-		set_terrain:
-			const TerrainTypeMapping::index_t terrain = terrain_data[idx];
-			TerrainTypeMapping const* mapping = terrain_type_manager.get_terrain_type_mapping_for(terrain);
-			if (mapping != nullptr) {
-				if (province_shape_image[idx].index != Province::NULL_INDEX) {
-					terrain_type_pixels_list[province_shape_image[idx].index - 1][&mapping->get_type()]++;
+			set_terrain:
+				pixels_per_province[idx]++;
+				x_sum_per_province[idx] += x;
+				y_sum_per_province[idx] += y;
```
Hop311, line 545:
```suggestion
				if (province_shape_image[idx].index != Province::NULL_INDEX) {
					const size_t array_index = province_shape_image[idx].index - 1;
					pixels_per_province[array_index]++;
					x_sum_per_province[array_index] += x;
					y_sum_per_province[array_index] += y;
				}
```
`idx` is not the province index here (it's the pixel index), what you need is `province_shape_image[idx].index`, and also to check that it isn't zero since the `set_terrain` code is run regardless of whether a non-zero province index is found or not.

src/openvic-simulation/map/Map.cpp
```diff
@@ -587,7 +605,9 @@ bool Map::_generate_province_adjacencies() {
 
 	for (size_t y = 0; y < height; ++y) {
 		for (size_t x = 0; x < width; ++x) {
-			Province* cur = get_province_by_index(province_shape_image[x + y * width].index);
+			const uint16_t province_index = province_shape_image[x + y * width].index;
+			Province* cur = get_province_by_index(province_index);
+
```
Hop311, line 610: Can you revert this please as this PR doesn't affect this function.

src/openvic-simulation/map/Map.cpp
wvpm: That's very confused for me. Anyhow I unindented it.

src/openvic-simulation/map/Map.cpp
wvpm: I corrected this and renamed the variables to prevent future confusion.

src/openvic-simulation/map/Map.cpp
wvpm: Done

src/openvic-simulation/map/Map.cpp
```diff
@@ -493,34 +493,48 @@ bool Map::load_map_images(fs::path const& province_path, fs::path const& terrain
...
+	std::vector<fixed_point_t> x_sum_per_province(provinces.size());
+	std::vector<fixed_point_t> y_sum_per_province(provinces.size());
 
 	for (size_t y = 0; y < height; ++y) {
 		for (size_t x = 0; x < width; ++x) {
-			const size_t idx = x + y * width;
+			const size_t pixel_index = x + y * width;
+			const colour_t province_colour = colour_at(province_data, pixel_index);
+			
+			const Province::index_t province_index = get_index_from_colour(province_colour);
```
Hop311, line 506: `get_index_from_colour` requires searching through a big list of colours, which is an expensive operation. All of the code between here and `set_terrain:` is precisely to *avoid* needing to call `get_index_from_colour` unless absolutely necessary (by checking the provinces to the left and above in case they're the same colour, and so same index). Please move this back below `set_terrain` so it can use the efficiently calculated province index instead of forcing a costly lookup and making the `if (x > 0)` and `if (y > 0)` optimisations pointless.

src/openvic-simulation/map/Map.cpp
wvpm: The index may not have been set on the shape_image at that point.

Hop311 (approved):
All good, ready to merge 👍

PR #96, opened 2023-12-13
Calculate distance between adjacencies

wvpm:
Calculate distances for adjacencies using unit position with fallback to province center.

### DIFF
--- a/src/openvic-simulation/map/Province.cpp
+++ b/src/openvic-simulation/map/Province.cpp
@@ -121,7 +121,7 @@ Province::adjacency_t::adjacency_t(Province const* province, distance_t distance
 	assert(province != nullptr);
 }
 
-bool Province::is_adjacent_to(Province const* province) {
+bool Province::is_adjacent_to(Province const* province) const {
 	for (adjacency_t adj : adjacencies) {
 		if (adj.province == province) {
 			return true;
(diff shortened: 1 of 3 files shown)


src/openvic-simulation/map/Province.cpp
```diff
@@ -137,6 +137,18 @@ bool Province::add_adjacency(Province const* province, distance_t distance, flag
...
 
+fvec2_t Province::get_unit_position() const {
+	return positions.unit.value_or(positions.center);
+}
+
+Province::distance_t Province::calculate_distance_to(Province const* province) const {
+	const fvec2_t my_unit_position = get_unit_position();
+	const fvec2_t other_unit_position = province->get_unit_position();
+	const fvec2_t distance_vector = other_unit_position - my_unit_position;
+	const fixed_point_t distance = distance_vector.length_squared();
```
wvpm, line 148: This is distance squared. I need to math.sqrt it. We just need a sqrt for fixed point types.
https://github.com/MikeLankamp/fpm/blob/master/include/fpm/math.hpp is an option

Hop311: The square root function you linked is based on [this algorithm](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Binary_numeral_system_.28base_2.29), looks like a good candidate 👍
For now though let's leave this as the square of the length, so we can get this merged and I can continue my work on `additional-adjacencies` on top of it.

src/openvic-simulation/map/Province.cpp
```diff
@@ -137,6 +137,18 @@ bool Province::add_adjacency(Province const* province, distance_t distance, flag
 	return true;
 }
 
+fvec2_t Province::get_unit_position() const {
+	return positions.unit.value_or(positions.center);
```
Hop311, line 141: Nice, very elegant solution.

src/openvic-simulation/map/Province.cpp
```diff
@@ -137,6 +137,18 @@ bool Province::add_adjacency(Province const* province, distance_t distance, flag
...
+}
+
+Province::distance_t Province::calculate_distance_to(Province const* province) const {
+	const fvec2_t my_unit_position = get_unit_position();
+	const fvec2_t other_unit_position = province->get_unit_position();
+	const fvec2_t distance_vector = other_unit_position - my_unit_position;
+	const fixed_point_t distance = distance_vector.length_squared();
+	return static_cast<Province::distance_t>(distance.to_int64_t());
```
Hop311, line 149: Pleas change the `distance_t` alias in `Province.hpp` to `fixed_point_t`, and then remove the cast and `to_int64_t` from here.

Hop311 (commented):
Great work, just needs `distance_t` changed from an integer type to to `fixed_point_t`.

Hop311 (approved):

PR #97, opened 2023-12-13
Made pop type strata moddable

Hop311:


src/openvic-simulation/pop/Pop.cpp
```diff
@@ -193,3 +218,22 @@ bool PopManager::load_pop_into_vector(
...
+		STRATA_MODIFIER("_vote", true);
+
+		STRATA_MODIFIER("_life_needs", false);
+		STRATA_MODIFIER("_everyday_needs", false);
+		STRATA_MODIFIER("_luxury_needs", false);
+	}
+	return ret;
+}
+
+#undef STRATA_MODIFIER
```
Spartan322, line 239:
```suggestion
bool PopManager::generate_modifiers(ModifierManager& modifier_manager) {
	bool ret = true;
	
	auto strata_modifier = [&](std::string_view name, bool positive_good) {
		ret &= modifier_manager.add_modifier_effect(StringUtils::append_string_views(strata.get_identifier(), name), positive_good);
	};
	for (Strata const& strata : get_stratas()) {
		strata_modifier("_income_modifier", true);
		strata_modifier("_savings_modifier", true);
		strata_modifier("_vote", true);

		strata_modifier("_life_needs", false);
		strata_modifier("_everyday_needs", false);
		strata_modifier("_luxury_needs", false);
	}
	return ret;
}
```
could also instead do
```suggestion
bool PopManager::generate_modifiers(ModifierManager& modifier_manager) {
	bool ret = true;
	
	auto strata_modifier = [&](std::string_view name, bool positive_good) {
		return modifier_manager.add_modifier_effect(StringUtils::append_string_views(strata.get_identifier(), name), positive_good);
	};
	for (Strata const& strata : get_stratas()) {
		ret &= strata_modifier("_income_modifier", true);
		ret &= strata_modifier("_savings_modifier", true);
		ret &= strata_modifier("_vote", true);

		ret &= strata_modifier("_life_needs", false);
		ret &= strata_modifier("_everyday_needs", false);
		ret &= strata_modifier("_luxury_needs", false);
	}
	return ret;
}
```
but former is likely preferable anyway

Spartan322 (approved):

Spartan322:
could you squash?

Hop311:
I think the modifier macro changes are significant enough to be their own commit, I should've ammended the first commit and only had the other macro changes in a separate commit, but it's done now, my PC is off and I need to go to bed.

PR #100, opened 2023-12-19
Change colour_t to be a strongly typed structure

Spartan322:
Make RGB default of `colour_t`
Distinguish RGB and ARGB colors by type and colour_traits
Add `_colour` and `_argb` colour user-defined literals
Add `OpenVic::utility::unreachable`


src/openvic-simulation/map/Map.cpp
```diff
@@ -229,15 +230,15 @@ bool Map::generate_mapmode_colours(Mapmode::index_t index, uint8_t* target) cons
 		const colour_t base_colour = static_cast<colour_t>(base_stripe);
 		const colour_t stripe_colour = static_cast<colour_t>(base_stripe >> (sizeof(colour_t) * 8));
```
Hop311, line 231: This is what I was referring to earlier, the only place where `base_stripe_t` is deconstructed. Having the component colours as member variables would be cleaner, and wouldn't break anything else so long as `sizeof(base_stripe_t) == 2 * sizeof(colour_t) == 8` still holds true.

src/openvic-simulation/map/Region.cpp
```diff
@@ -65,7 +66,7 @@ ProvinceSet::provinces_t const& ProvinceSet::get_provinces() const {
 	return provinces;
 }
 
-static constexpr colour_t ERROR_REGION_COLOUR = COLOUR_COMPONENT << 16;
+static constexpr colour_t ERROR_REGION_COLOUR = colour_t::component << 16;
```
Hop311, line 69: We don't want a hardcoded bitshift here to make it red, best solution IMO would be to have static functions for red, green, blue and alpha that take in a component type variable and return a colour with just that component shifted to the right place.

src/openvic-simulation/types/Colour.hpp
```diff
@@ -1,42 +1,218 @@
...
+			return value;
+		}
+
+		static constexpr basic_colour_t from_float(float f, float min = 0.0f, float max = 1.0f) {
+			return static_cast<value_type>(std::clamp(min + f * (max - min), min, max) * 255.0f);
+		}
+
+		static constexpr basic_colour_t from_fraction(int n, int d, float min = 0.0f, float max = 1.0f) {
+			return from_float(static_cast<float>(n) / static_cast<float>(d), min, max);
+		}
```
Hop311, line 68: These create just a single component, could use the component return type?

src/openvic-simulation/types/Colour.hpp
```diff
@@ -1,42 +1,218 @@
...
+
+		static constexpr basic_colour_t from_float_to_alpha(float a) {
+			return from_float(a) << 24;
```
Hop311, line 71: Hardcoded shift value, could use the alpha variant of the component -> colour functions I mentioned earlier.

src/openvic-simulation/types/Colour.hpp
```diff
@@ -1,42 +1,218 @@
...
+		}
+
+		inline std::string to_hex_string(bool alpha = false) const {
+			using namespace std::string_view_literals;
+			static constexpr std::string_view digits = "0123456789ABCDEF"sv;
```
Hop311, line 76: Interesting, didn't know about these, would it be better to use them throughout the codebase for `static constexpr std::string_view`s? (e.g. all the file names in `Dataloader.cpp`)

src/openvic-simulation/types/Colour.hpp
```diff
@@ -1,42 +1,218 @@
...
+
+			const std::int8_t length = alpha ? 8 : 6;
+			std::array<char, 8> address;
+
+			for (std::size_t i = 0, j = (length - 1) * 4; i < length; ++i, j -= 4) {
+				address[i] = digits[(value >> j) & 0x0f];
+			}
+			return { address.data() };
```
Hop311, line 84: Very nice, love a good two-variable for loop. I assume `address` default zero-initialises its memory?

src/openvic-simulation/types/Colour.hpp
```diff
@@ -1,42 +1,218 @@
...
+			value &= rhs;
+			return *this;
+		}
+
+		friend constexpr basic_colour_t operator<<(value_type lhs, const basic_colour_t& rhs) {
+			return lhs << rhs.value;
+		}
+		friend constexpr basic_colour_t operator>>(value_type lhs, const basic_colour_t& rhs) {
+			return lhs >> rhs.value;
+		}
```
Hop311, line 147: Do we need shift operators with a `basic_colour_t` right hand side? I would consider someone storing a number of bits to shift in a colour variable to be a mistake/discouraged. Same applies to the `<<`/`>>` and `<<=`/`>>=` operators below.

src/openvic-simulation/types/Colour.hpp
```diff
@@ -1,42 +1,218 @@
...
+		 * should be allowed.
+		 */
+		static constexpr value_type null = traits.null;
+		static constexpr value_type component = traits.component;
+		static constexpr value_type max_rgb = traits.max_rgb;
+		static constexpr value_type max_argb = traits.max_argb;
 
-	constexpr float colour_byte_to_float(colour_t colour) {
-		return std::clamp(static_cast<float>(colour) / 255.0f, 0.0f, 1.0f);
-	}
```
Hop311, line 35: We need a variant of this for converting `colour_t` to `godot::Color` in the extension - maybe a static component type -> float conversion function, then non-static `redf`, `greenf`, `bluef`, `alphaf` getters using it?

src/openvic-simulation/map/Map.cpp
```diff
@@ -47,8 +48,8 @@ bool Map::add_province(std::string_view identifier, colour_t colour) {
 		);
 		return false;
 	}
-	if (colour == NULL_COLOUR || colour > MAX_COLOUR_RGB) {
-		Logger::error("Invalid province colour for ", identifier, ": ", colour_to_hex_string(colour));
+	if (colour == colour_t::null || colour > colour_t::max_rgb) {
```
Hop311, line 51:
```suggestion
	if (colour.is_null() || colour > colour_t::max_rgb) {
```

src/openvic-simulation/types/IdentifierRegistry.cpp
```diff
@@ -17,11 +17,11 @@ std::ostream& OpenVic::operator<<(std::ostream& stream, HasIdentifier const* obj
 }
 
 HasColour::HasColour(colour_t new_colour, bool cannot_be_null, bool can_have_alpha) : colour(new_colour) {
-	assert((!cannot_be_null || colour != NULL_COLOUR) && colour <= (!can_have_alpha ? MAX_COLOUR_RGB : MAX_COLOUR_ARGB));
+	assert((!cannot_be_null || colour != colour_t::null) && colour <= (!can_have_alpha ? colour_t::max_rgb : colour_t::max_argb));
```
Hop311, line 20:
```suggestion
	assert((!cannot_be_null || !colour.is_null()) && (can_have_alpha || colour <= colour_t::max_rgb));
```

The `max_argb` check isn't necessary since `colour_t` can't take greater values. Also a `has_alpha` function for `colour_t` would be useful, it could replace all the existing checks against `colour_t::max_rgb` and avoids potential confusion around which comparison operator to use.

Hop311 (commented):
Amazing work, this is so much nicer than the hacky integer alias version

src/openvic-simulation/map/Map.cpp
```diff
@@ -18,11 +19,11 @@ Mapmode::Mapmode(
 }
 
 const Mapmode Mapmode::ERROR_MAPMODE {
-	"mapmode_error", 0, [](Map const& map, Province const& province) -> colour_t { return 0xFFFF0000; }
+	"mapmode_error", 0, [](Map const& map, Province const& province) -> base_stripe_t { return 0xFFFF0000; }
```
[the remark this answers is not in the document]
Spartan322: This is the advantage of using strong types.

src/openvic-simulation/types/Colour.hpp
Spartan322: std::array default initializes its memory, since char is a integral type, its default value is 0.

src/openvic-simulation/types/Colour.hpp
Spartan322: It can be cleaner to do so, if you find it easier to read, yeah sure.

Hop311 (approved):
LGTM 👍 Only comments are on some stuff the formatter seems to have done, specifically getting rid of spaces after `requires` and putting everything after `requires` on the same line, even when it goes way past the 128 column limit. I haven't been using it much recently because the formatting I do myself is generally cleaner and more consistent, e.g. I prefer always breaking at the beginning of a bracketed list if there's a line break later in it, the formatter does that fine for `()` but for `{}` it insists on mixing tabs and spaces to get a perfect indent without a line break at the start.

Spartan322:
`sizeof(colour_t)` becomes 3 with this, BMP expects palettes to be `uint32_t`, (sizeof 4) causing the offset to be off and thus fail, corrected by #104.

PR #102, opened 2023-12-23
Fixed TGC compatibility + misc cleanup

Hop311:
Fixed TGC compatibility (now loads successfully), mostly adding missing modifier effects and changing the order in which some files are loaded.


src/openvic-simulation/economy/Good.cpp
```diff
@@ -106,9 +106,15 @@ bool GoodManager::generate_modifiers(ModifierManager& modifier_manager) const {
 		}
 	};
 
+	good_modifier("artisan_goods_input");
```
wvpm, line 109: Nice solution. This is all of them.

src/openvic-simulation/military/Unit.cpp
```diff
@@ -204,6 +204,8 @@ bool UnitManager::generate_modifiers(ModifierManager& modifier_manager) const {
 
 		ret &= modifier_manager.register_complex_modifier(identifier);
 
+		stat_modifier("_attack", true, RAW_DECIMAL);
```
wvpm, line 207: Do ships also use attack and defence modifiers?
I couldn't find an example of that in vanilla.

wvpm (approved):

src/openvic-simulation/military/Unit.cpp
Hop311: I didn't think they did, but i found it in some mods and when I tested it did show up in the invention tooltips. Also when you check the tooltip of a naval unit in the military menu, it lists their attack and defence stats. We should check if the modifier actually has an effect, but even if it doesn't we'll still want it registered as a valid key, just with a warning that it doesn't have an effect.

PR #103, opened 2023-12-25
Province building scaffolding

Hop311:
- Record province building types in a list, with their index in that list being used to identify them in the UI-SIM interface (eg when expanding buildings).
- Selected province is now stored as a province pointer rather than a province index.
- Date to double conversion removed, float is still used for expansion progress (as it has no impact on the simulation) but in general floating point use should be discouraged.

### DIFF
--- a/src/openvic-simulation/map/Province.cpp
+++ b/src/openvic-simulation/map/Province.cpp
@@ -58,9 +58,10 @@ bool Province::load_positions(BuildingTypeManager const& building_type_manager, 
 	return ret;
 }
 
-bool Province::expand_building(std::string_view building_type_identifier) {
-	BuildingInstance* building = buildings.get_item_by_identifier(building_type_identifier);
+bool Province::expand_building(size_t building_index) {
+	BuildingInstance* building = buildings.get_item_by_index(building_index);
 	if (building == nullptr) {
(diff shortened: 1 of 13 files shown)


Spartan322 (approved):
Aside from needing #104 to fix the bug introduced in #100, seems to work fine on Linux.

src/openvic-simulation/map/Province.cpp
```diff
@@ -58,9 +58,10 @@ bool Province::load_positions(BuildingTypeManager const& building_type_manager,
 	return ret;
 }
 
-bool Province::expand_building(std::string_view building_type_identifier) {
-	BuildingInstance* building = buildings.get_item_by_identifier(building_type_identifier);
+bool Province::expand_building(size_t building_index) {
```
schombert, line 61: in the same way you made color strongly typed, you might consider making this integer strongly typed

Hop311: I'm not planning on using integer indices to identify province buildings internally, just for communication with the UI which needs to use primitive types for the actual godot-cpp interface functions, so a stong type wouldn't have much use. `Province::index_t` would be a good candidate for making a strong type, especially since there are a couple other `index_t` type aliases with different sizes.

PR #106, opened 2023-12-26
Added political/policy rules

Hop311:
`Rule` and `RuleSet` operate similarly to `ModifierEffect` and `ModifierValue`, with `bool` values instead of `fixed_point_t`. `RuleSet`s can be combined using the overloaded `|` and `|=` operators, with conflicting rules defaulting to `true`/`yes`. `RuleSet`s can be loaded from `ast::NodeCPtr`s using `expect_rule_set`.

(NOT FULLY TESTED - will be used and tested in zaaarf's issue/reform loading PR)


zaaarf (approved):
Looks good, tested with issue/reform rules (will commit after this is merged). Approved.

PR #107, opened 2023-12-26
Dataloading Issue/Reform Modifiers and RuleSets

zaaarf:
Tested, seems to work fine. There can plausibly be some typos, but the change is relatively small so another pair of eyes can check that quite rapidly.

### DIFF
--- a/src/openvic-simulation/politics/Issue.cpp
+++ b/src/openvic-simulation/politics/Issue.cpp
@@ -5,16 +5,21 @@ using namespace OpenVic::NodeTools;
 
 IssueGroup::IssueGroup(std::string_view new_identifier) : HasIdentifier { new_identifier } {}
 
-Issue::Issue(std::string_view identifier, IssueGroup const& group) : HasIdentifier { identifier }, group { group } {}
+Issue::Issue(std::string_view new_identifier, ModifierValue&& new_values, IssueGroup const& new_group, RuleSet&& new_rules, bool new_jingoism)
+	: Modifier { new_identifier, std::move(new_values), 0 }, group { new_group }, rules { std::move(new_rules) },
+	jingoism { new_jingoism } {}
 
-ReformType::ReformType(std::string_view new_identifier, bool uncivilised)
-	: HasIdentifier { new_identifier }, uncivilised { uncivilised } {}
+ReformType::ReformType(std::string_view new_identifier, bool new_uncivilised)
+	: HasIdentifier { new_identifier }, uncivilised { new_uncivilised } {}
 
-ReformGroup::ReformGroup(std::string_view identifier, ReformType const& type, bool ordered, bool administrative)
-	: IssueGroup { identifier }, type { type }, ordered { ordered }, administrative { administrative } {}
+ReformGroup::ReformGroup(std::string_view new_identifier, ReformType const& new_type, bool new_ordered, bool new_administrative)
+	: IssueGroup { new_identifier }, type { new_type }, ordered { new_ordered }, administrative { new_administrative } {}
 
-Reform::Reform(std::string_view identifier, ReformGroup const& group, size_t ordinal)
-	: Issue { identifier, group }, ordinal { ordinal }, reform_group { group } {}
+Reform::Reform(
+	std::string_view new_identifier, ModifierValue&& new_values, ReformGroup const& new_group, size_t new_ordinal,
+	RuleSet&& new_rules, tech_cost_t new_technology_cost
+) : Issue { new_identifier, std::move(new_values), new_group, std::move(new_rules), false },
+	ordinal { new_ordinal }, reform_group { new_group }, technology_cost { new_technology_cost } {}
 
 bool IssueManager::add_issue_group(std::string_view identifier) {
 	if (identifier.empty()) {
@@ -25,7 +30,7 @@ bool IssueManager::add_issue_group(std::string_view identifier) {
 	return issue_groups.add_item({ identifier });
 }
 
-bool IssueManager::add_issue(std::string_view identifier, IssueGroup const* group) {
+bool IssueManager::add_issue(std::string_view identifier, ModifierValue&& values, IssueGroup const* group, RuleSet&& rules, bool jingoism) {
 	if (identifier.empty()) {
 		Logger::error("Invalid issue identifier - empty!");
 		return false;
@@ -36,7 +41,7 @@ bool IssueManager::add_issue(std::string_view identifier, IssueGroup const* grou
 		return false;
 	}
 
-	return issues.add_item({ identifier, *group });
+	return issues.add_item({ identifier, std::move(values), *group, std::move(rules), jingoism });
 }
 
 bool IssueManager::add_reform_type(std::string_view identifier, bool uncivilised) {
@@ -62,7 +67,10 @@ bool IssueManager::add_reform_group(std::string_view identifier, ReformType cons
 	return reform_groups.add_item({ identifier, *type, ordered, administrative });
 }
 
-bool IssueManager::add_reform(std::string_view identifier, ReformGroup const* group, size_t ordinal) {
+bool IssueManager::add_reform(
+	std::string_view identifier, ModifierValue&& values, ReformGroup const* group, size_t ordinal,
+	RuleSet&& rules, Reform::tech_cost_t technology_cost
+) {
 	if (identifier.empty()) {
 		Logger::error("Invalid issue identifier - empty!");
 		return false;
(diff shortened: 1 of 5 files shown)


src/openvic-simulation/politics/Issue.cpp
```diff
@@ -101,9 +119,22 @@ bool IssueManager::_load_reform_group(
 	return ret;
 }
 
-bool IssueManager::_load_reform(size_t& ordinal, std::string_view identifier, ReformGroup const* group, ast::NodeCPtr node) {
-	// TODO: conditions to allow, policy modifiers, policy rule changes
-	return add_reform(identifier, group, ordinal);
+bool IssueManager::_load_reform(
+	ModifierManager& modifier_manager, RuleManager& rule_manager, size_t& ordinal, std::string_view identifier,
```
Hop311, line 123: `ordinal` doesn't need to be a reference here, it can just be a regular `size_t`

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -5,16 +5,21 @@ using namespace OpenVic::NodeTools;
 
 IssueGroup::IssueGroup(std::string_view new_identifier) : HasIdentifier { new_identifier } {}
 
-Issue::Issue(std::string_view identifier, IssueGroup const& group) : HasIdentifier { identifier }, group { group } {}
+Issue::Issue(std::string_view new_identifier, ModifierValue&& new_values, IssueGroup const& new_group, RuleSet&& new_rules, bool new_jingoism)
```
Hop311, line 8:
```suggestion
Issue::Issue(std::string_view new_identifier, ModifierValue&& new_values, IssueGroup const& new_group, RuleSet&& new_rules, bool new_jingoism)
```
Trailing space character at the end of this line

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -5,16 +5,21 @@ using namespace OpenVic::NodeTools;
...
-Issue::Issue(std::string_view identifier, IssueGroup const& group) : HasIdentifier { identifier }, group { group } {}
+Issue::Issue(std::string_view new_identifier, ModifierValue&& new_values, IssueGroup const& new_group, RuleSet&& new_rules, bool new_jingoism) 
+	: Modifier { new_identifier, std::move(new_values), 0 }, group { new_group }, rules { std::move(new_rules) },
+	jingoism { new_jingoism } {}
 
 ReformType::ReformType(std::string_view new_identifier, bool uncivilised)
 	: HasIdentifier { new_identifier }, uncivilised { uncivilised } {}
 
 ReformGroup::ReformGroup(std::string_view identifier, ReformType const& type, bool ordered, bool administrative)
 	: IssueGroup { identifier }, type { type }, ordered { ordered }, administrative { administrative } {}
```
Hop311, line 16: Can you add a `new_` prefix to `ReformGroup`'s constructor arguments

wvpm:
I wanted to ask it on your [previous PR](https://github.com/OpenVicProject/OpenVic-Simulation/pull/106) but will ask it here:
What to do when multiple rules are true in the same group? For example if rich_only & state_vote are true.

Hop311:
> I wanted to ask it on your [previous PR](https://github.com/OpenVicProject/OpenVic-Simulation/pull/106) but will ask it here: What to do when multiple rules are true in the same group? For example if rich_only & state_vote are true.
> 
> I'd expect a warning if not an outright error. In vanilla it is allowed and the game can run with it. It just doesn't make sense as some rules are mutually exclusive. I tested with rich_only & same_as_ruling_party true. This resulting in same_as_ruling_party being used. When using rich_only & state_vote, state_vote is used.

Good point, I'll add mutually exclusive "rule groups" in another PR + warnings when conflicting rules are loaded. Another thing to consider is the default state of rules - are they all `no` when there's no explicit `yes` or `no` rule, even the economic ones?

src/openvic-simulation/politics/Issue.hpp
```diff
@@ -72,10 +74,13 @@ namespace OpenVic {
 	private:
 		ReformGroup const& PROPERTY(reform_group); // stores an already casted reference
 		const size_t PROPERTY(ordinal); // assigned by the parser to allow policy sorting
+		RuleSet PROPERTY(rules);
```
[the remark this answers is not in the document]
zaaarf: Right, I'd noticed that too but forgot to remove it it seems

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -73,19 +81,33 @@ bool IssueManager::add_reform(std::string_view identifier, ReformGroup const* gr
...
+	return reforms.add_item({ identifier, std::move(values), *group, ordinal, std::move(rules), technology_cost });
 }
 
-bool IssueManager::_load_issue_group(size_t& expected_issues, std::string_view identifier, ast::NodeCPtr node) {
+bool IssueManager::_load_issue_group(size_t expected_issues, std::string_view identifier, ast::NodeCPtr node) {
 	return expect_length([&expected_issues](size_t size) -> size_t {
 		expected_issues += size;
 		return size;
 	})(node) & add_issue_group(identifier);
 }
```
Hop311, line 96:
```suggestion
bool IssueManager::_load_issue_group(size_t& expected_issues, std::string_view identifier, ast::NodeCPtr node) {
	return expect_length(add_variable_callback(expected_issues))(node) & add_issue_group(identifier);
}
```
`expected_issues` needs to be a reference and the lambda function inside `expect_length` can be replaced with `add_variable_callback`.

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -101,9 +123,22 @@ bool IssueManager::_load_reform_group(
...
-	// TODO: conditions to allow, policy modifiers, policy rule changes
-	return add_reform(identifier, group, ordinal);
+bool IssueManager::_load_reform(
+	ModifierManager const& modifier_manager, RuleManager const& rule_manager, size_t ordinal, std::string_view identifier,
+	ReformGroup const* group, ast::NodeCPtr node
+) {
+	// TODO: conditions to allow, policy rule changes
+	ModifierValue values;
+	RuleSet rules;
+	Reform::tech_cost_t technology_cost;
```
Hop311, line 133:
```suggestion
	Reform::tech_cost_t technology_cost = 0;
```
Default initialise, otherwise on MSVC it will contain random data.

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -157,22 +190,22 @@ bool IssueManager::load_issues_file(ast::NodeCPtr root) {
...
-		return expect_dictionary([this, type_key](std::string_view group_key, ast::NodeCPtr group_value) -> bool {
+	ret &= expect_dictionary([this, &modifier_manager, &rule_manager](std::string_view type_key, ast::NodeCPtr type_value) -> bool {
+		return expect_dictionary([this, &modifier_manager, &rule_manager, type_key](std::string_view group_key, ast::NodeCPtr group_value) -> bool {
 			if (type_key == "party_issues") {
 				IssueGroup const* issue_group = get_issue_group_by_identifier(group_key);
-				return expect_dictionary([this, issue_group](std::string_view key, ast::NodeCPtr value) -> bool {
-					bool ret = _load_issue(key, issue_group, value);
+				return expect_dictionary([this, &modifier_manager, &rule_manager, issue_group](std::string_view key, ast::NodeCPtr value) -> bool {
+					bool ret = _load_issue(modifier_manager, rule_manager, key, issue_group, value);
 					return ret;
```
Hop311, line 199:
```suggestion
					return _load_issue(modifier_manager, rule_manager, key, issue_group, value);
```

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -157,22 +190,22 @@ bool IssueManager::load_issues_file(ast::NodeCPtr root) {
...
 				size_t ordinal = 0;
-				return expect_dictionary([this, reform_group, &ordinal](std::string_view key, ast::NodeCPtr value) -> bool {
+				return expect_dictionary([this, &modifier_manager, &rule_manager, reform_group, &ordinal](std::string_view key, ast::NodeCPtr value) -> bool {
 					if (key == "next_step_only" || key == "administrative") {
 						return true;
 					}
-					bool ret = _load_reform(ordinal, key, reform_group, value);
+					bool ret = _load_reform(modifier_manager, rule_manager, ordinal, key, reform_group, value);
 					ordinal++;
 					return ret;
```
Hop311, line 210:
```suggestion
					return _load_reform(modifier_manager, rule_manager, ordinal++, key, reform_group, value);
```
This will increase `ordinal` by 1, but use the value before incrememting as the function argument (as opposed to `++ordinal`, which would use the incremented value as the argument).

Hop311 (commented):
A couple comments, mostly stuff I didn't notice the first time I reviewed. This won't work 100% correctly with mods right now because `Rules` don't support things like `build_xxx` for custom factories. Still, `Issue.*pp` shouldn't need any changes for that, just `Rule.*pp`, which I'll deal with in the next couple days.

src/openvic-simulation/politics/Issue.cpp
zaaarf: Will do, I knew about this but I tend to avoid these things for the sake of clarity.

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -73,19 +81,31 @@ bool IssueManager::add_reform(std::string_view identifier, ReformGroup const* gr
 		return false;
 	}
 
-	return reforms.add_item({ identifier, *group, ordinal });
+	if (!group->get_type().is_uncivilised() && technology_cost != 0) {
+		Logger::warning("Technology cost ", technology_cost, " found in civilised reform ", identifier, "!");
+	}
```
Hop311, line 86:
```suggestion
	if (group->get_type().is_uncivilised()) {
		if (technology_cost <= 0) {
			Logger::warning("Non-positive technology cost ", technology_cost, " found in uncivilised reform ", identifier, "!");
		}
	} else if (technology_cost != 0) {
		Logger::warning("Non-zero technology cost ", technology_cost, " found in civilised reform ", identifier, "!");
	}
```
Sorry I should've been clearer, I meant change the message, we want a warning if there's a non-zero tech cost for a civilised one OR if there's a zero tech cost for an uncivilised one (since 0 tech cost will likely mean the tech cost key was missing from the reform).

Hop311 (approved):

PR #108, opened 2023-12-26
Dataloading climate and continents

zaaarf:
The `Climate` and `Continent` structs would've been perfectly identical, so I created one `ProvinceSetModifier` struct with `Climate`and `Continent` as aliases. They are both a list of provinces associated with some modifiers, so they should each be a `ProvinceSet`, and conversely the provinces should have references to their `Climate` and `Continent`. Both references - the one stored in the `Climate`/`Continent` and the one(s) stored in `Province` - should be stored as `const` pointers.

~~To ensure that the reference to `Climate` and `Continent` each `Province` holds is immutable, I had to make that annoying for loop at the end. It's only ran once, so the double loop isn't too much a performance concern.~~

~~It's maybe worth looking into making a variant of `ProvinceSet` holding `Province const*` instead, but I'll wait on hearing from someone else before embarking in that.~~

### DIFF
--- a/src/openvic-simulation/map/Map.cpp
+++ b/src/openvic-simulation/map/Map.cpp
@@ -1,7 +1,9 @@
 #include "Map.hpp"
 
 #include <cassert>
+#include <cstddef>
 #include <unordered_set>
+#include <vector>
 
 #include "openvic-simulation/economy/Good.hpp"
 #include "openvic-simulation/history/ProvinceHistory.hpp"
@@ -642,5 +644,93 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const& 
 			ret &= Province::add_special_adjacency(*from, *to, type, through, data);
 		}
 	);
+	return ret;
+}
+
+bool Map::load_climate_file(ModifierManager const& modifier_manager, ast::NodeCPtr root) {
+	bool ret = expect_dictionary_reserve_length(climates, [this, &modifier_manager](std::string_view identifier, ast::NodeCPtr node) -> bool {
+		if (identifier.empty()) {
+			Logger::error("Invalid climate identifier - empty!");
+			return false;
+		}
+
+		bool ret = true;
+		Climate* cur_climate = climates.get_item_by_identifier(identifier);
+		if (cur_climate == nullptr) {
+			ModifierValue values;
+			ret &= modifier_manager.expect_modifier_value(move_variable_callback(values))(node);
+			ret &= climates.add_item({ identifier, std::move(values) });
+		} else {
+			ret &= expect_list_reserve_length(*cur_climate, expect_province_identifier(
+				[cur_climate, &identifier](Province& province) {
+					if (province.climate != cur_climate) {
+						cur_climate->add_province(&province);
+						if (province.climate != nullptr) {
+							Climate* old_climate = const_cast<Climate*>(province.climate);
+							old_climate->remove_province(&province);
+							Logger::warning(
+								"Province with id ", province.get_identifier(),
+								" found in multiple climates: ", identifier,
+								" and ", old_climate->get_identifier()
+							);
+						}
+						province.climate = cur_climate;
+					} else {
+						Logger::warning(
+							"Province with id ", province.get_identifier(),
+							"defined twice in climate ", identifier
+						);
+					}
+					return true;
+				}
+			))(node);
+			cur_climate->lock();
+		}
+		return ret;
+	})(root);
+
+	lock_climates();
+
+	return ret;
+}
+
+bool Map::load_continent_file(ModifierManager const& modifier_manager, ast::NodeCPtr root) {
+	bool ret = expect_dictionary_reserve_length(continents, [this, &modifier_manager](std::string_view identifier, ast::NodeCPtr node) -> bool {
+		if (identifier.empty()) {
+			Logger::error("Invalid continent identifier - empty!");
+			return false;
+		}
+
+		ModifierValue values;
+		ProvinceSetModifier::provinces_t prov_list;
+		bool ret = modifier_manager.expect_modifier_value_and_keys(move_variable_callback(values),
+			"provinces", ONE_EXACTLY, expect_list_reserve_length(prov_list, expect_province_identifier(
+				[&prov_list](Province const& province) -> bool {
+					if (province.continent == nullptr) {
+						prov_list.emplace_back(&province);
+					}
+					return true;
+				}
+			))
+		)(node);
+
+		Continent continent = { identifier, std::move(values) };
+		continent.add_provinces(prov_list);
+		continent.lock();
+
+		if (continents.add_item(std::move(continent))) {
+			Continent const& moved_continent = continents.get_items().back();
+			for (Province const* prov : moved_continent.get_provinces()) {
+				remove_province_const(prov)->continent = &moved_continent;
+			}
+		} else {
+			ret = false;
+		}
+
+		return ret;
+	})(root);
+
+	lock_continents();
+
 	return ret;
 }
--- a/src/openvic-simulation/map/Map.hpp
+++ b/src/openvic-simulation/map/Map.hpp
@@ -49,7 +49,6 @@ namespace OpenVic {
 	 * MAP-4
 	 */
 	struct Map {
-
 #pragma pack(push, 1)
 		/* Used to represent tightly packed 3-byte integer pixel information. */
 		struct shape_pixel_t {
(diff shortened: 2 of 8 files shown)


wvpm (approved):
Looks good

src/openvic-simulation/map/Map.cpp
```diff
@@ -670,3 +670,82 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const&
...
+				Province* cur_province = provinces.get_item_by_identifier(cur_province_id);
+				if (cur_province != nullptr) {
+					cur_climate->add_province(cur_province);
+					return ret;
+				} else {
+					Logger::error("Invalid province id ", cur_province_id, " in definition of climate ", identifier, "!");
+					return false;
+				}
+			})(node);
+			return ret;
```
Hop311, line 700:
```suggestion
			return expect_list(expect_province_identifier(
				[cur_climate](Province const& province) {
					cur_climate->add_province(&province));
					return true;
				}
			))(node);
```
`expect_province_identifier` can be put directly into `expect_list`, since it returns a `node_callback_t` and `expect_list` takes it as its argument. `add_province` will deal with any duplicates *in the same climate*, returning false and printing warning messages, but we don't want to return that false value because it will mark `load_climate_file` as failing, when this should only be considered a minor issue that doesn't affect loading the rest of the climate.

This does not address the issue of duplicates across multiple climates, for that I was thinking you could record each province read from a climate list in a single `std::set<Province const*>`, checking if it `contains` each province you find in each climate list and if so not adding the province to that climate, otherwise adding it to the climate and the std::set `std::set<Province const*>`. This would require the lists to be read in reverse order, so that later lists have priority over earlier ones.

src/openvic-simulation/map/Map.cpp
```diff
@@ -670,3 +670,82 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const&
...
+		}
+	})(root);
+
+	lock_climates();
+
+	for (Climate const& cl : climates.get_items()) {
+		for(Province* prov : cl.get_provinces()) {
+			prov->climate = &cl;
+		}
+	}
```
Hop311, line 710: With my changes in #109, this can be changed to:
```suggestion
	for (Climate const& climate : climates.get_items()) {
		if (climate.empty()) {
			Logger::warning("Climate ", climate.get_identifier(), " has no provinces!");
		} else {
			for(Province const* prov : climate.get_provinces()) {
				remove_province_const(prov)->climate = &climate;
			}
		}
	}
```

src/openvic-simulation/map/Map.cpp
```diff
@@ -670,3 +670,82 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const&
...
+
+	return ret;
+}
+
+bool Map::load_continent_file(ModifierManager const& modifier_manager, ast::NodeCPtr root) {
+	bool ret = expect_dictionary([this, &modifier_manager](std::string_view identifier, ast::NodeCPtr node) -> bool {
```
Hop311, line 716:
```suggestion
	bool ret = expect_dictionary_reserve_length(continents, [this, &modifier_manager](std::string_view identifier, ast::NodeCPtr node) -> bool {
```
Since each entry is a new continent, we can reserve space in `continents` using the number of entries in the file.

src/openvic-simulation/map/Map.cpp
```diff
@@ -670,3 +670,82 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const&
...
+		return ret;
+	})(root);
+
+	lock_continents();
+
+	for (Continent const& con : continents.get_items()) {
+		for(Province* prov : con.get_provinces()) {
+			prov->continent = &con;
+		}
+	}
```
Hop311, line 748: With my changes in #109, this can be changed to:
```suggestion
	for (Continent const& continent : continents.get_items()) {
		if (continent.empty()) {
			Logger::warning("Continent ", continent.get_identifier(), " has no provinces!");
		} else {
			for(Province const* prov : continent.get_provinces()) {
				remove_province_const(prov)->continent = &continent;
			}
		}
	}
```

src/openvic-simulation/map/Map.cpp
```diff
@@ -670,3 +670,82 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const&
...
+					prov.emplace_back(cur_province);
+					return ret;
+				} else {
+					Logger::error("Invalid province id ", cur_province_id, " in definition of continent ", identifier, "!");
+					return false;
+				}
+			})
+		)(node);
+		ret &= continents.add_item({ identifier, std::move(values), std::move(prov) });
+		return ret;
```
Hop311, line 739: The `expect_list` here can use `expect_province_identifier` similar to what I suggested for the climate loading function, the only difference is that here you don't already have a `Continent` object to `add_province` to, but with the changes in #109, you can just construct a `Continent` once you've got the `ModifierValue` loaded and use `add_provinces` on it at the end, so line 738 here becomes:
```c++
		Continent continent { identifier, std::move(values) };
		ret &= continent.add_provinces(prov);
		ret &= continents.add_item(std::move(continent));		
```

There are no duplicate provinces across continents like there are for climates in vanilla, probably because continents are fixed geographical things whereas climates are fuzzier and get changed over time. This should still use a `std::set<Province const*>` to check for duplicates across different continents and not add them if they already exist in another continent, but IMO it's fine to let the first continent a province is defined in be the one that matters, and ignore later ones, so no reverse parsing needed.

Hop311 (commented):
Left a bunch of comments, especially in the 2 loading functions, if anything's unclear please ask, this stuff is difficult to put into words concisely

src/openvic-simulation/map/Map.hpp
```diff
@@ -102,6 +103,9 @@ namespace OpenVic {
 
 		bool add_mapmode(std::string_view identifier, Mapmode::colour_func_t colour_func);
 
+		bool add_climate(std::string_view identifier, ModifierValue&& values);
+		bool add_continent(std::string_view identifier, ModifierValue&& values);
```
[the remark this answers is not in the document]
zaaarf: Oh, right - I was originally going to have them be in their own functions, but for a variety of reasons I figured it wasn't worth it (also, this way should be slightly more efficient since we don't waste time processing otherwise invalid identifiers).

src/openvic-simulation/map/Map.cpp
zaaarf: Maybe I'm wrong, but this code doesn't seem right. for-each on empty sets just never iterate, so that warning would never be printed.

src/openvic-simulation/map/Map.cpp
```diff
@@ -644,3 +646,89 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const&
...
+		}
+
+		bool ret = true;
+		Climate* cur_climate = climates.get_item_by_identifier(identifier);
+		if (cur_climate == nullptr) {
+			ModifierValue values;
+			ret &= modifier_manager.expect_modifier_value(move_variable_callback(values))(node);
+			ret &= climates.add_item({ identifier, std::move(values) });
+		} else {
+			ret &= expect_list(expect_province_identifier([this, &cur_climate, &identifier](Province& province) {
```
Hop311, line 664:
```suggestion
			ret &= expect_list_reserve_length(*cur_climate, expect_province_identifier([cur_climate, &identifier](Province& province) {
```
`Climate` works with `expect_list_reserve_length`, since it inherits a `reserve` function from `ProvinceSet`. Also `this` isn't needed in the lambda capture, and `cur_climate` is better off being captured by value rather than reference, since it's already a pointer so there's no need to reference to it (which is basically using a pointer to pointer) when you can just copy it and point directly to the climate.

src/openvic-simulation/map/Map.cpp
```diff
@@ -644,3 +646,89 @@ bool Map::generate_and_load_province_adjacencies(std::vector<LineObject> const&
...
+		if (identifier.empty()) {
+			Logger::error("Invalid continent identifier - empty!");
+			return false;
+		}
+
+		ModifierValue values;
+		ProvinceSetModifier::provinces_t prov_list;
+		bool ret = modifier_manager.expect_modifier_value_and_keys(move_variable_callback(values),
+			"provinces", ONE_EXACTLY, expect_list(expect_province_identifier(
+				[this, &prov_list, &identifier](Province const& province) -> bool {
```
Hop311, line 706:
```suggestion
			"provinces", ONE_EXACTLY, expect_list_reserve_length(prov_list, expect_province_identifier(
				[&prov_list](Province const& province) -> bool {
```
Again, `expect_list_reserve_length` can be used, this time reserving space in a `std::vector` but it still works the same. Neither `this` nor `&identifier` are needed in the lambda capture.

Hop311 (commented):
Great work, this turned out much trickier than expected, well done 👍

Hop311 (approved):

PR #111, opened 2023-12-29
Condition definition checking/loading

zaaarf:
This one's a mess. 

We have `Condition`s, representing any "valid node" in condition blocks (which includes both condition scopes and proper conditions). This behaviour may be changed in the future. `Condition` stores information about that "type of node" - what value it can take, whether it's an identifier of something and of what if it is, and so on.

Then you have `ConditionNode`s, which store a `Condition` and its value, (which may also be a list of `ConditionNode`s, forming a tree-like structure).

Following up on the skeleton implemented in #112, blocks requiring conditions of some kind will store a `ConditionScript`, which will contain an empty `ConditionNode`, a `NodeCPtr` storing a reference to the tree node to parse and, if needed, metadata (such as the initial scope, `factor` values if present, and so on).

The `ConditionManager` provides the `expect` callbacks that actually get called when, in the second pass once everything else is loaded, we actually parse the nodes.

### On the dumping of other identifiers into the `IdentifierRegistry`
It's not elegant, but I don't really think there's any good alternative to it. We'll *always* need a `Condition const*` to store in the `ConditionNode` for this approach to make any sense. If we don't have a corresponding condition to every, say, ideology, we'll need to craft an ad-hoc one every time, but then that would go out of scope and be destroyed, and our pointer would die on us.

I suppose there may be the alternative of having `ConditionNode` hold an `std::variant` with both `Condition` and other types that can be used as conditions, but I don't especially like it since it requires me to hardcode special handling, while this other approach "just works". An upside to this is that we could then have a separate `Scope` struct - though I feel that's a bit superfluous.

### DIFF
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -948,68 +948,40 @@ bool Dataloader::load_defines(GameManager& game_manager) {
 		Logger::error("Failed to load diplomatic actions!");
 		ret = false;
 	}
+	if(!game_manager.get_script_manager().get_condition_manager().setup_conditions(game_manager)) {
+		Logger::error("Failed to set up conditions!");
+		ret = false;
+	}
 
-	parse_scripts(game_manager);
+	ret &= parse_scripts(game_manager);
 
 	free_cache();
 
 	return ret;
 }
 
+#define PARSE_SCRIPTS(name, mgr) \
+	if (!mgr.parse_scripts(game_manager)) { Logger::error("Failed to parse ", name, " scripts!"); ret = false; } \
+	else Logger::info("Successfully parsed ", name, " scripts!");
+
 bool Dataloader::parse_scripts(GameManager& game_manager) const {
 	bool ret = true;
-
-	if (!game_manager.get_pop_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse pop scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_politics_manager().get_ideology_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse ideology scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_politics_manager().get_issue_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse reform scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_economy_manager().get_production_type_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse production type scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_politics_manager().get_rebel_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse rebel type scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_research_manager().get_technology_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse technology scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_crime_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse crime scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_modifier_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse triggered modifier scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_research_manager().get_invention_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse invention scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_military_manager().get_wargoal_type_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse wargoal type scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_decision_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse decision scripts!");
-		ret = false;
-	}
-	if (!game_manager.get_event_manager().parse_scripts(game_manager)) {
-		Logger::error("Failed to parse event scripts!");
-		ret = false;
-	}
-
+	PARSE_SCRIPTS("pop", game_manager.get_pop_manager());
+	PARSE_SCRIPTS("ideology", game_manager.get_politics_manager().get_ideology_manager());
+	PARSE_SCRIPTS("reform", game_manager.get_politics_manager().get_issue_manager());
+	PARSE_SCRIPTS("production type", game_manager.get_economy_manager().get_production_type_manager());
+	PARSE_SCRIPTS("rebel type", game_manager.get_politics_manager().get_rebel_manager());
+	PARSE_SCRIPTS("technology", game_manager.get_research_manager().get_technology_manager());
+	PARSE_SCRIPTS("crime", game_manager.get_crime_manager());
+	PARSE_SCRIPTS("triggered modifier", game_manager.get_modifier_manager());
+	PARSE_SCRIPTS("invention", game_manager.get_research_manager().get_invention_manager());
+	PARSE_SCRIPTS("wargoal type", game_manager.get_military_manager().get_wargoal_type_manager());
+	PARSE_SCRIPTS("decision", game_manager.get_decision_manager());
+	PARSE_SCRIPTS("event", game_manager.get_event_manager());
 	return ret;
 }
+
+#undef PARSE_SCRIPTS
 
 static bool _load_localisation_file(Dataloader::localisation_callback_t callback, std::vector<csv::LineObject> const& lines) {
 	bool ret = true;
--- a/src/openvic-simulation/dataloader/NodeTools.cpp
+++ b/src/openvic-simulation/dataloader/NodeTools.cpp
@@ -70,13 +70,17 @@ node_callback_t NodeTools::expect_identifier_or_string(callback_t<std::string_vi
 }
 
 node_callback_t NodeTools::expect_bool(callback_t<bool> callback) {
-	static const string_map_t<bool> bool_map = { { "yes", true }, { "no", false } };
+	static const case_insensitive_string_map_t<bool> bool_map = { { "yes", true }, { "no", false } };
 	return expect_identifier(expect_mapped_string(bool_map, callback));
 }
 
 node_callback_t NodeTools::expect_int_bool(callback_t<bool> callback) {
-	static const string_map_t<bool> bool_map = { { "1", true }, { "0", false } };
-	return expect_identifier(expect_mapped_string(bool_map, callback));
+	return expect_uint64([callback](uint64_t num) -> bool {
+		if (num > 1) {
+			Logger::warning("Found int bool with value >1: ", num);
+		}
+		return callback(num != 0);
+	});
 }
 
 node_callback_t NodeTools::expect_int64(callback_t<int64_t> callback, int base) {
--- a/src/openvic-simulation/misc/Event.cpp
+++ b/src/openvic-simulation/misc/Event.cpp
@@ -6,8 +6,8 @@
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
-Event::EventOption::EventOption(std::string_view new_title, EffectScript&& new_effect, ConditionalWeight&& new_ai_chance)
-  : title { new_title }, effect { std::move(new_effect) }, ai_chance { std::move(new_ai_chance) } {}
+Event::EventOption::EventOption(std::string_view new_name, EffectScript&& new_effect, ConditionalWeight&& new_ai_chance)
+  : name { new_name }, effect { std::move(new_effect) }, ai_chance { std::move(new_ai_chance) } {}
 
 bool Event::EventOption::parse_scripts(GameManager& game_manager) {
 	bool ret = true;
@@ -86,6 +86,17 @@ bool EventManager::register_event(
 		}
 	}
 
+	if (options.empty()) {
+		Logger::error("Event with ID ", identifier, " has no options!");
+		return false;
+	} else {
+		for (Event::EventOption const& option : options) {
+			if (option.name.empty()) {
+				Logger::warning("Event with ID ", identifier, " has an option with no name!");
+			}
+		}
+	}
+
 	// TODO - error if is_triggered_only with triggers or MTTH defined
 
 	return events.add_item({
@@ -108,24 +119,28 @@ bool EventManager::load_event_file(IssueManager const& issue_manager, ast::NodeC
 	return expect_dictionary(
 		[this, &issue_manager](std::string_view key, ast::NodeCPtr value) -> bool {
 			Event::event_type_t type;
+			scope_t initial_scope;
+
+			if (key == "country_event") {
+				type = Event::event_type_t::COUNTRY;
+				initial_scope = scope_t::COUNTRY;
+			} else if (key == "province_event") {
+				type = Event::event_type_t::PROVINCE;
+				initial_scope = scope_t::PROVINCE;
+			} else {
+				Logger::error("Invalid event type: ", key);
+				return false;
+			}
+
 			std::string_view identifier, title, description, image, news_title, news_desc_long, news_desc_medium,
 				news_desc_short;
 			bool triggered_only = false, major = false, fire_only_once = false, allows_multiple_instances = false,
 				news = false, election = false;
 			IssueGroup const* election_issue_group = nullptr;
-			ConditionScript trigger;
-			ConditionalWeight mean_time_to_happen;
+			ConditionScript trigger { initial_scope, initial_scope, scope_t::NO_SCOPE };
+			ConditionalWeight mean_time_to_happen { initial_scope, initial_scope, scope_t::NO_SCOPE };
 			EffectScript immediate;
 			std::vector<Event::EventOption> options;
-
-			if (key == "country_event") {
-				type = Event::event_type_t::COUNTRY;
-			} else if (key == "province_event") {
-				type = Event::event_type_t::PROVINCE;
-			} else {
-				Logger::error("Invalid event type: ", key);
-				return false;
-			}
 
 			bool ret = expect_dictionary_keys(
 				"id", ONE_EXACTLY, expect_identifier(assign_variable_callback(identifier)),
@@ -144,20 +159,24 @@ bool EventManager::load_event_file(IssueManager const& issue_manager, ast::NodeC
 				"election", ZERO_OR_ONE, expect_bool(assign_variable_callback(election)),
 				"issue_group", ZERO_OR_ONE,
 					issue_manager.expect_issue_group_identifier(assign_variable_callback_pointer(election_issue_group)),
-				"option", ONE_OR_MORE, [&options](ast::NodeCPtr node) -> bool {
-					std::string_view title;
+				"option", ONE_OR_MORE, [&options, initial_scope](ast::NodeCPtr node) -> bool {
+					std::string_view name;
 					EffectScript effect;
-					ConditionalWeight ai_chance;
+					ConditionalWeight ai_chance {
+						initial_scope,
+						initial_scope,
+						scope_t::COUNTRY | scope_t::PROVINCE
+					};
 
 					bool ret = expect_dictionary_keys_and_default(
 						key_value_success_callback,
-						"name", ONE_EXACTLY, expect_identifier_or_string(assign_variable_callback(title)),
+						"name", ONE_EXACTLY, expect_identifier_or_string(assign_variable_callback(name)),
 						"ai_chance", ZERO_OR_ONE, ai_chance.expect_conditional_weight(ConditionalWeight::FACTOR)
 					)(node);
 
 					ret &= effect.expect_script()(node);
 
-					options.push_back({ title, std::move(effect), std::move(ai_chance) });
+					options.push_back({ name, std::move(effect), std::move(ai_chance) });
 					return ret;
 				},
 				"trigger", ZERO_OR_ONE, trigger.expect_script(),
--- a/src/openvic-simulation/pop/Pop.cpp
+++ b/src/openvic-simulation/pop/Pop.cpp
@@ -1,14 +1,8 @@
 #include "Pop.hpp"
 
-#include <cassert>
-
-#include "openvic-simulation/dataloader/NodeTools.hpp"
-#include "openvic-simulation/map/Province.hpp"
 #include "openvic-simulation/politics/Ideology.hpp"
 #include "openvic-simulation/politics/Issue.hpp"
 #include "openvic-simulation/politics/Rebel.hpp"
-#include "openvic-simulation/types/Colour.hpp"
-#include "openvic-simulation/utility/Logger.hpp"
 #include "openvic-simulation/utility/TslHelper.hpp"
 
 using namespace OpenVic;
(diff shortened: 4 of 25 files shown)


src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,318 @@
...
+#include <vector>
+
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
+
+Condition::Condition(std::string_view new_identifier, value_type_t new_value_type, std::set<scope_t>&& new_scopes)
+	: HasIdentifier { new_identifier }, value_type { new_value_type }, scopes { new_scopes } {}
+
+ConditionNode::ConditionNode(Condition const* new_condition, value_t&& new_value)
+	: condition { new_condition }, value { new_value } {}
```
Hop311, line 11:
```suggestion
	: condition { new_condition }, value { std::move(new_value) } {}
```

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -0,0 +1,81 @@
...
+		enum class scope_t {
+			GLOBAL,
+			COUNTRY,
+			PROVINCE,
+			POP
+		};
+
+	private:
+		const value_type_t PROPERTY(value_type);
+		std::set<scope_t> PROPERTY(scopes);
```
Hop311, line 34: You could use a bitfield for this, so:
```c++
enum scope_t : uint8_t {
	COUNTRY  = 1 << 0, // 0b001
	PROVINCE = 1 << 1, // 0b010
	POP      = 1 << 2, // 0b100
	GLOBAL   = COUNTRY | PROVINCE | POP // 0b111
}
...
scope_t PROPERTY(scopes);
```
It would be faster to modify and to check what scopes are enabled, e.g. `(scopes & new_scope) == new_scope` checks `scopes` contains all of `new_scope`, you could also just check the result is non-zero but that wouldn't work for checking if it's `GLOBAL` or a combination of 2. Also it would treat `GLOBAL` as the same as adding all 3 others individually, and make it so you just have to check `scopes | PROVINCE` instead of `scopes.contains(PROVINCE) || scopes.contains(GLOBAL)`. `class` is removed from the enum declaration because otherwise you can't use bitwise operators on it, e.g. `scope |= scope_t::PROVINCE;` (with an `enum class` you'd have to `static_cast` to `uint8_t` every time you want to use `&` or `|`).

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+	// ConditionNode: represents a node, which is a map of individual conditions associated with a value, tied by
+	// a logical relationship. Simple conditions are considered to be single-element ANDs.
+	struct ConditionNode {
+		using string_t = std::string;
+		using boolean_t = bool;
+		using integer_t = int32_t;
+		using real_t = fixed_point_t;
+		//TODO add complex types
+		using condition_set_t = std::set<ConditionNode>;
```
Hop311, line 50:
```suggestion
		using condition_list_t = std::vector<ConditionNode>;
```
We need to preserve the order of conditions in the defines so they show up properly in tooltips, plus there's no benefit to using a `std::set` since there aren't any duplicates we're worried about. This should also fix (some of) the Mac compile errors, which are to do with `std::set` needing to be able to do `<` less than comparisons on its elements.

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -0,0 +1,81 @@
...
+	public:
+		template<typename T> //TODO replace this with an expect
+		constexpr auto get_type_expect_callback(Condition::value_type_t t) {
+			switch (t) {
+				case Condition::value_type_t::IDENTIFIER: return NodeTools::expect_identifier;
+				case Condition::value_type_t::STRING: return NodeTools::expect_string;
+				case Condition::value_type_t::BOOLEAN: return NodeTools::expect_bool;
+				case Condition::value_type_t::INTEGER: return NodeTools::expect_int;
+				case Condition::value_type_t::REAL: return NodeTools::expect_fixed_point;
+				case Condition::value_type_t::COMPLEX: return this->expect_condition_dictionary;
```
Hop311, line 77: This confused me, I expected it to cause a build error, but of course it doesn't because the template isn't used anywhere, so the code isn't actually compiled.

src/openvic-simulation/scripts/Condition.hpp
Hop311: There's also `std::bitset` which is designed for this kind of stuff, although it wouldn't have the same type as the enums so that might cause complications.

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,318 @@
...
+	ret &= add_condition("administration_spending", REAL, COUNTRY);
+	ret &= add_condition("ai", BOOLEAN, COUNTRY);
+	ret &= add_condition("alliance_with", IDENTIFIER, COUNTRY);
+	ret &= add_condition("average_consciousness", REAL, COUNTRY);
+	ret &= add_condition("average_militancy", REAL, COUNTRY);
+	ret &= add_condition("badboy", REAL, COUNTRY); //NOTE: the actual amount of infamy checked is badboy * 25 for some reason
+	ret &= add_condition("big_producer", IDENTIFIER, COUNTRY);
+	ret &= add_condition("blockade", REAL, COUNTRY);
+	ret &= add_condition("brigades_compare", INTEGER, COUNTRY);
+	ret &= add_condition("can_build_factory_in_capital_state", IDENTIFIER, COUNTRY);
```
wvpm, line 96: boolean?

zaaarf: Not according to the wiki!

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -0,0 +1,233 @@
...
+		switch (value) {
+			case value_type_t::IDENTIFIER: return "IDENTIFIER";
+			case value_type_t::STRING: return "STRING";
+			case value_type_t::BOOLEAN: return "BOOLEAN";
+			case value_type_t::BOOLEAN_INT: return "BOOLEAN_INT";
+			case value_type_t::INTEGER: return "INTEGER";
+			case value_type_t::REAL: return "REAL";
+			case value_type_t::COMPLEX: return "COMPLEX";
+			case value_type_t::GROUP: return "GROUP";
+		}
```
Hop311, line 48: `get_value_type_string`, `get_scope_string` and `get_identifier_type_string` currently return nothing for inputs that are combinations of multiple bits, and `get_value_type_string` also returns nothing for `0`. If they get any of these inputs the behaviour is undefined and it could crash the program. You can avoid this worst case scenario by adding a `default` case, either returning an empty string or some sort of warning/error message.

Also all 3 of these functions can return `std::string_view`s, raw string literals are stored in the program's read-only memory they won't get freed when they go out of scope. Currently that memory is getting copied into a `std::string` every time the functions are called, which would be avoided by using `std::string_view`.

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -0,0 +1,233 @@
...
+	private:
+		const value_type_t PROPERTY(value_type);
+		const scope_t PROPERTY(scope);
+		const scope_t PROPERTY(scope_change);
+		const identifier_type_t PROPERTY(key_identifier_type);
+		const identifier_type_t PROPERTY(value_identifier_type);
+
+		Condition(
+			std::string_view new_identifier, value_type_t new_value_type, scope_t new_scope,
+			scope_t scope_change, identifier_type_t new_key_identifier_type,
```
Hop311, line 159:
```suggestion
			scope_t new_scope_change, identifier_type_t new_key_identifier_type,
```

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -0,0 +1,233 @@
...
+	private:
+		Condition const* PROPERTY(condition);
+		value_t PROPERTY(value);
+		HasIdentifier const* PROPERTY(condition_key_item);
+		HasIdentifier const* PROPERTY(condition_value_item);
+
+		ConditionNode(
+			Condition const* new_condition, value_t&& new_value,
+			HasIdentifier const* condition_key_item = nullptr,
+			HasIdentifier const* condition_value_item = nullptr
```
Hop311, line 189:
```suggestion
			HasIdentifier const* new_condition_key_item = nullptr,
			HasIdentifier const* new_condition_value_item = nullptr
```

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -0,0 +1,233 @@
...
+	private:
+		IdentifierRegistry<Condition> IDENTIFIER_REGISTRY(condition);
+		std::vector<ConditionNode> node_roots; //no need for accessors
+		bool PROPERTY_CUSTOM_PREFIX(late_init_done, is);
+
+		bool add_condition(
+			std::string_view identifier, value_type_t value_type, scope_t scope,
+			scope_t scope_change = scope_t::UNSPECIFIED,
+			identifier_type_t new_key_identifier_type = identifier_type_t::NONE,
+			identifier_type_t new_value_identifier_type = identifier_type_t::NONE
```
Hop311, line 204:
```suggestion
			identifier_type_t key_identifier_type = identifier_type_t::NONE,
			identifier_type_t value_identifier_type = identifier_type_t::NONE
```

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,659 @@
...
+
+	if (scope == 0) {
+		Logger::error("Condition ", identifier, " has no scope specified!");
+		return false;
+	}
+
+	if (scope > (scope_t::COUNTRY | scope_t::PROVINCE | scope_t::POP | scope_t::MAGIC)) {
+		Logger::error("Condition ", identifier, " has no scope specified!");
+		return false;
+	}
```
Hop311, line 50:
```suggestion
	if (value_type == 0 || value_type > value_type_t::MAX_VALUE) {
		Logger::error("Condition ", identifier, " has invalid value type: ", static_cast<uint64_t>(value_type));
		return false;
	}

	if (scope == scope_t::UNSPECIFIED || scope > scope_t::MAX_SCOPE) {
		Logger::error("Condition ", identifier, " has invalid scope: ", static_cast<uint64_t>(scope));
		return false;
	}
```
- Better to use constants defined in the enums rather than hardcoded numbers or manually combining all the scopes, simplest way to define them is: `MAX_VALUE = (1 << 8) - 1` (don't worry about `1 << 8` being bigger than a `uint8_t`, the compiler works the whole constant out before making it 8-bit) and `MAX_SCOPE = (1 << 4) - 1`. If these are defined as enum values it'll be easier to update them if new values are ever added.
- The values are cast to `uint64_t`s in the error message as 8-bit values by default are treated as characters, not numbers, so it needs to be cast to at least 16-bit, and 64-bit covers any future enlargement of the enum type.

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,659 @@
...
+
+	if ((value_identifier_type == identifier_type_t::NONE) && (value_type == value_type_t::IDENTIFIER)) {
+		Logger::error("Condition ", identifier, " has no identifier type!");
+		return false;
+	}
+
+	if ((value_identifier_type != identifier_type_t::NONE) && (value_type != value_type_t::IDENTIFIER)) {
+		Logger::warning("Condition ", identifier, " specified an identifier type, but doesn't have an identifier!");
+	}
```
Hop311, line 59: This is missing `return false;`, unless it's intentional and I'm missing something. Also you could check both this and the case above in one go using `(value_identifier_type == identifier_type_t::NONE) == (value_type == value_type_t::IDENTIFIER)`, at the cost of needing a unified error message.

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,659 @@
...
+	ret &= add_condition("militancy", REAL, POP);
+	ret &= add_condition("political_movement", REAL, POP);
+	ret &= add_condition("pop_majority_issue", IDENTIFIER, POP, UNSPECIFIED, NONE, ISSUE);
+	ret &= add_condition("religion", IDENTIFIER, POP, UNSPECIFIED, NONE, RELIGION);
+	ret &= add_condition("social_movement", BOOLEAN, POP);
+	ret &= add_condition("strata", IDENTIFIER, POP, UNSPECIFIED, NONE, POP_STRATA);
+	ret &= add_condition("type", IDENTIFIER, POP, UNSPECIFIED, NONE, POP_TYPE);
+
+	#define IMPORT_IDENTIFIERS(fun, value_type, scope, scope_change, key_id_type, val_id_type) \
+		for (std::string_view id : fun) { ret &= add_condition(id, value_type, scope, scope_change, key_id_type, val_id_type); }
```
Hop311, line 330:
```suggestion
	const auto import_identifiers = [this, &ret](
		std::vector<std::string_view> const& identifiers, value_type_t value_type, scope_t scope,
		scope_t scope_change = scope_t::UNSPECIFIED,
		identifier_type_t key_identifier_type = identifier_type_t::NONE,
		identifier_type_t value_identifier_type = identifier_type_t::NONE) -> void {
		for (std::string_view identifier : identifiers) {
			ret &= add_condition(
				identifier, value_type, scope, scope_change, key_identifier_type, value_identifier_type
			);
		}
	};
```
Spartan's been advocating against using macros where lambdas also work, and in general they are better, mainly in terms of having specific code that an error message can refer to. In this case they also allow default arguments, e.g. all but one use of `IMPORT_IDENTIFIERS` have their last argument as `NONE`.

Hop311 (commented):
Left comments (all very minor) on everything apart from the expect and late init functions, I'll need a bit more time to look through them properly.

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,659 @@
...
+	ret &= add_condition("any_neighbor_province", GROUP, PROVINCE, PROVINCE);
+	ret &= add_condition("controller", GROUP, PROVINCE, COUNTRY);
+	ret &= add_condition("owner", GROUP, PROVINCE, COUNTRY);
+	ret &= add_condition("state_scope", GROUP, PROVINCE);
+
+	/* Trigger Pop Scopes */
+	ret &= add_condition("location", GROUP, POP, PROVINCE);
+	ret &= add_condition("country", GROUP, POP, COUNTRY);
+
+	/* Magic Conditions */
```
wvpm, line 110: Avoid the term magic in code. It can only create confusion.
AND & OR are collections of conditions.
NOT is basically an inverse OR and can thus be seen as a collection.

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,659 @@
...
+	ret &= add_condition("AND", GROUP, COUNTRY | PROVINCE | POP);
+	ret &= add_condition("OR", GROUP, COUNTRY | PROVINCE | POP);
+	ret &= add_condition("NOT", GROUP, COUNTRY | PROVINCE | POP);
+
+	/* Global Conditions */
```
wvpm, line 115: The term global confused me for some time here. 2/3rd of these conditions are really global (as in world wide). The rest are merely valid in all scopes. I'd be fine with combining this with the mixed scope conditions.

src/openvic-simulation/scripts/Condition.cpp
zaaarf: Spartan is right -- I've just been using macros because they are faster to write and for a prototype speed > correcteness.

src/openvic-simulation/scripts/Condition.cpp
zaaarf: It's intentional here - I decided to leave this as a warning because it doesn't compromise anything.

It just means that the condition has an extra value that will go unused - nothing is missing.

src/openvic-simulation/scripts/Condition.cpp
zaaarf: Fair enough, I'd moved them there recently, I'll move em back

Hop311:
Pushed commit to `conditions-checking-checking` adding helper functions and bug fixes. Summary of changes:

- Added `ScriptManager` in preparation for also having an `EffectManager`.
- Removed `country_migration_target` and `migration_target` FROM scopes.
- Moved scopes from `expect_conditional_weight`'s arguments to variables initialised by `ConditionalWeight`'s constructor, similar to `ConditionScript`.
- Changed `MAX_SCOPE` shift value from 5 to 6.
- Added `share_value_type`, `share_scope` and `share_identifier_type` bitfield helper functions.
- Changed bitfield string functions to use share functions, replaced 0 value cases that could never be triggered with separate if statements, added missing `TECH_SCHOOL` case, changed macro to use compiler string concatenation (adjacent string literals are automatically combined by the compiler, e.g. `"a" "b"` is the same as `"ab"`, including if one of the strings is made from a `#macro_argument`).
- Added `root_condition` variable to `ConditionManager`, set at the end of `setup_conditions`, to avoid needing to look it up for every script.
- Added `expect_condition_script`, used in `ConditionScript::_parse_script`, and made other `ConditionManager` expect functions private.
- Added `case_insensitive_string_set_t` and used it for checking identifier fallback strings.
- Combined `EXPECT_CALL` and `EXPECT_CALL_FALLBACK` using `__VA_OPT__` to add checks for fallback strings if they are provided.
- Added `top_scope` argument to `expect_condition_node_list`, so that keys like `factor` aren't allowed at lower condition scopes. Made `top_scope_fallback` a separate function, as it doesn't depend on any of `expect_condition_node_list`'s arguments and having it as a non-lambda function allows it to be used in a `. ? . : .` ternary bool operator together with `key_value_invalid_callback`, which is a non-lambda function with the same signature. Also removed other keys (`picture`, `title`, etc.) which were appearing due to a now-fixed issue with `Event` scipts not being cached.
- Made `expect_bool` case insensitive.

- Changed `integer_t` to `uint64_t`, as all currently registered conditions with integer type don't accept negative values.
- Added country tag identifier as value option for `is_accepted_culture`, same as `is_primary_culture` already had.
- Changed `revanchism` to country scope.
- Added country tag identifier as value option for `industrial_score`, same as `military_score` already had.
- Added `pop_type` condition, alias of `type`.
- Changed `brigades_compare` and `lost_national` to real value type.
- Added province identifier as value option for `is_core`.
- Allowed `owned_by`, `empty`, `any_neighbor_province`, `any_owned_province` and `controlled_by` to be used from state scope.

zaaarf:
Merged huber's stuff and added a few final tweaks. Should be good to go!

wvpm (dismissed):

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -188,13 +182,17 @@ bool PopManager::load_pop_type_file(
 		"luxury_needs", ZERO_OR_ONE, good_manager.expect_good_decimal_map(move_variable_callback(luxury_needs)),
 		"everyday_needs", ZERO_OR_ONE, good_manager.expect_good_decimal_map(move_variable_callback(everyday_needs)),
 		"life_needs", ZERO_OR_ONE, good_manager.expect_good_decimal_map(move_variable_callback(life_needs)),
-		"country_migration_target", ZERO_OR_ONE, country_migration_target.expect_conditional_weight(ConditionalWeight::FACTOR),
-		"migration_target", ZERO_OR_ONE, migration_target.expect_conditional_weight(ConditionalWeight::FACTOR),
+		"country_migration_target", ZERO_OR_ONE, country_migration_target.expect_conditional_weight(
+			ConditionalWeight::FACTOR, scope_t::COUNTRY, scope_t::POP, scope_t::POP
```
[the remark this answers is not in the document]
Hop311: `FROM` isn't used in `country_migration_target` or `migration_target` in vanilla or any of the mods I checked, I've already changed it to `NO_SCOPE` for both of them.

src/openvic-simulation/dataloader/NodeTools.cpp
```diff
@@ -75,8 +75,9 @@ node_callback_t NodeTools::expect_bool(callback_t<bool> callback) {
 }
 
 node_callback_t NodeTools::expect_int_bool(callback_t<bool> callback) {
-	static const string_map_t<bool> bool_map = { { "1", true }, { "0", false } };
-	return expect_identifier(expect_mapped_string(bool_map, callback));
+	return expect_uint64([callback](uint64_t num) -> bool {
```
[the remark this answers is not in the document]
zaaarf: Because that's how vanilla handles it - there's an occurrence of it in vanilla defines, too, so I can only assume that it's expected behaviour.

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -0,0 +1,716 @@
...
+		COUNTRY,
+		NO_SCOPE,
+		REFORM_GROUP,
+		REFORM
+	);
+
+	import_identifiers(
+		game_manager.get_politics_manager().get_issue_manager().get_reform_identifiers(),
+		REAL,
+		COUNTRY,
```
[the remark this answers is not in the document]
zaaarf: we have fallbacks, so scoping it to country means anything works

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -948,69 +948,41 @@ bool Dataloader::load_defines(GameManager& game_manager) {
...
+	ret &= parse_scripts(game_manager);
 
 	free_cache();
 
 	return ret;
 }
 
+#define PARSE_SCRIPTS(name, parse) \
+	if (!parse) { Logger::error("Failed to parse ", name, " scripts!"); ret = false; } \
+	else Logger::info("Successfully parsed ", name, " scripts!");
```
Hop311, line 965:
```suggestion
#define PARSE_SCRIPTS(name, parse) \
	if (!parse.parse_scripts(game_manager)) { Logger::error("Failed to parse ", name, " scripts!"); ret = false; } \
	else Logger::info("Successfully parsed ", name, " scripts!");
```
You could move the `parse_scripts` call to the macro and just pass the manager to it

src/openvic-simulation/misc/Event.cpp
```diff
@@ -2,12 +2,14 @@
 
 #include "openvic-simulation/dataloader/NodeTools.hpp"
 #include "openvic-simulation/politics/Issue.hpp"
+#include "scripts/Condition.hpp"
+#include "utility/Logger.hpp"
```
Hop311, line 6:
```suggestion
#include "openvic-simulation/scripts/Condition.hpp"
#include "openvic-simulation/utility/Logger.hpp"
```

src/openvic-simulation/dataloader/NodeTools.cpp
```diff
@@ -70,13 +70,15 @@ node_callback_t NodeTools::expect_identifier_or_string(callback_t<std::string_vi
...
-	static const string_map_t<bool> bool_map = { { "yes", true }, { "no", false } };
+	static const case_insensitive_string_map_t<bool> bool_map = { { "yes", true }, { "no", false } };
 	return expect_identifier(expect_mapped_string(bool_map, callback));
 }
 
 node_callback_t NodeTools::expect_int_bool(callback_t<bool> callback) {
-	static const string_map_t<bool> bool_map = { { "1", true }, { "0", false } };
-	return expect_identifier(expect_mapped_string(bool_map, callback));
+	return expect_uint64([callback](uint64_t num) -> bool {
+		if (num > 1) Logger::warning("Found int bool with value >1: ", num);
```
Hop311, line 79:
```suggestion
		if (num > 1) {
			Logger::warning("Found int bool with value >1: ", num);
		}
```
Can you add brackets after the `if`

src/openvic-simulation/misc/Event.cpp
```diff
@@ -144,20 +161,20 @@ bool EventManager::load_event_file(IssueManager const& issue_manager, ast::NodeC
 				"election", ZERO_OR_ONE, expect_bool(assign_variable_callback(election)),
 				"issue_group", ZERO_OR_ONE,
 					issue_manager.expect_issue_group_identifier(assign_variable_callback_pointer(election_issue_group)),
-				"option", ONE_OR_MORE, [&options](ast::NodeCPtr node) -> bool {
-					std::string_view title;
+				"option", ONE_OR_MORE, [&options, initial_scope](ast::NodeCPtr node) -> bool {
+					std::string_view name;
 					EffectScript effect;
-					ConditionalWeight ai_chance;
+					ConditionalWeight ai_chance { initial_scope, initial_scope, scope_t::NO_SCOPE };
```
Hop311, line 167: I thought `ai_chance` could have country or province as its `FROM` scope?

src/openvic-simulation/dataloader/Dataloader.cpp
zaaarf: didnt think of it, nice

Hop311 (approved):
Great work!

PR #114, opened 2024-01-02
Integer and fixed point squared roots

### LINKED ISSUE #99: Distance function ignores east-west map wrapping
The function below (from Province.cpp) does not appear to take into account the east-west wrapping of the map. It will yield incorrectly large distances between provinces on the west edge of the map and those on the east edge.

```
Province::distance_t Province::calculate_distance_to(Province const* province) const {
	const fvec2_t my_unit_position = get_unit_position();
	const fvec2_t other_unit_position = province->get_unit_position();
	const fvec2_t distance_vector = other_unit_position - my_unit_position;
	const fixed_point_t distance = distance_vector.length_squared();
	return static_cast<Province::distance_t>(distance);
}
```

zaaarf:
Implemented [this](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Binary_numeral_system_(base_2)) simple deterministic algorithm.

...and since now we can, I fixed #99 in the other commit.


src/openvic-simulation/map/Province.cpp
```diff
@@ -1,6 +1,10 @@
 #include "Province.hpp"
 
 #include "openvic-simulation/history/ProvinceHistory.hpp"
+#include "openvic-simulation/map/Map.hpp"
+#include "types/Vector.hpp"
+#include "types/fixed_point/FixedPoint.hpp"
```
Spartan322, line 6:
```suggestion
#include "openvic-simulation/types/Vector.hpp"
#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
```

zaaarf: AAAAAAA

src/openvic-simulation/map/Map.hpp
```diff
@@ -5,6 +5,7 @@
 
 #include <openvic-dataloader/csv/LineObject.hpp>
 
+#include "map/Province.hpp"
```
Hop311, line 8:
```suggestion
#include "openvic-simulation/map/Province.hpp"
```

src/openvic-simulation/map/Map.hpp
```diff
@@ -89,6 +90,13 @@ namespace OpenVic {
 		bool add_province(std::string_view identifier, colour_t colour);
 		IDENTIFIER_REGISTRY_NON_CONST_ACCESSORS_CUSTOM_INDEX_OFFSET(province, 1);
 
+		Province::distance_t calculate_distance_between(Province const& from, Province const& to);
+		bool add_standard_adjacency(Province& from, Province& to);
+		bool add_special_adjacency(
+			Province& from, Province& to, Province::adjacency_t::type_t type, Province const* through,
+			Province::adjacency_t::data_t data
+		);
```
Hop311, line 98:
```suggestion
		Province::distance_t calculate_distance_between(Province const& from, Province const& to) const;
		bool add_standard_adjacency(Province& from, Province& to) const;
		bool add_special_adjacency(
			Province& from, Province& to, Province::adjacency_t::type_t type, Province const* through,
			Province::adjacency_t::data_t data
		) const;
```

src/openvic-simulation/map/Province.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Province.hpp"
 
 #include "openvic-simulation/history/ProvinceHistory.hpp"
+#include <algorithm>
```
Hop311, line 4: I think this is leftover from when you were using `std::min` here, can you remove it since it's not needed anymore

src/openvic-simulation/map/Map.cpp
```diff
@@ -69,6 +69,173 @@ bool Map::add_province(std::string_view identifier, colour_t colour) {
...
 
+Province::distance_t Map::calculate_distance_between(Province const& from, Province const& to) {
+	const fvec2_t to_pos = to.get_unit_position();
+	const fvec2_t from_pos = from.get_unit_position();
+
+	const fixed_point_t min_x = std::min(
+		to_pos.x - from_pos.x,
+		std::min(
+			to_pos.x - from_pos.x + width,
+			to_pos.x - from_pos.x - width
```
Hop311, line 80:
```suggestion
		(to_pos.x - from_pos.x).abs(),
		std::min(
			(to_pos.x - from_pos.x + width).abs(),
			(to_pos.x - from_pos.x - width).abs()
```
We need to take the absolute value otherwise it'll select large negative ones.

Hop311 (commented):
As well as the comments below, the fixed point sqrt function is always returning `0.9999847412109375` for some reason

src/openvic-simulation/utility/NumberUtils.hpp
```diff
@@ -24,4 +24,24 @@ namespace OpenVic::NumberUtils {
 		}
 		return ret;
 	}
+
+	constexpr uint64_t sqrt(uint64_t n) {
```
wvpm, line 28: Does not work.
sqrt of 100.0 (fixed_point_t) gives 0.9999847412109375

wvpm (changes requested):
Please fix and test sqrt

src/openvic-simulation/map/Map.cpp
zaaarf: oh, right!

wvpm (approved):

Hop311 (approved):
Working for me 👍

PR #116, opened 2024-01-03
Mutually exclusive rule groups + modded building rules

Hop311:
- Rules now have a group, if that group is mutually exclusive then `trim_and_resolve_conflicts` removes disabled rules in that group as well as any enabled rules beyond the one with the lowest index (from order of definiton).
- Added `build_*` rules for every building type type (i.e. the `type` identifier of `BuildingType`s), with a special case for `railroad` which has type `infrastructure` but rule `build_railway`.
- Fixed reform tech cost warnings.
- Moved issue&reform and delayed poptype loading until after rules setup, which has itself been moved to after building type loading.
- Added some missing modifiers (for TGC compatibility), changed artisan/factory input modifiers not to be `positive_good` and removed RGO input modifiers.

### DIFF
--- a/src/openvic-simulation/politics/Issue.cpp
+++ b/src/openvic-simulation/politics/Issue.cpp
@@ -95,8 +95,19 @@ bool IssueManager::add_reform(
 	}
 
 	if (group->get_type().is_uncivilised()) {
-		if (technology_cost <= 0) {
-			Logger::warning("Non-positive technology cost ", technology_cost, " found in uncivilised reform ", identifier, "!");
+		if (ordinal == 0) {
+			if (technology_cost != 0) {
+				Logger::warning(
+					"Non-zero technology cost ", technology_cost, " found in ordinal 0 uncivilised reform ", identifier, "!"
+				);
+			}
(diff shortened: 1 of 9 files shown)


src/openvic-simulation/politics/Issue.cpp
```diff
@@ -184,8 +195,11 @@ bool IssueManager::load_issues_file(ModifierManager const& modifier_manager, Rul
 			if (key == "party_issues") {
 				return expect_length(add_variable_callback(expected_issue_groups))(value);
 			} else {
+				static const string_set_t uncivilised_reform_groups {
+					"economic_reforms", "education_reforms", "military_reforms"
+				};
 				return expect_length(add_variable_callback(expected_reform_groups))(value)
-					& add_reform_type(key, key == "economic_reforms" || key == "education_reforms" || key == "military_reforms");
+					& add_reform_type(key, uncivilised_reform_groups.contains(key));
```
wvpm, line 202: Nice

wvpm (approved):

PR #118, opened 2024-01-03
Added case insensitive ordered set and map and IdentifierRegistry

Hop311:
- Added case insensitive string hash and equal structs.
- Added `case_insensitive_ordered_map` and `case_insensitive_ordered_set`.
- Added `CaseInsensitiveIdentifierRegistry` and `CaseInsensitiveIdentifierInstanceRegistry`.
- Updated `string_map_t` and `UniqueKeyRegistry` related templates to accept `Hash` and `KeyEqual` parameters.
- Fixed event file `Parser`s not being cached.
- Added temporary `days` and `years` keys to `ConditionalWeight` so that mod parsing succeeds, these will be replaced as part of an upcoming `expect_dictionary_keys` rework PR.
- Fixed `get_item_identifiers` returning loads of empty identifiers (it was allocating space for all the identifiers but then adding them on top of that, so the allocated space just contained empty `std::string_view`s).

### DIFF
--- a/src/openvic-simulation/types/OrderedContainers.hpp
+++ b/src/openvic-simulation/types/OrderedContainers.hpp
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <concepts>
+#include <cctype>
 #include <deque>
 #include <functional>
 #include <memory>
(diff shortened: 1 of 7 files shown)


zaaarf (approved):

src/openvic-simulation/types/OrderedContainers.hpp
```diff
@@ -98,4 +99,67 @@ namespace OpenVic {
...
+
+	/* Case-Insensitive Containers */
+	struct case_insensitive_string_hash {
+		using is_transparent = void;
+
+	private:
+		/* Based on the byte array hashing functions in MSVC's <type_traits>. */
+		[[nodiscard]] static constexpr size_t _hash_bytes_case_insensitive(char const* first, size_t count) {
+			constexpr size_t _offset_basis = 14695981039346656037ULL;
+			constexpr size_t _prime = 1099511628211ULL;
```
Spartan322, line 111: Probably should be:
```suggestion
			static constexpr size_t _offset_basis = 14695981039346656037ULL;
			static constexpr size_t _prime = 1099511628211ULL;
```

Hop311: Clangd says: `Definition of a static variable in a constexpr function is a C++23 extension`
I initially had them as `static constexpr` member variables of `case_insensitive_string_hash`, but seeing as they're only used in this function and only once each, I thought it simpler to just have them be local, I'd expect them to be subsituted in to the instructions that use them anyway (assign a `size_t` variable to `_offset_basis` and `*=` that variable by `_prime`).

PR #119, opened 2024-01-05
Add skeleton CountryRelationManager

Spartan322:
Change `DiplomaticAction::Argument` to refer to `CountryInstance*` instead of `Country*`
Add `GameManager&` argument to `DiplomaticActionManager::setup_diplomatic_actions`
Add some ErrorMacros, likely and unlikely macros and stringization macros (based on Godot's macros)
Implement `increase_relations` DiplomaticAction commit function
Implement `decrease_relations` DiplomaticAction commit function


BrickPi (approved):
LGTM, should be good to merge as long as its tested. Assert macros look especially useful, should mention that in the dev chat as something to use going forward.

PR #121, opened 2024-01-07
Reparations & optional end date

wvpm:


src/openvic-simulation/history/DiplomaticHistory.hpp
```diff
@@ -34,10 +35,9 @@ namespace OpenVic {
 
 		private:
 			Country const* PROPERTY(country);
-			Date PROPERTY_CUSTOM_PREFIX(joined, get_date);
-			std::optional<Date> PROPERTY_CUSTOM_PREFIX(exited, get_date);
+			Period period;
```
Hop311, line 38:
```suggestion
			Period PROPERTY(period);
```
This is correctly non-const, because you need to update it when they exit the war 👍

src/openvic-simulation/history/DiplomaticHistory.hpp
```diff
@@ -55,10 +55,20 @@ namespace OpenVic {
 	private:
 		Country const* PROPERTY(first);
 		Country const* PROPERTY(second);
-		const Date start;
-		const Date end;
+		Period PROPERTY(period);
```
Hop311, line 58:
```suggestion
		const Period PROPERTY(period);
```

src/openvic-simulation/history/DiplomaticHistory.hpp
```diff
@@ -55,10 +55,20 @@ namespace OpenVic {
...
+		AllianceHistory(Country const* new_first, Country const* new_second, const Period period);
+	};
+
+	struct ReparationsHistory {
+		friend struct DiplomaticHistoryManager;
+
+	private:
+		Country const* PROPERTY(receiver);
+		Country const* PROPERTY(sender);
+		const Period period;
```
Hop311, line 69:
```suggestion
		const Period PROPERTY(period);
```

src/openvic-simulation/history/DiplomaticHistory.hpp
```diff
@@ -74,15 +84,15 @@ namespace OpenVic {
 		Country const* PROPERTY(overlord);
 		Country const* PROPERTY(subject);
 		const type_t PROPERTY_CUSTOM_PREFIX(type, get_subject);
-		const Date start;
-		const Date end;
+		const Period period;
```
Hop311, line 87:
```suggestion
		const Period PROPERTY(period);
```

src/openvic-simulation/history/DiplomaticHistory.cpp
```diff
@@ -84,61 +103,86 @@ bool DiplomaticHistoryManager::load_diplomacy_history_file(CountryManager const&
 		"alliance", ZERO_OR_MORE, [this, &country_manager](ast::NodeCPtr node) -> bool {
 			Country const* first;
 			Country const* second;
```
Hop311, line 105:
```suggestion
			Country const* first = nullptr;
			Country const* second = nullptr;
```
Can you null-initialise these here and in call cases below (with MSVC they'll have random addresses if they're not explicitly initialised, and even though we expect a first and second country we still need to account for the possibility that they're not in the defines).

src/openvic-simulation/history/DiplomaticHistory.cpp
```diff
@@ -84,61 +103,86 @@ bool DiplomaticHistoryManager::load_diplomacy_history_file(CountryManager const&
...
 				"first", ONE_EXACTLY, expect_identifier_or_string(country_manager.expect_country_str(assign_variable_callback_pointer(first))),
 				"second", ONE_EXACTLY, expect_identifier_or_string(country_manager.expect_country_str(assign_variable_callback_pointer(second))),
 				"start_date", ONE_EXACTLY, expect_identifier_or_string(expect_date_str(assign_variable_callback(start))),
-				"end_date", ONE_EXACTLY, expect_identifier_or_string(expect_date_str(assign_variable_callback(end)))
+				"end_date", ZERO_OR_ONE, expect_identifier_or_string(expect_date_str(assign_variable_callback(end)))
 			)(node);
 
-			alliances.push_back({ first, second, start, end });
+			const Period period = { start, end };
+			alliances.push_back({ first, second, period });
```
Hop311, line 117:
```suggestion
			alliances.push_back({ first, second, { start, end } });
```
Here you're creating the variable `period` then copying it into the alliance on the next line, if instead you directly create the period object as an argument when constructing the alliance you avoid needing to allocate memory for the intermediate `period` variable and avoid having to do a copy. Same thing applies to all the other cases below.

src/openvic-simulation/history/Period.hpp
```diff
@@ -0,0 +1,18 @@
...
+
+#include <optional>
+#include "openvic-simulation/types/Date.hpp"
+#include "openvic-simulation/utility/Logger.hpp"
+
+namespace OpenVic {
+	struct Period {
+	private:
+		const Date start_date;
+		std::optional<Date> end_date;
```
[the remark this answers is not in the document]
wvpm: Why should they be accessible?
They are only used inside the period itself.

src/openvic-simulation/history/Period.cpp
```diff
@@ -0,0 +1,27 @@
...
+	const std::optional<Date> new_end_date
+) : start_date {new_start_date}, end_date {new_end_date} {}
+
+bool Period::is_date_in_period(const Date date) const {
+	return start_date <= date && (!end_date.has_value() || end_date.value() >= date);
+}
+
+bool Period::try_set_end(const Date date) {
+	if (end_date.has_value()) {
+		Logger::error("Period already has end date " + end_date.value().to_string());
```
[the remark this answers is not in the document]
wvpm: Good to know. I'll change it.

src/openvic-simulation/history/DiplomaticHistory.cpp
wvpm: Good to know, will do.

src/openvic-simulation/history/Period.hpp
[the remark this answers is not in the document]
Hop311: Oh yeah, forgot this wasn't a general purpose time period struct

Hop311 (approved):
Good work 👍

PR #123, opened 2024-01-08
Added map_callback and expect_item_dictionary_reserve_length

Hop311:
- Added `map_callback`, which includes a check for duplicates, and used it in dataloading code.
- Added more `expect_item_dictionary` variants, including `and_length` and `reserve_length` variants.

### DIFF
--- a/src/openvic-simulation/country/Country.cpp
+++ b/src/openvic-simulation/country/Country.cpp
@@ -1,10 +1,6 @@
 #include "Country.hpp"
 
-#include <filesystem>
 #include <string_view>
-#include <system_error>
-#include <unordered_map>
-#include <vector>
 
 #include <openvic-dataloader/v2script/AbstractSyntaxTree.hpp>
 
@@ -125,8 +121,7 @@ node_callback_t CountryManager::load_country_party(
 						return politics_manager.get_issue_manager().expect_issue_identifier(
 							[&policies, &group](Issue const& issue) -> bool {
 								if (&issue.get_group() == &group) {
-									policies.emplace(&group, &issue);
-									return true;
+									return map_callback(policies, &group)(&issue);
 								}
 								// TODO - change this back to error/false once TGC no longer has this issue
 								Logger::warning("Invalid policy ", issue.get_identifier(), ", group is ",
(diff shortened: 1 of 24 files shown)


Spartan322 (approved):

src/openvic-simulation/country/Country.cpp
```diff
@@ -184,15 +170,10 @@ bool CountryManager::load_country_data_file(
...
-						Logger::error("Country ", name, " has duplicate entry for ", unit.get_identifier(), " name list");
-						return false;
-					}
-					return name_list_callback([&unit_names, &unit](std::vector<std::string>&& list) -> bool {
-						return unit_names.emplace(&unit, std::move(list)).second;
-					})(value);
+			game_manager.get_military_manager().get_unit_manager().expect_unit_dictionary_reserve_length(
+				unit_names,
+				[&unit_names](Unit const& unit, ast::NodeCPtr value) -> bool {
+					return name_list_callback(map_callback(unit_names, &unit))(value);
```
wvpm, line 176: Unit names can be duplicate I believe. It makes them appear more often.

src/openvic-simulation/dataloader/NodeTools.cpp
```diff
@@ -411,6 +411,17 @@ node_callback_t NodeTools::name_list_callback(callback_t<std::vector<std::string
 	};
 }
 
+std::ostream& OpenVic::operator<<(std::ostream& stream, name_list_t const& name_list) {
```
wvpm, line 414: Why?
std::vector<std::string> makes more sense imo

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -721,7 +721,7 @@ node_callback_t ConditionManager::expect_condition_node_list(
 		};
 
 		bool ret = conditions.expect_item_dictionary_and_default(
-			expect_node, top_scope ? top_scope_fallback : key_value_invalid_callback
+			top_scope ? top_scope_fallback : key_value_invalid_callback, expect_node
```
wvpm, line 724: Are you sure this is intentional and not some leftover code?
I thought we just got conditions working.

wvpm (approved):

src/openvic-simulation/country/Country.cpp
Hop311: `name_list_callback` parses the actual names and stores them in a vector of strings, the map here is from unit types to name lists, so each unit has a single list of names, but that list can include duplicate names.

src/openvic-simulation/scripts/Condition.cpp
Hop311: I switched the order of arguments in `expect_item_dictionary_and_default` to match `expect_dictionary_keys_and_default`.

src/openvic-simulation/dataloader/NodeTools.cpp
Hop311: I added the type alias to make it clear what the string vector is expected to represent in this operator overload, and to make it clear that only this type has an overloaded print operator, instead of there being more general vector printing operators. Tbh this operator isn't actually needed, I added it because `map_callback` previously also printed the value when the key was duplicate, but I removed that because the requirement that values had to be printable became too restrictive.

PR #124, opened 2024-01-09
Improved validation of production types and renamed properties to be descriptive

wvpm:
Artisanal production ignores owners and employees.
RGO & factory production requires both.
`poptype = artisan` is never valid. It's simply ignored for artisanal production.

Several properties were renamed to reflect their meaning instead of their name in production_types.txt.

### DIFF
--- a/src/openvic-simulation/economy/ProductionType.cpp
+++ b/src/openvic-simulation/economy/ProductionType.cpp
@@ -3,21 +3,24 @@
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
-EmployedPop::EmployedPop(
-	PopType const* new_pop_type, bool new_artisan, effect_t new_effect, fixed_point_t new_effect_multiplier,
-	fixed_point_t new_amount
-) : pop_type { new_pop_type }, artisan { new_artisan }, effect { new_effect }, effect_multiplier { new_effect_multiplier },
-	amount { new_amount } {}
+Job::Job(
+	PopType const* const new_pop_type, effect_t new_effect_type, fixed_point_t new_effect_multiplier,
+	fixed_point_t new_desired_workforce_share
+)
+	: pop_type { new_pop_type }, effect_type { new_effect_type }, effect_multiplier { new_effect_multiplier },
+	  desired_workforce_share { new_desired_workforce_share } {}
 
 ProductionType::ProductionType(
-	std::string_view new_identifier, EmployedPop new_owner, std::vector<EmployedPop> new_employees, type_t new_type,
-	Pop::pop_size_t new_workforce, Good::good_map_t&& new_input_goods, Good const* new_output_goods,
-	fixed_point_t new_value, std::vector<bonus_t>&& new_bonuses, Good::good_map_t&& new_efficiency, bool new_coastal,
-	bool new_farm, bool new_mine
-) : HasIdentifier { new_identifier }, owner { new_owner }, employees { new_employees }, type { new_type },
-	workforce { new_workforce }, input_goods { std::move(new_input_goods) }, output_goods { new_output_goods },
-	value { new_value }, bonuses { std::move(new_bonuses) }, efficiency { std::move(new_efficiency) },
-	coastal { new_coastal }, farm { new_farm }, mine { new_mine } {}
+	std::string_view new_identifier, Job new_owner, std::vector<Job> new_jobs, template_type_t new_template_type,
+	Pop::pop_size_t new_base_workforce_size, Good::good_map_t&& new_input_goods, Good const* const new_output_goods,
+	fixed_point_t new_base_output_quantity, std::vector<bonus_t>&& new_bonuses, Good::good_map_t&& new_maintenance_requirements,
+	bool new_is_coastal, bool new_is_farm, bool new_is_mine
+)
+	: HasIdentifier { new_identifier }, owner { new_owner }, jobs { new_jobs }, template_type { new_template_type },
+	  base_workforce_size { new_base_workforce_size }, input_goods { std::move(new_input_goods) },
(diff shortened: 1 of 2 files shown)


src/openvic-simulation/economy/ProductionType.cpp
```diff
@@ -29,96 +32,99 @@ bool ProductionType::parse_scripts(GameManager const& game_manager) {
...
-	GoodManager const& good_manager, PopManager const& pop_manager, callback_t<EmployedPop&&> cb
+node_callback_t ProductionTypeManager::_expect_job(
+	GoodManager const& good_manager, PopManager const& pop_manager, callback_t<Job&&> cb
 ) {
 
 	return [this, &good_manager, &pop_manager, cb](ast::NodeCPtr node) -> bool {
 		std::string_view pop_type;
-		EmployedPop::effect_t effect;
-		fixed_point_t effect_multiplier = 1, amount = 1;
+		Job::effect_t effect_type;
```
Hop311, line 41:
```suggestion
		Job::effect_t effect_type = THROUGHPUT;
```
Just to be on the safe side, this should have a default value, so it doesn't contain nonesense data if `effect` is missing or invalid. If you move `using enum Job::effect_t;` from a couple lines below to the top of the file you won't need to write the enum's `Job::effect_t::` prefix.

src/openvic-simulation/economy/ProductionType.cpp
```diff
@@ -29,96 +32,99 @@ bool ProductionType::parse_scripts(GameManager const& game_manager) {
...
 		bool res = expect_dictionary_keys(
 			"poptype", ONE_EXACTLY, expect_identifier(assign_variable_callback(pop_type)),
-			"effect", ONE_EXACTLY, expect_identifier(expect_mapped_string(effect_map, assign_variable_callback(effect))),
+			"effect", ONE_EXACTLY, expect_identifier(expect_mapped_string(effect_map, assign_variable_callback(effect_type))),
 			"effect_multiplier", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(effect_multiplier)),
-			"amount", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(amount))
+			"amount", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(desired_workforce_share))
 		)(node);
 
 		const PopType* found_pop_type = pop_manager.get_pop_type_by_identifier(pop_type);
```
Hop311, line 56:
```suggestion
		PopType const* found_pop_type = pop_manager.get_pop_type_by_identifier(pop_type);
		if (found_pop_type == nullptr && pop_type != "artisan") {
			Logger::error("Invalid pop type identifier \"", pop_type, "\"!");
			res = false;
		}
```
Even though it's checked later in `add_production_type`, there should still be an error message here if the pop type is invalid, which can include the actual invalid string.

src/openvic-simulation/economy/ProductionType.cpp
```diff
@@ -29,96 +32,99 @@ bool ProductionType::parse_scripts(GameManager const& game_manager) {
...
+		}
+
+		for (int i = 0; i < jobs.size(); i++) {
+			const Job job = jobs[i];
+			if(job.get_pop_type()== nullptr) {
+				Logger::error("Production type ", identifier, " has invalid pop type in employees[",i,"].");
+				return false;
+			}
+		}
+	}
```
Hop311, line 121:
```suggestion
	if (template_type == ProductionType::template_type_t::ARTISAN) {
		if(owner.get_pop_type() != nullptr || !jobs.empty()) {
			Logger::warning("Artisanal production types don't use owner and employees. Effects are ignored.");
		}
	} else {
		if (owner.get_pop_type() == nullptr) {
			Logger::error("Production type ", identifier, " lacks owner or has an invalid pop type.");
			return false;
		}

		if (jobs.empty()) {
			Logger::error("Production type ", identifier, " lacks jobs ('employees').");
			return false;
		}

		for (size_t i = 0; i < jobs.size(); i++) {
			if (jobs[i].get_pop_type() == nullptr) {
				Logger::error("Production type ", identifier, " has invalid pop type in employees[",i,"].");
				return false;
			}
		}
	}
```
Minor formatting tweaks, also:
- `jobs.size()` is a `size_t`, so might as well use that type for `i`.
- `const Job job = jobs[i];` creates a copy of the `Job` object, you could instead create a const reference `Job const& job = jobs[i];` which doesn't require a copy or just use `jobs[i]` directly since it's only needed once.

src/openvic-simulation/economy/ProductionType.cpp
```diff
@@ -175,46 +181,46 @@ bool ProductionTypeManager::load_production_types_file(
 				return true;
 			}
 
-			EmployedPop owner;
-			std::vector<EmployedPop> employees;
-			ProductionType::type_t type;
+			Job owner {};
+			std::vector<Job> jobs;
+			ProductionType::template_type_t template_type;
```
Hop311, line 186:
```suggestion
			ProductionType::template_type_t template_type = FACTORY;
```
Again, safer to have a default value, and you can move `using enum ProductionType::template_type_t;` to the top of the function to avoid needing the enum's prefix.

src/openvic-simulation/economy/ProductionType.cpp
wvpm: You can't do that. "artisan" is an invalid poptype. It's just that for artisanal production types, the owner and employees are ignored. If you add "artisan" in a factory or RGO, the game will crash. I tried to pass the production type to this method but it's not parsed yet.

The proper way to handle it is to first figure out if it is an artisanal production type or something else.
If it is, then warn about any owner or employee (apart from vanilla's artisan placeholder).
Else, error even on "artisan".

Hop311 (approved):

PR #125, opened 2024-01-09
Added producer types

wvpm:


src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,86 @@
...
+				static_cast<fixed_point_t>(production_type.get_base_workforce_size()) * size_multiplier
+			);
+		}
+
+	public:
+		money_t PROPERTY(revenue_yesterday);
+		good_quantity_t PROPERTY(output_quantity_yesterday);
+		good_quantity_t PROPERTY(unsold_quantity_yesterday);
+		fixed_point_t PROPERTY(size_multiplier);
+		std::map<Pop const*, Pop::pop_size_t> PROPERTY(employees)
```
Hop311, line 45:
```suggestion
		ordered_map<Pop const*, Pop::pop_size_t> PROPERTY(employees);
```
We've switched to using `ordered_map` across the repo, both for consistency and performance. Also although it's technically not needed it's nice the include the `;` like we do with the other `PROPERTY` variables.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,86 @@
...
+	};
+
+	class FactoryProducer final : NonArtisanalProducer, Manufacturer {
```
Hop311, line 48: - When extending a class, by default everything is made private, even if it's public in the base class. To get around this, you either need to change the `class`es to `struct`s or you need to change them here to `: public NonArtisanalProducer, public Manufacturer`.
- As `NonArtisanalProducer` and `Manufacturer` both extend `Producer`, by default `FactoryProducer` will contain two separate copies of `Producer`. To get around this you need to use virtual inheritance, so `: virtual NonArtisanalProducer, virtual Manufacturer` (and make them `public`, unless you change them to `struct`s). You'll also need to add a virtual destructor to `Producer`. I'd recommend you look up virtual inheritance, as it's a bit of a weird concept in C++.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+using money_t = OpenVic::fixed_point_t;
+using good_quantity_t = OpenVic::fixed_point_t;
+
+namespace OpenVic {
+	class Producer {
+	public:
+		ProductionType PROPERTY(production_type)
```
Hop311, line 18:
```suggestion
		ProductionType const& PROPERTY(production_type);
```
We won't store copies of `ProductionType`s, rather just a reference to the global copy of them created when loading the defines (reference rather than pointer, because `production_type` should never be null).

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,86 @@
...
+		good_quantity_t PROPERTY(output_quantity_yesterday);
+		good_quantity_t PROPERTY(unsold_quantity_yesterday);
+		fixed_point_t PROPERTY(size_multiplier);
+		std::map<Pop const*, Pop::pop_size_t> PROPERTY(employees)
+	};
+
+	class FactoryProducer final : NonArtisanalProducer, Manufacturer {
+	private:
+		int PROPERTY(profit_history_current);
+		money_t[7] PROPERTY(daily_profit_history);
```
Hop311, line 51: Rather than using `7` directly, make a `static constexpr size_t` constant variable and use it here. Also better to use `std::array<money_t, the_constant>`, as it's safer than a raw C-style array.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+		fixed_point_t get_max_budget() {
+			return defines.economy.MAX_FACTORY_MONEY_SAVE * size_multiplier;
```
Hop311, line 54: We'll need a variable somewhere which we initialise with the defines value using `DefineManager`, not sure where we should store it yet, maybe a static variable in `FactoryProducer`, or somehow accessible via a global `DefineManager` instance.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,86 @@
...
+	public:
+		money_t PROPERTY(budget);
+		money_t PROPERTY(balance_yesterday);
+		money_t PROPERTY(received_investments_yesterday);
+		money_t PROPERTY(market_spendings_yesterday);
+		money_t PROPERTY(paychecks_yesterday);
+		uint32_t PROPERTY(unprofitable_days);
+		uint32_t PROPERTY(injected_days);
+		uint32_t PROPERTY(days_without_input);
+		std::byte PROPERTY_RW_ACCESS(hiring_priority, public);
```
Hop311, line 66: `std::byte` only has bitwise operators, you can't use `+`, `-`, `>`, `<`, `>=`, `<=`, etc. with it. Instead `uint8_t` should be used.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,86 @@
...
+			for (int i = 0; i <= profit_history_current; i++) {
+				sum += daily_profit_history[i];
+			}
+
+			return sum / (1 + profit_history_current);
+		}
+	};
+
+	class ResourceGatheringOpertion final : NonArtisanalProducer {
+		ResourceGatheringOpertion(const ProductionType production_type, const fixed_point_t size_multiplier);
```
Hop311, line 84: As mention in `Producer`, `production_type` should be a `ProductionType const*`. Also function arguments never need to be `const` (pointers and references can need `const` before the `*`/`&`, but that's different from the variable itself being const). In the header definition `const` arguments are extra unnecessary, it doesn't actually have any effect, the implementation's arguments don't have to match the `const`-ness of the header definition's arguments.

src/openvic-simulation/economy/Producer.hpp
wvpm: I missed the `;`. As for ordered_map, I thought about it and realised order is irrelevant here. Do you still want an ordered map?

src/openvic-simulation/economy/Producer.hpp
wvpm: I did not expect that. I read some stuff about c++ not having interfaces (which I'd love to use here).
I'll make them public & virtual.

src/openvic-simulation/economy/Producer.hpp
wvpm: I removed methods that don't require instance specific data.
They might be added in a runtime manager/behaviour/mechanics class.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <cstddef>
+#include <cstdint>
+
+#include "openvic-simulation/economy/Good.hpp"
+#include "openvic-simulation/economy/ProductionType.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+using money_t = OpenVic::fixed_point_t;
+using good_quantity_t = OpenVic::fixed_point_t;
```
Spartan322, line 12: Avoid polluting the global scope

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,73 @@
...
+		fixed_point_t PROPERTY(unsold_quantity_yesterday);
+		fixed_point_t PROPERTY(size_multiplier);
+		ordered_map<Pop const*, Pop::pop_size_t> PROPERTY(employees);
+	};
+
+	class FactoryProducer final : public NonArtisanalProducer, public Manufacturer {
+	private:
+		static constexpr size_t days_of_history = 7;
+		int PROPERTY(profit_history_current);
+		std::array<fixed_point_t, days_of_history> PROPERTY(daily_profit_history);
```
Hop311, line 41:
```suggestion
		static constexpr size_t DAYS_OF_HISTORY = 7;
		size_t PROPERTY(profit_history_current);
		std::array<fixed_point_t, DAYS_OF_HISTORY> PROPERTY(daily_profit_history);
```
- Although we're not always 100% consistent with this, it's useful to give constants all-caps names so that it's clear that they're not mutable variables.
- Let's make `profit_history_current` a `size_t` so it's the same type as `DAYS_OF_HISTORY` which it'll be compared to. This doesn't take up any more space, as there are currently 4 bytes / 32 bits of padding space after `profit_history_current` so that `daily_profit_history` is at an 8 byte / 64 bit aligned position. By default variables in a struct are aligned to `min(8, their size) rounded up to a power of 2` byte positions, with padding space added in between.

Also `profit_history_current` will need to be initialised to 0 by `FactoryProducer`'s constructor, as MSVC doesn't automatically default initialise primitive variables.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,73 @@
...
+		Good::good_map_t PROPERTY(current_needs);
+	};
+
+	class NonArtisanalProducer : public virtual Producer {
+	public:
+		fixed_point_t PROPERTY(revenue_yesterday);
+		fixed_point_t PROPERTY(output_quantity_yesterday);
+		fixed_point_t PROPERTY(unsold_quantity_yesterday);
+		fixed_point_t PROPERTY(size_multiplier);
+		ordered_map<Pop const*, Pop::pop_size_t> PROPERTY(employees);
```
Hop311, line 34:
```suggestion
		ordered_map<Pop*, Pop::pop_size_t> PROPERTY(employees);
```
The `Pop` pointers here should probably be non-const, so that we can do stuff like give them wages.

Hop311 (commented):
Looks like we've got all the data we need here, now just to make a corresponding `Producer.cpp` file containing constructor implementations and with the implementations of `get_profitability_yesterday` and `get_average_profitability_last_seven_days` moved into it.

One thing to bear in mind: derived class constructors can't directly initialise their base class' variables, they have to call the base class' constructor.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include "openvic-simulation/economy/Good.hpp"
+#include "openvic-simulation/economy/ProductionType.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+namespace OpenVic {
+	class Producer {
+	public:
+		ProductionType const& PROPERTY(production_type);
```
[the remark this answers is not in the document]
wvpm: Tbh I struggle with understand and picking the PROPERTY stuff. It's all very custom.
I agree they should be protected and will change it. As for explicit default, I like it that way. You instantly see the accessibility instead of having to look up/remember the default setting of the container.

src/openvic-simulation/economy/Producer.hpp
```diff
@@ -0,0 +1,73 @@
...
+	public:
+		fixed_point_t PROPERTY(budget);
+		fixed_point_t PROPERTY(balance_yesterday);
+		fixed_point_t PROPERTY(received_investments_yesterday);
+		fixed_point_t PROPERTY(market_spendings_yesterday);
+		fixed_point_t PROPERTY(paychecks_yesterday);
+		uint32_t PROPERTY(unprofitable_days);
+		uint32_t PROPERTY(injected_days);
+		uint32_t PROPERTY(days_without_input);
+		uint8_t PROPERTY_RW_ACCESS(hiring_priority, public);
```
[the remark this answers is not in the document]
wvpm: Why bother with the property stuff if they are just private fields anyway?
I made them properties as I want to expose the getter as public members.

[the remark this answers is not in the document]
Hop311: `PROPERTY` does create public getters, then it switches back to private so that the variables themselves aren't public, which would defeat the purpose of the getters. `PROPERTY_RW` also makes the setter public.

src/openvic-simulation/economy/FactoryProducer.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+	public:
+		typedef ordered_map<Pop* const, Pop::pop_size_t> employees_map_t;
+		typedef std::array<fixed_point_t, DAYS_OF_HISTORY> daily_profit_history_t;
+
+	private:
+		uint8_t PROPERTY(profit_history_current);
+		daily_profit_history_t PROPERTY(daily_profit_history);
+
+	public:
```
Hop311, line 26: `PROPERTY` [always creates `public` getter functions](https://github.com/OpenVicProject/OpenVic-Simulation/blob/master/src/openvic-simulation/utility/Getters.hpp#L109), then resets the accesibility state back to `private`. If the variables themselves are `public` then they can be accessed without the getter functions, and in particular they can be modified externally. All `PROPERTY` variables in this file and the others added in this PR should be `private`, `PROPERTY` will make sure the getters will still be `public`, and if a variable needs to be modified like `hiring_priority`, then it should use `PROPERTY_RW`.

Currently all the variables here apart from `production_type` are already `private`, because of `PROPERTY` setting the accesibility state to `private` after declaring the `public` getter function. The use of `PROPERTY_RW_ACCESS` for `hiring_priority` changes the accessibility status to `public` after the getter function, but not before, so `hiring_priority` is also already `private`. The `*_ACCESS` variants are meant for when the variable isn't `private` (so when it's `protected`), if you're changing the accessibility status to something different from the variable then that should be done with a regular `public` keyword to make it clear that the stuff following it has a different accessibility state.

src/openvic-simulation/economy/ResourceGatheringOpertion.hpp
```diff
@@ -0,0 +1,27 @@
...
+namespace OpenVic {
+	class ResourceGatheringOpertion final {
+	public:
+		typedef ordered_map<Pop* const, Pop::pop_size_t> employees_map_t;
+		ProductionType const& PROPERTY(production_type);
+		fixed_point_t PROPERTY(revenue_yesterday);
+		fixed_point_t PROPERTY(output_quantity_yesterday);
+		fixed_point_t PROPERTY(unsold_quantity_yesterday);
+		fixed_point_t PROPERTY(size_multiplier);
+		ordered_map<Pop* const, Pop::pop_size_t> PROPERTY(employees);
```
Hop311, line 18: `employees` isn't using the `employees_map_t` alias here, which should also be defined with `using` rather than `typedef`. Also the `Pop` pointer should be non-const here too.

src/openvic-simulation/utility/Constructors.hpp
```diff
@@ -0,0 +1,119 @@
+#pragma once
+
+#include <type_traits>
+
+template<typename T>
+constexpr T value_or_move(T&& value) {
+	return std::is_rvalue_reference<T>::value ? std::move(value) : value;
```
Hop311, line 7: This won't work, even if you have a function returning a `&&` reference, the result of that needs to be moved again when initialising the variable. I was thinking to use a separate function for working out if the constructor argument should be an r-value reference or not, so that you wouldn't have to include the argument types in the macro arguments. In this case, the constructor variables should be initialised using:
```c++
ARGUMENT_NAME { \
	std::is_rvalue_reference_v<ARGUMENT_TYPE> ? std::move(ARGUMENT_NAME) : ARGUMENT_NAME \
}
```
This way the argument is actually moved into the variable, instead of being moved into an intermediate variable returned by `value_or_move`, which is then used to copy-construct the variable instead of move-constructing it. This should fix the compilation errors you were having because of  `ProductionType const&`.

src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -0,0 +1,29 @@
...
+	FactoryProducer, ProductionType const&, production_type, const fixed_point_t, size_multiplier, const fixed_point_t,
+	revenue_yesterday, const fixed_point_t, output_quantity_yesterday, const fixed_point_t, unsold_quantity_yesterday,
+	FactoryProducer::employees_map_t&&, employees, Good::good_map_t&&, stockpile, fixed_point_t, budget, const fixed_point_t,
+	balance_yesterday, const fixed_point_t, received_investments_yesterday, const fixed_point_t, market_spendings_yesterday,
+	const fixed_point_t, paychecks_yesterday, const uint32_t, unprofitable_days, const uint32_t, injected_days, const uint32_t,
+	days_without_input, const uint8_t, hiring_priority, const uint8_t, profit_history_current,
+	FactoryProducer::daily_profit_history_t&&, daily_profit_history
+);
+FactoryProducer::FactoryProducer(ProductionType const& new_production_type, const fixed_point_t new_size_multiplier)
+	: FactoryProducer(new_production_type, new_size_multiplier, 0, 0, 0, {}, {}, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, {}) {}
```
Hop311, line 15:
```suggestion
	: FactoryProducer { new_production_type, new_size_multiplier, 0, 0, 0, {}, {}, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, {} } {}
```
In general, [curly brackets should be preferred to round brackets](https://stackoverflow.com/a/18222927) for initialising things, round brackets allow certain implicit type conversions that can hide errors.
`ResourceGatheringOpertion`'s second constructor should also use them.

Spartan322 (commented):
I would also really prefer all the maros be prefixed with OV_ or \_OV depending on expected use case, those which you use directly should be OV_, those you usually don't should be \_OV, and something like value_or_move should be in a namespace even if its in macros like

```cpp
namespace OpenVic::detail {
    template<typename T>
    constexpr T value_or_move(T&& value);
}

#define MACRO(Arg) ::OpenVic::detail::value_or_move(Arg);
```

src/openvic-simulation/economy/FactoryProducer.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "utility/Constructors.hpp"
+
+namespace OpenVic {
+	class FactoryProducer final {
+	private:
+		static constexpr uint8_t DAYS_OF_HISTORY = 7;
+
+	public:
+		typedef ordered_map<Pop* const, Pop::pop_size_t> employees_map_t;
+		typedef std::array<fixed_point_t, DAYS_OF_HISTORY> daily_profit_history_t;
```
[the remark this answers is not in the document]
Spartan322: I don't believe typedef is any less type-safe, its just a pain to manage because its ordered weird and isn't clear about what it refers to, typedef and using aliases are functionally synonyms.

src/openvic-simulation/economy/Producer.hpp
wvpm: I made it `Pop* const` so the pointer is constant but the pop is mutable.

src/openvic-simulation/economy/FactoryProducer.hpp
```diff
@@ -0,0 +1,49 @@
...
+		fixed_point_t PROPERTY(output_quantity_yesterday);
+		fixed_point_t PROPERTY(unsold_quantity_yesterday);
+		fixed_point_t PROPERTY(size_multiplier);
+		ordered_map<Pop*, Pop::pop_size_t> PROPERTY(employees);
+		Good::good_map_t PROPERTY(stockpile);
+		fixed_point_t PROPERTY(budget);
+		fixed_point_t PROPERTY(balance_yesterday);
+		fixed_point_t PROPERTY(received_investments_yesterday);
+		fixed_point_t PROPERTY(market_spendings_yesterday);
+		fixed_point_t PROPERTY(paychecks_yesterday);
```
Hop311, line 27: I would suggest using `wages` instead of `paychecks`, to match Vic2's terminology.

src/openvic-simulation/economy/FactoryProducer.hpp
```diff
@@ -0,0 +1,49 @@
...
+		uint32_t PROPERTY(unprofitable_days);
+		uint32_t PROPERTY(injected_days);
```
Hop311, line 29: I'm not 100% sure what this refers to, but if it's days of being subsidised then I would recommend calling it `subsidised_days`, again matching Vic2's terminology.

src/openvic-simulation/economy/FactoryProducer.hpp
```diff
@@ -0,0 +1,49 @@
...
+	public:
+		FactoryProducer(
+			ProductionType const& new_production_type, const fixed_point_t new_size_multiplier,
+			const fixed_point_t new_revenue_yesterday, const fixed_point_t new_output_quantity_yesterday,
+			const fixed_point_t new_unsold_quantity_yesterday, ordered_map<Pop*, Pop::pop_size_t>&& new_employees,
+			Good::good_map_t&& new_stockpile, fixed_point_t new_budget, const fixed_point_t new_balance_yesterday,
+			const fixed_point_t new_received_investments_yesterday, const fixed_point_t new_market_spendings_yesterday,
+			const fixed_point_t new_paychecks_yesterday, const uint32_t new_unprofitable_days, const uint32_t new_injected_days,
+			const uint32_t new_days_without_input, const uint8_t new_hiring_priority, const uint8_t new_profit_history_current,
+			std::array<fixed_point_t, DAYS_OF_HISTORY>&& new_d
```
Hop311, line 42:
```suggestion
			std::array<fixed_point_t, DAYS_OF_HISTORY>&& new_daily_profit_history
```
Argument name cut short

src/openvic-simulation/economy/ArtisanalProducer.cpp
```diff
@@ -0,0 +1,10 @@
+#include "ArtisanalProducer.hpp"
+
+using namespace OpenVic;
+
+ArtisanalProducer::ArtisanalProducer(
+	ProductionType const& new_production_type, Good::good_map_t&& new_stockpile, const fixed_point_t new_current_production,
+	Good::good_map_t&& new_current_needs
+)
+	: production_type { new_production_type }, stockpile { new_stockpile }, current_production { new_current_production },
+	  current_needs { std::move(new_current_needs) } {}
```
Hop311, line 10:
```suggestion
	: production_type { new_production_type }, stockpile { std::move(new_stockpile) },
	  current_production { new_current_production }, current_needs { std::move(new_current_needs) } {}
```
`new_stockpile` needs moving

src/openvic-simulation/economy/FactoryProducer.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include "openvic-simulation/economy/ProductionType.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+namespace OpenVic {
+	class FactoryProducer final {
+	private:
+		static constexpr uint8_t DAYS_OF_HISTORY = 7;
+		uint8_t PROPERTY(profit_history_current);
+		std::array<fixed_point_t, DAYS_OF_HISTORY> PROPERTY(daily_profit_history);
```
Hop311, line 15: `std::array<fixed_point_t, DAYS_OF_HISTORY>` is a good candidate for a type alias, for example:
```c++
using daily_profit_history_t = std::array<fixed_point_t, DAYS_OF_HISTORY>;
```

src/openvic-simulation/economy/FactoryProducer.hpp
wvpm: I copied the term straight from a savegame. I prefer your suggestion though.

src/openvic-simulation/economy/FactoryProducer.hpp
wvpm: Both the UI and savegame format use paychecks. See http://www.amicabile.com/v2/f1/B2.jpg
I'll leave it paychecks for now.

Hop311 (approved):
Great work! Once this is merged, "draft for factory production" #144 will automatically re-target itself onto `master`, however you'll still need to rebase it to fix merge conflicts.

PR #127, opened 2024-01-12
Added NationalFocus ModifierEffects + bool variables

Hop311:
- Added `<pop_type> = 0.2` (means 20%) and `<good> = 20` (means 20%) modifiers, replacing custom system used for them in `NationalFocus`.
- Added `has_flashpoint` (default false), `own_provinces` (default true), and `outliner_show_as_percent` (default false) bool variables to `NationalFocus`

### DIFF
--- a/src/openvic-simulation/politics/NationalFocus.cpp
+++ b/src/openvic-simulation/politics/NationalFocus.cpp
@@ -7,20 +7,22 @@ NationalFocusGroup::NationalFocusGroup(std::string_view new_identifier) : HasIde
 
 NationalFocus::NationalFocus(
 	std::string_view new_identifier,
-	uint8_t new_icon,
 	NationalFocusGroup const& new_group,
+	uint8_t new_icon,
+	bool new_has_flashpoint,
+	bool new_own_provinces,
+	bool new_outliner_show_as_percent,
 	ModifierValue&& new_modifiers,
-	pop_promotion_map_t&& new_encouraged_promotion,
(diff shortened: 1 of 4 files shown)


src/openvic-simulation/politics/NationalFocus.cpp
```diff
@@ -90,40 +93,28 @@ bool NationalFocusManager::load_national_foci_file(
...
-						return expect_fixed_point(map_callback(promotions, pop_type))(value);
-					}
-					Good const* good = good_manager.get_good_by_identifier(key);
-					if (good != nullptr) {
-						return expect_fixed_point(map_callback(production, good))(value);
-					}
-					return key_value_invalid_callback(key, value);
-				},
 				"icon", ONE_EXACTLY, expect_uint(assign_variable_callback(icon)),
 				"ideology", ZERO_OR_ONE,
```
wvpm, line 107: ideology & loyalty_value together are also a state modifier.
They add party loyalty over time. Besides that, looks good.

wvpm (approved):

src/openvic-simulation/politics/NationalFocus.cpp
Hop311: Yeah I wanted to make that a `ModifierEffect` too but there's no way to do it within the modifier framework, everything is based on individual keys, whether that has a number as its value or a list of sub-modifiers, but `ideology` and `loyalty_value` are at the same level.

PR #131, opened 2024-01-17
Parse missing variables, Logger counting, misc cleanup

Hop311:
- Removed the need to include the `mod` folder when specifying mods to be loaded in headless mode.
- Added message count tracking to `Logger`, and a summary output at the end of headless mode (useful for checking if a change increased the number of warnings or errors).
- Implemented missing parsing for some basic defines variables (and added comments explaining why others weren't being loaded):
  - Loaded `primary` and `union` countries for `Culture`s and `CultureGroup`s, which required moving culture loading to be after country loading. Also loaded `radicalism` for `Culture`.
  - Loaded `picture` string for OOB `Leader`s.
  - Loaded `PopType` incomes from national budget, research and leadership points and optimum ratio for whichever points a `PopType` generates, tax efficiency and equivalent `PopType` (used for `labourers` <-> `farmers` conversion).
- Changed set and map callbacks to support warnings or errors based on a bool argument, all using a common `warn_or_error` function which `UniqueKeyRegistry::expect_item_str` now also uses.
- Small bits of format cleanup across the repo.
- Added default map colours (so that uncolonised land isn't shown as natural terrain in the political mapmode).

### DIFF
--- a/src/openvic-simulation/economy/ProductionType.cpp
+++ b/src/openvic-simulation/economy/ProductionType.cpp
@@ -4,23 +4,30 @@ using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
 Job::Job(
-	PopType const* const new_pop_type, effect_t new_effect_type, fixed_point_t new_effect_multiplier,
+	PopType const* new_pop_type, effect_t new_effect_type, fixed_point_t new_effect_multiplier,
 	fixed_point_t new_desired_workforce_share
-)
-	: pop_type { new_pop_type }, effect_type { new_effect_type }, effect_multiplier { new_effect_multiplier },
-	  desired_workforce_share { new_desired_workforce_share } {}
+) : pop_type { new_pop_type }, effect_type { new_effect_type }, effect_multiplier { new_effect_multiplier },
+	desired_workforce_share { new_desired_workforce_share } {}
 
 ProductionType::ProductionType(
-	std::string_view new_identifier, Job new_owner, std::vector<Job> new_jobs, template_type_t new_template_type,
-	Pop::pop_size_t new_base_workforce_size, Good::good_map_t&& new_input_goods, Good const* const new_output_goods,
-	fixed_point_t new_base_output_quantity, std::vector<bonus_t>&& new_bonuses, Good::good_map_t&& new_maintenance_requirements,
-	bool new_is_coastal, bool new_is_farm, bool new_is_mine
-)
-	: HasIdentifier { new_identifier }, owner { new_owner }, jobs { new_jobs }, template_type { new_template_type },
-	  base_workforce_size { new_base_workforce_size }, input_goods { std::move(new_input_goods) },
-	  output_goods { new_output_goods }, base_output_quantity { new_base_output_quantity }, bonuses { std::move(new_bonuses) },
-	  maintenance_requirements { std::move(new_maintenance_requirements) }, coastal { new_is_coastal }, farm { new_is_farm },
-	  mine { new_is_mine } {}
+	std::string_view new_identifier,
+	std::optional<Job> new_owner,
+	std::vector<Job>&& new_jobs,
+	template_type_t new_template_type,
+	Pop::pop_size_t new_base_workforce_size,
+	Good::good_map_t&& new_input_goods,
+	Good const* new_output_goods,
+	fixed_point_t new_base_output_quantity,
+	std::vector<bonus_t>&& new_bonuses,
+	Good::good_map_t&& new_maintenance_requirements,
+	bool new_is_coastal,
+	bool new_is_farm,
+	bool new_is_mine
+) : HasIdentifier { new_identifier }, owner { new_owner }, jobs { std::move(new_jobs) }, template_type { new_template_type },
+	base_workforce_size { new_base_workforce_size }, input_goods { std::move(new_input_goods) },
+	output_goods { new_output_goods }, base_output_quantity { new_base_output_quantity }, bonuses { std::move(new_bonuses) },
+	maintenance_requirements { std::move(new_maintenance_requirements) }, coastal { new_is_coastal }, farm { new_is_farm },
+	mine { new_is_mine } {}
 
 bool ProductionType::parse_scripts(GameManager const& game_manager) {
 	bool ret = true;
@@ -33,10 +40,9 @@ bool ProductionType::parse_scripts(GameManager const& game_manager) {
 ProductionTypeManager::ProductionTypeManager() : rgo_owner_sprite { 0 } {}
 
 node_callback_t ProductionTypeManager::_expect_job(
-	GoodManager const& good_manager, PopManager const& pop_manager, callback_t<Job&&> cb
+	GoodManager const& good_manager, PopManager const& pop_manager, callback_t<Job&&> callback
 ) {
-
-	return [this, &good_manager, &pop_manager, cb](ast::NodeCPtr node) -> bool {
+	return [this, &good_manager, &pop_manager, callback](ast::NodeCPtr node) -> bool {
 		using enum Job::effect_t;
 
 		std::string_view pop_type {};
@@ -54,27 +60,36 @@ node_callback_t ProductionTypeManager::_expect_job(
 			"amount", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(desired_workforce_share))
 		)(node);
 
-		PopType const* const found_pop_type = pop_manager.get_pop_type_by_identifier(pop_type);
-		return res & cb({ found_pop_type, effect_type, effect_multiplier, desired_workforce_share });
+		PopType const* found_pop_type = pop_manager.get_pop_type_by_identifier(pop_type);
+		return res & callback({ found_pop_type, effect_type, effect_multiplier, desired_workforce_share });
 	};
 }
 
 node_callback_t ProductionTypeManager::_expect_job_list(
-	GoodManager const& good_manager, PopManager const& pop_manager, callback_t<std::vector<Job>&&> cb
+	GoodManager const& good_manager, PopManager const& pop_manager, callback_t<std::vector<Job>&&> callback
 ) {
-	return [this, &good_manager, &pop_manager, cb](ast::NodeCPtr node) -> bool {
+	return [this, &good_manager, &pop_manager, callback](ast::NodeCPtr node) -> bool {
 		std::vector<Job> jobs;
 		bool ret = expect_list(_expect_job(good_manager, pop_manager, vector_callback(jobs)))(node);
-		ret &= cb(std::move(jobs));
+		ret &= callback(std::move(jobs));
 		return ret;
 	};
 }
 
 bool ProductionTypeManager::add_production_type(
-	std::string_view identifier, Job owner, std::vector<Job> jobs, ProductionType::template_type_t template_type,
-	Pop::pop_size_t base_workforce_size, Good::good_map_t&& input_goods, Good const* const output_goods,
-	fixed_point_t base_output_quantity, std::vector<ProductionType::bonus_t>&& bonuses,
-	Good::good_map_t&& maintenance_requirements, bool is_coastal, bool is_farm, bool is_mine
+	std::string_view identifier,
+	std::optional<Job> owner,
+	std::vector<Job>&& jobs,
+	ProductionType::template_type_t template_type,
+	Pop::pop_size_t base_workforce_size,
+	Good::good_map_t&& input_goods,
+	Good const* output_goods,
+	fixed_point_t base_output_quantity,
+	std::vector<ProductionType::bonus_t>&& bonuses,
+	Good::good_map_t&& maintenance_requirements,
+	bool is_coastal,
+	bool is_farm,
+	bool is_mine
 ) {
 	if (identifier.empty()) {
 		Logger::error("Invalid production type identifier - empty!");
@@ -96,13 +111,30 @@ bool ProductionTypeManager::add_production_type(
 		return false;
 	}
 
-	if (template_type == ProductionType::template_type_t::ARTISAN) {
-		if (owner.get_pop_type() != nullptr || !jobs.empty()) {
-			Logger::warning("Artisanal production types don't use owner and employees. Effects are ignored.");
+	using enum ProductionType::template_type_t;
+
+	if (template_type == ARTISAN) {
+		if (owner.has_value()) {
+			Logger::warning(
+				"Artisanal production type ", identifier, " should not have an owner - it is being ignored."
+			);
+			owner.reset();
+		}
+
+		if (!jobs.empty()) {
+			Logger::warning(
+				"Artisanal production type ", identifier, " should not have employees - ", jobs.size(), " are being ignored."
+			);
+			jobs.clear();
 		}
 	} else {
-		if (owner.get_pop_type() == nullptr) {
-			Logger::error("Production type ", identifier, " lacks owner or has an invalid pop type.");
+		if (!owner.has_value()) {
+			Logger::error("Production type ", identifier, " is missing an owner.");
+			return false;
+		}
+
+		if (owner->get_pop_type() == nullptr) {
+			Logger::error("Production type ", identifier, " owner has an invalid pop type.");
 			return false;
 		}
 
@@ -120,12 +152,12 @@ bool ProductionTypeManager::add_production_type(
 	}
 
 	const bool ret = production_types.add_item({
-		identifier, owner, jobs, template_type, base_workforce_size, std::move(input_goods),
-		output_goods, base_output_quantity, std::move(bonuses), std::move(maintenance_requirements), is_coastal, is_farm, is_mine
+		identifier, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), output_goods,
+		base_output_quantity, std::move(bonuses), std::move(maintenance_requirements), is_coastal, is_farm, is_mine
 	});
-	if (rgo_owner_sprite <= 0 && ret && template_type == ProductionType::template_type_t::RGO && owner.get_pop_type() != nullptr) {
+	if (rgo_owner_sprite <= 0 && ret && template_type == RGO && owner.has_value() && owner->get_pop_type() != nullptr) {
 		/* Set rgo owner sprite to that of the first RGO owner we find. */
-		rgo_owner_sprite = owner.get_pop_type()->get_sprite();
+		rgo_owner_sprite = owner->get_pop_type()->get_sprite();
 	}
 	return ret;
 }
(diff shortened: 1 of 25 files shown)


src/openvic-simulation/economy/ProductionType.cpp
```diff
@@ -96,7 +94,14 @@ bool ProductionTypeManager::add_production_type(
 		return false;
 	}
 
-	if (template_type == ProductionType::template_type_t::ARTISAN) {
+	using enum ProductionType::template_type_t;
+
+	if (template_type == ARTISAN) {
+		if (owner.get_pop_type() != nullptr && owner.get_pop_type()->get_is_artisan()) {
```
wvpm, line 100: Discusssed on discord.
Specifying any pop here is meaningless and silently accepting it can create the impression that it does have an effect.
For modders: either leave out the owner scope or leave 'artisan'* as vanilla.

*technically any name that is not a pop type is fine.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -132,6 +326,20 @@ bool PopManager::add_pop_type(
...
+
+static NodeCallback auto expect_needs_income(PopType::income_type_t& type, fixed_point_t& weight) {
+	static const string_map_t<PopType::income_type_t> income_type_map {
+		{ "administration", ADMINISTRATION },
+		{ "education", EDUCATION },
+		{ "military", MILITARY },
+		{ "reforms", REFORMS }
+	};
+	return expect_dictionary_keys(
+		"type", ONE_EXACTLY, expect_identifier(expect_mapped_string(income_type_map, assign_variable_callback(type))),
```
wvpm, line 340: There can be multiple types!!
Victoria 2 treats it as a list of types and you get paid by each of them (great for making certain pops rich).

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -132,6 +326,20 @@ bool PopManager::add_pop_type(
...
+		"weight", ONE_EXACTLY, expect_fixed_point(assign_variable_callback(weight))
```
wvpm, line 341: Zero or one and discard the result (no point storing it in the pop type).

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -145,11 +353,15 @@ bool PopManager::load_pop_type_file(
 	Strata const* strata = nullptr;
 	PopType::sprite_t sprite = 0;
 	Good::good_map_t life_needs, everyday_needs, luxury_needs;
+	PopType::income_type_t life_needs_income_type = NONE, everyday_needs_income_type = NONE, luxury_needs_income_type = NONE;
+	fixed_point_t life_needs_income_weight = 0, everyday_needs_income_weight = 0, luxury_needs_income_weight = 0;
 	PopType::rebel_units_t rebel_units;
+	Pop::pop_size_t max_size = 0, merge_max_size = 0;
```
wvpm, line 359: max value of pop_size_t as default?
using 0 as there is no max adds complexity.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -145,11 +353,15 @@ bool PopManager::load_pop_type_file(
...
 	bool state_capital_only = false, demote_migrant = false, is_artisan = false, allowed_to_vote = true, is_slave = false,
 		can_be_recruited = false, can_reduce_consciousness = false, administrative_efficiency = false, can_build = false,
 		factory = false, can_work_factory = false, unemployment = false;
-	Pop::pop_size_t max_size = 0, merge_max_size = 0;
+	fixed_point_t research_points = 0, leadership_points = 0, research_leadership_optimum = 0, tax_efficiency = 0;
```
wvpm, line 363: I don't think tax_efficiency has any effect.
I'll check it later. If it is meaningless, we can log a warning (vanilla excluded).

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -206,7 +418,7 @@ bool PopManager::load_pop_type_file(
 		"issues", ZERO_OR_ONE, assign_variable_callback(issues_node),
 		"demote_migrant", ZERO_OR_ONE, expect_bool(assign_variable_callback(demote_migrant)),
 		"administrative_efficiency", ZERO_OR_ONE, expect_bool(assign_variable_callback(administrative_efficiency)),
-		"tax_eff", ZERO_OR_ONE, success_callback, // TODO - tax collection modifier
+		"tax_eff", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(tax_efficiency)),
 		"can_build", ZERO_OR_ONE, expect_bool(assign_variable_callback(can_build)),
 		"factory", ZERO_OR_ONE, expect_bool(assign_variable_callback(factory)),
 		"workplace_input", ZERO_OR_ONE, success_callback, // TODO - work out what these do
```
wvpm, line 424: meaningless

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -206,7 +418,7 @@ bool PopManager::load_pop_type_file(
 		"issues", ZERO_OR_ONE, assign_variable_callback(issues_node),
 		"demote_migrant", ZERO_OR_ONE, expect_bool(assign_variable_callback(demote_migrant)),
 		"administrative_efficiency", ZERO_OR_ONE, expect_bool(assign_variable_callback(administrative_efficiency)),
-		"tax_eff", ZERO_OR_ONE, success_callback, // TODO - tax collection modifier
+		"tax_eff", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(tax_efficiency)),
 		"can_build", ZERO_OR_ONE, expect_bool(assign_variable_callback(can_build)),
 		"factory", ZERO_OR_ONE, expect_bool(assign_variable_callback(factory)),
```
wvpm, line 423: I think this is only used for the UI 'unemployed workers' icon.
Need to check. If so, please change the name in code.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -206,7 +418,7 @@ bool PopManager::load_pop_type_file(
 		"issues", ZERO_OR_ONE, assign_variable_callback(issues_node),
 		"demote_migrant", ZERO_OR_ONE, expect_bool(assign_variable_callback(demote_migrant)),
 		"administrative_efficiency", ZERO_OR_ONE, expect_bool(assign_variable_callback(administrative_efficiency)),
-		"tax_eff", ZERO_OR_ONE, success_callback, // TODO - tax collection modifier
+		"tax_eff", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(tax_efficiency)),
 		"can_build", ZERO_OR_ONE, expect_bool(assign_variable_callback(can_build)),
```
wvpm, line 422: can_build is somewhat misleading, this determines whether the pop can invest into factories & railroads (depending on economic policy)
is_capitalist? (might be too strict)
is_private_investor? (might be confused with shadowy financeers)
What do you think is a good name?
maybe can_build is fine. Just think about it for a second and see if you know a better name.

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -73,6 +79,8 @@ namespace OpenVic {
 	struct PopType : HasIdentifierAndColour {
 		friend struct PopManager;
 
+		enum struct income_type_t : uint8_t { NONE, ADMINISTRATION, EDUCATION, MILITARY, REFORMS };
```
wvpm, line 82: bit flags?

src/openvic-simulation/research/Invention.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "openvic-simulation/economy/BuildingType.hpp"
 #include "openvic-simulation/map/Crime.hpp"
+#include "openvic-simulation/military/Unit.hpp"
```
wvpm, line 5: Why?

Perhaps this is a leftover from another branch?

src/openvic-simulation/utility/StringUtils.hpp
```diff
@@ -1,10 +1,11 @@
 #pragma once
 
+#include <algorithm>
```
wvpm, line 3: Why change includes?
Did you mean to include code changes or was there a compilation issue or something with intellisense?

wvpm (changes requested):

src/openvic-simulation/economy/ProductionType.cpp
Hop311: Since owners don't do anything in vanilla and should be removed from there too, e.g. if someone is making a mod based off of vanilla, I think the warning should apply for any owner being given, regardless of whether the pop type is `artisan`, a real pop type, or any other key.

src/openvic-simulation/pop/Culture.cpp
```diff
@@ -62,7 +65,12 @@ bool CultureManager::add_culture(
 		Logger::error("Invalid culture identifier - empty!");
 		return false;
 	}
-	return cultures.add_item({ identifier, colour, group, std::move(first_names), std::move(last_names) });
+
+	// TODO - check radicalism range
```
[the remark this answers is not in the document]
Hop311: I've been thinking about how to handle fixed points with limited ranges, we could make a template type that takes the min and max as arguments and automatically clamps it every time any change is made. This will be very useful for stuff like militancy and conciousness, and anything that can't be negative.

src/openvic-simulation/pop/Pop.cpp
Hop311: How would we detect if it was vanilla or not?

src/openvic-simulation/pop/Pop.cpp
Hop311: I'll change it to `can_invest`

src/openvic-simulation/pop/Pop.hpp
Hop311: If they can't be used simulatenously sure, for now I'm storing them in a vector to allow for repeats

src/openvic-simulation/research/Invention.cpp
Hop311: No, I removed an include in another place and replaced it with a forward declaration, this file just happened to rely on that removed include.

src/openvic-simulation/utility/StringUtils.hpp
Hop311: This was added to make intellisense calm down about something, also in general it's good practice to include stuff directly in the file even if it's also included via other headers. I'm not as strict about it as I should be, Spartan's usually the most on it.

src/openvic-simulation/pop/Pop.cpp
Hop311: What do you mean? None of these do anything?
```
workplace_input = -0.1
workplace_output = 0
starter_share = 1.0
```

src/openvic-simulation/pop/Pop.hpp
wvpm: Each type can only be used once. Bit flags seems a great solution for that.

src/openvic-simulation/pop/Pop.cpp
wvpm: In vanilla only bureaucrats have it with value 1.0

src/openvic-simulation/pop/Pop.cpp
wvpm: None of them have an effect

src/openvic-simulation/economy/ProductionType.cpp
```diff
@@ -96,13 +111,30 @@ bool ProductionTypeManager::add_production_type(
...
+				"Artisanal production type ", identifier, " should not have an owner - it is being ignored."
+			);
+			owner.reset();
+		}
+
+		if (!jobs.empty()) {
+			Logger::warning(
+				"Artisanal production type ", identifier, " should not have employees - ", jobs.size(), " are being ignored."
+			);
+			jobs.clear();
```
wvpm, line 128: good idea to clear them rather than just logging a warning

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -73,7 +83,10 @@ namespace OpenVic {
 	struct PopType : HasIdentifierAndColour {
 		friend struct PopManager;
 
+		enum struct income_type_t : uint8_t { ADMINISTRATION, EDUCATION, MILITARY, REFORMS };
+
 		using sprite_t = uint8_t;
+		using income_type_vector_t = std::vector<income_type_t>;
```
wvpm, line 89: Shouldn't be a vector as the values have to be unique.

wvpm (approved):

src/openvic-simulation/pop/Pop.cpp
wvpm: `factory` & `can_work_factory` are rather confusing terms. I'm looking into what they do and will update this comment.

src/openvic-simulation/pop/Pop.cpp
schombert: starter share has to do with how much money pops that become capitalists magically get (yeah, promoting to a capitalist creates cash from nothing). This is required so that if a state doesn't start with capitalists and is LF it is possible for those capitalists to build a factory even though there are no current factories to provide them with an income. Usually it isn't enough and you see multiple waves of capitalist demotion/promotion cycles before they get enough money to finish the project.

src/openvic-simulation/pop/Pop.cpp
schombert: tax efficiency determines how much the pop contributes to state admin efficiency (not national admin efficiency, the one that you can see in the province window that varies from state to state) which in turn determines crime fighting efficiency

wvpm: Thanks, I tested it and it works.
State efficiency from bureacrats = tax_eff * bureaucrat state population % / defines.country.MAX_BUREAUCRACY_PERCENTAGE

src/openvic-simulation/pop/Pop.cpp
wvpm: I tested it but couldn't find any magic money.
New capitalists start with the savings & cash from their previous pop. start_share has no impact on that.

schombert: Well, I trust that the difference will start to show up when the economy is implemented and you can start doing side-by-side comparisons of Victoria 2 saves and open vic saves. Just keep it in mind for then.

src/openvic-simulation/economy/ProductionType.cpp
```diff
@@ -4,23 +4,30 @@ using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
 Job::Job(
-	PopType const* const new_pop_type, effect_t new_effect_type, fixed_point_t new_effect_multiplier,
```
wvpm, line 7: What's wrong with the const here?

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -97,27 +192,100 @@ bool PopManager::add_pop_type(
...
+	}
+	if (
+		(life_needs_income_types == NO_INCOME_TYPE) != (everyday_needs_income_types == NO_INCOME_TYPE) ||
+		(everyday_needs_income_types == NO_INCOME_TYPE) != (luxury_needs_income_types == NO_INCOME_TYPE)
+	) {
+		Logger::warning(
+			"Pop type ", identifier, " has inconsistent income types - all must be none, or all must be not none: life = ",
+			life_needs_income_types, ", everyday = ", everyday_needs_income_types, ", luxury = ", luxury_needs_income_types
+		);
+	}
```
wvpm, line 206: Please add new line between if statements unless they are related via else if.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -97,27 +192,100 @@ bool PopManager::add_pop_type(
...
+	if (max_size <= 0) {
+		Logger::error("Invalid pop type max size for ", identifier, ": ", max_size, " (must be positive)");
+		return false;
+	}
+	if (merge_max_size <= 0) {
+		Logger::error("Invalid pop type merge max size for ", identifier, ": ", merge_max_size, " (must be positive)");
+		return false;
+	}
+
+	if (research_points < 0) {
```
wvpm, line 216: Could be negative I think. Imagine a mod wants to simulate saboteur pops or something.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -97,27 +192,100 @@ bool PopManager::add_pop_type(
...
+		Logger::error("Invalid pop type research points for ", identifier, ": ", research_points, " (cannot be negative)");
+		return false;
+	}
+	if (leadership_points < 0) {
```
wvpm, line 220: Also might be negative.

src/openvic-simulation/pop/Pop.cpp
Hop311: Does that work in Vic2?

src/openvic-simulation/pop/Pop.cpp
Hop311: Does that work in Vic2?

src/openvic-simulation/economy/ProductionType.cpp
Hop311: It wasn't needed, just adds clutter, and the corresponding const had already been removed from the header file

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -97,27 +192,91 @@ bool PopManager::add_pop_type(
...
 		return false;
 	}
+	if (research_leadership_optimum < 0) {
+		Logger::error(
+			"Invalid pop type research/leadership optimum for ", identifier, ": ", research_leadership_optimum,
+			" (cannot be negative)"
+		);
+		return false;
+	}
+	if ((research_points > 0 || leadership_points > 0) != (research_leadership_optimum > 0)) {
```
wvpm:
```suggestion
	if ((research_points != 0 || leadership_points != 0) != (research_leadership_optimum > 0)) {
```
Only non-zero matters here. Other checks already handle negative values.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -132,6 +276,29 @@ bool PopManager::add_pop_type(
...
+			[&types](PopType::income_type_t type) -> bool {
+				if (!share_income_type(types, type)) {
+					types |= type;
+					return true;
+				}
+				Logger::error("Duplicate income type ", type, " in pop type income types!");
+				return false;
+			}
+		)),
+		"weight", ZERO_OR_ONE, success_callback /* Has no effect in game */
```
wvpm, line 300: Should we add a warning for weight != 1?
If it is added with a different value then the modder could be informed that it has no effect.

wvpm (approved):

src/openvic-simulation/pop/Pop.cpp
Hop311: Vanilla adds it with values other than 1 (e.g. soldiers have 1.5), it's just gonna create warning spam.

PR #132, opened 2024-01-18
Minor enum cleanup, changed string functions into stream overloads

Hop311:


src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -6,6 +6,10 @@
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
+using enum value_type_t;
```
wvpm, line 9: Isn't this what spartan calls "polluting the global scope"?

wvpm (approved):

src/openvic-simulation/scripts/Condition.cpp
Hop311: This is inside a `cpp` file, so it doesn't affect any other file. Almost every single function in that file uses those 3 enums, it's no more polluting than the two `namespace` includes above (actually with fewer potential negative consequences, since an enum can't contain functions).

PR #133, opened 2024-01-19
Harmonize empty result behavior for `_search_for_game_path`

Spartan322:
Fixes OpenVicProject/OpenVic#195

Enables actually searching for game path
Reports info or warning depending on why it fails

Remove StringUtils.hpp include in Vic2PathSearch.cpp
Add <cstdint> to StringUtils.hpp


Hop311 (dismissed):
Thanks for fixing this, the search with `".."` as hint works on my machine as well 👍

Hop311 (approved):

PR #141, opened 2024-01-27
Add GapDeque

Spartan322:
See #170
Add utility::any_of concept

### DIFF
--- a/src/openvic-simulation/types/GapDeque.hpp
+++ b/src/openvic-simulation/types/GapDeque.hpp
@@ -1,0 +1,936 @@
+#pragma once
+
+// Based heavily on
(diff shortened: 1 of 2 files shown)


wvpm:
Would it be possible/feasible to write unit tests for the GapDeque?
I think the code looks alright but I can't predict how well it will work.

Spartan322:
Does kinda mean we'll have to integrate a unit test system, I don't think that's the responsibility of this PR at least.

Spartan322:
Currently preferring to investigate the viability of https://github.com/mattreecebentley/plf_hive instead.

Spartan322:
Should be resolved by #170.

PR #144, opened 2024-01-31
draft for factory production

wvpm:
TODO:
- hiring & firing
- maintenance/upkeep goods
- optimisations
- state code
- country code

### DIFF
--- a/src/openvic-simulation/economy/FactoryProducer.cpp
+++ b/src/openvic-simulation/economy/FactoryProducer.cpp
@@ -1,5 +1,10 @@
 #include "FactoryProducer.hpp"
 
+#include "openvic-simulation/economy/Good.hpp"
+#include "openvic-simulation/pop/Pop.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+
+
 using namespace OpenVic;
 
 FactoryProducer::FactoryProducer(
@@ -20,7 +25,8 @@ FactoryProducer::FactoryProducer(
 	  market_spendings_yesterday { new_market_spendings_yesterday }, paychecks_yesterday { new_paychecks_yesterday },
 	  unprofitable_days { new_unprofitable_days }, subsidised_days { new_subsidised_days },
 	  days_without_input { new_days_without_input }, hiring_priority { new_hiring_priority },
-	  profit_history_current { new_profit_history_current }, daily_profit_history { std::move(new_daily_profit_history) } {}
+	  profit_history_current { new_profit_history_current }, daily_profit_history { std::move(new_daily_profit_history) },
+	  employees_per_job_cache {} {}
 FactoryProducer::FactoryProducer(ProductionType const& new_production_type, const fixed_point_t new_size_multiplier)
 	: FactoryProducer { new_production_type, new_size_multiplier, 0, 0, 0, {}, {}, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, {} } {}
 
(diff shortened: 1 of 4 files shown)


src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -37,3 +44,64 @@ fixed_point_t FactoryProducer::get_average_profitability_last_seven_days() const
 
 	return sum / (1 + profit_history_current);
 }
+
+void FactoryProducer::produce(
+	const fixed_point_t input_modifier, const fixed_point_t throughput_modifier, const fixed_point_t output_modifier
+) {
+	std::map<PopType const* const, fixed_point_t> employees_per_job {};
```
wvpm, line 51: This could be cached as a private field.

src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -37,3 +44,64 @@ fixed_point_t FactoryProducer::get_average_profitability_last_seven_days() const
 
 	return sum / (1 + profit_history_current);
 }
+
+void FactoryProducer::produce(
+	const fixed_point_t input_modifier, const fixed_point_t throughput_modifier, const fixed_point_t output_modifier
```
wvpm, line 49: I imagine the country calls some method on each state, which calls this method for each factory.
The country & state create the modifier values here.

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -30,7 +30,7 @@ namespace OpenVic {
 	struct Pop {
 		friend struct PopManager;
 
-		using pop_size_t = int64_t;
+		using pop_size_t = int32_t;
```
wvpm, line 33: I had to change this as fixed_point_t can't handle int64. Also pop size shouldn't exceed int32.

src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -37,3 +44,71 @@ fixed_point_t FactoryProducer::get_average_profitability_last_seven_days() const
...
+			my_input_multiplier += job.get_effect_multiplier() * number_of_employees_in_job /
+				(size_multiplier * production_type.get_base_workforce_size());
+			break;
+		case Job::effect_t::OUTPUT:
+			my_output_multiplier += job.get_effect_multiplier() * number_of_employees_in_job /
+				(size_multiplier * production_type.get_base_workforce_size());
+			break;
+		case Job::effect_t::THROUGHPUT:
+			my_throughput_multiplier *=
+				job.get_effect_multiplier() * number_of_employees_in_job / production_type.get_base_workforce_size();
```
wvpm, line 80: Note size_multiplier is left out here since it would be cancelled by multiplying the base input & output quantities.

src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -37,3 +44,71 @@ fixed_point_t FactoryProducer::get_average_profitability_last_seven_days() const
 
 	return sum / (1 + profit_history_current);
 }
+
+void FactoryProducer::produce(
+	const fixed_point_t input_modifier, const fixed_point_t throughput_modifier, const fixed_point_t output_modifier
+) {
+	std::map<PopType const* const, fixed_point_t> employees_per_job {};
```
Hop311, line 51: Better to use `ordered_map` from `openvic-simulation/types/OrderedContainers.hpp`, we've switched to using it in the rest of the repo over the `std` maps and sets both for performance (it's a hash set) and ordering control (`std` maps only allow either no order or key comparison order).

src/openvic-simulation/economy/FactoryProducer.cpp
Hop311: Yeah this definitely should be cached, we'll also need it for tooltips

src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -37,3 +44,71 @@ fixed_point_t FactoryProducer::get_average_profitability_last_seven_days() const
...
+	for (const Job job : production_type.get_jobs()) {
```
Hop311, line 52:
```suggestion
	for (Job const& job : production_type.get_jobs()) {
```
Every time a non-pointer or reference variable is created, it's entire contents is copied to a new block of memory. Better to use references and avoid creating unnecessary intermediate variables. Same for the other `for` loops in this file.

src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -37,3 +44,71 @@ fixed_point_t FactoryProducer::get_average_profitability_last_seven_days() const
...
+		case Job::effect_t::INPUT:
+			my_input_multiplier += job.get_effect_multiplier() * number_of_employees_in_job /
+				(size_multiplier * production_type.get_base_workforce_size());
+			break;
+		case Job::effect_t::OUTPUT:
+			my_output_multiplier += job.get_effect_multiplier() * number_of_employees_in_job /
+				(size_multiplier * production_type.get_base_workforce_size());
+			break;
+		case Job::effect_t::THROUGHPUT:
+			my_throughput_multiplier *=
```
Hop311, line 79: So throughput job effects are multiplicative?

src/openvic-simulation/economy/FactoryProducer.cpp
```diff
@@ -37,3 +44,71 @@ fixed_point_t FactoryProducer::get_average_profitability_last_seven_days() const
...
+
+	for (std::pair<Pop const* const, const fixed_point_t> pop_and_number_employed : employees) {
+		Pop const* const pop = pop_and_number_employed.first;
+		const fixed_point_t number_employed = pop_and_number_employed.second;
+		PopType const* const pop_type = &(pop->get_type());
+		employees_per_job[pop_type] += number_employed;
+	}
+
+	fixed_point_t my_input_multiplier = input_modifier, my_throughput_multiplier = throughput_modifier,
+				  my_output_multiplier = output_modifier;
```
Hop311, line 65: These are assumed to include the base `1.0` / `100%`, does that come from any value in the defines (e.g. a default input modifier of `100%`) or will it need to be used as the base when calculating the sum of exernal modifiers?

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -534,7 +534,7 @@ bool PopManager::load_pop_into_vector(
 	}
 
 	if (culture != nullptr && religion != nullptr && size >= 1) {
```
Hop311, line 536:
```suggestion
	if (culture != nullptr && religion != nullptr && size >= 1 && size <= std::numeric_limits<Pop::pop_size_t>::max()) {
```
Can you add a max check here

Hop311 (commented):
Great work with the equations, nice to see that we already have all these components available.

src/openvic-simulation/economy/FactoryProducer.cpp
wvpm: I think tooltips should be calculated on demand. It saves a memory and ensures the tooltips are up to date.

src/openvic-simulation/economy/FactoryProducer.cpp
wvpm: Yes. Output scales directly with the amount of throughput employees.

src/openvic-simulation/economy/FactoryProducer.cpp
wvpm: As mentioned in a different comment, the parameters of the produce method comes from the state. The state will have a similar method that gets its parameters from a method in the country. The country has to handle the `1 + modifiers`.

PR #148, opened 2024-02-18
Regions are now coloured from colors.txt

Jarturog:


src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -679,6 +680,7 @@ bool Dataloader::_load_map_dir(GameManager& game_manager) const {
 	static constexpr std::string_view default_continent = "continent.txt";
 	static constexpr std::string_view default_adjacencies = "adjacencies.csv";
 	static constexpr std::string_view default_region = "region.txt";
+	static constexpr std::string_view default_region_colours = "colors.txt"; /* File from interface_directory */
```
Hop311, line 683: The region colours filepath is hardcoded, it can't be customised in Vic2's `default.map`. Move this line to the bottom of this block (next to `climate_file`) and rename it from `default_region_colours` to just `region_colours`.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -699,6 +701,7 @@ bool Dataloader::_load_map_dir(GameManager& game_manager) const {
 	F(continent) \
 	F(adjacencies) \
 	F(region) \
+	F(region_colours) \
```
Hop311, line 704: Because the region colours file isn't customised, having it here in the list of things which we expect in `default.map` causes an error message. This line just needs to be removed, the `region_colours` variable created above will provide the path.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -735,9 +738,17 @@ bool Dataloader::_load_map_dir(GameManager& game_manager) const {
...
-	if (!map.load_region_file(parse_defines(lookup_file(append_string_views(map_directory, region))).get_file_node())) {
-		Logger::error("Failed to load region file!");
-		ret = false;
+	{
+		std::vector<colour_t> colours;
+		if (!map.load_region_colours(parse_defines(lookup_file(append_string_views(interface_directory, region_colours))).get_file_node(), colours)) {
+			Logger::error("Failed to load region colours file!");
+			ret = false;
+		}
+
```
Hop311, line 747:
```suggestion
		using namespace OpenVic::colour_literals;
		colours.push_back(0xFFFFFF_rgb);
```
White needs to be added to the end of the list of colours. It can be created from the integer literal `0xFFFFFF` with the suffix `_colour`, athough I'd prefer if the suffix was `_rgb`, to match `_argb`. Can you edit [this line in Colour.hpp](https://github.com/OpenVicProject/OpenVic-Simulation/blob/master/src/openvic-simulation/types/Colour.hpp#L402) to make it `_rgb` instead of `_colour`.

src/openvic-simulation/map/Map.cpp
```diff
@@ -562,15 +562,28 @@ bool Map::load_province_positions(BuildingTypeManager const& building_type_manag
 	})(root);
 }
 
-bool Map::load_region_file(ast::NodeCPtr root) {
+bool Map::load_region_colours(ast::NodeCPtr root, std::vector<colour_t>& colours) {
+	return expect_dictionary(
+		[&colours](std::string_view key, ast::NodeCPtr value) -> bool {
+			if (key != "color") {
+				return false;
```
Hop311, line 569:
```suggestion
				Logger::error("Invalid key in region colours: \"", key, "\"");
				return false;
```
Can you add an error message here

src/openvic-simulation/map/Map.cpp
```diff
@@ -562,15 +562,28 @@ bool Map::load_province_positions(BuildingTypeManager const& building_type_manag
...
+			}
+			colour_t colour = colour_t::null();
+			const bool ret = expect_colour(assign_variable_callback(colour))(value);
+			colours.push_back(colour);
+			return ret;
```
Hop311, line 574:
```suggestion
			return expect_colour(vector_callback(colours))(value);
```
No need to use a temporary colour variable, `vector_callback` can do this for you automatically

Hop311 (commented):
Really great work, in particular keeping the colour vector in the load function accessed via (correctly const-qualified) references rather than as a member variable of `Map` that would stick around unnecessarily.

src/openvic-simulation/map/Map.hpp
```diff
@@ -139,7 +139,8 @@ namespace OpenVic {
 		bool load_province_definitions(std::vector<ovdl::csv::LineObject> const& lines);
 		/* Must be loaded after adjacencies so we know what provinces are coastal, and so can have a port */
 		bool load_province_positions(BuildingTypeManager const& building_type_manager, ast::NodeCPtr root);
-		bool load_region_file(ast::NodeCPtr root);
+		bool load_region_colours(ast::NodeCPtr root, std::vector<colour_t>& colours);
```
Hop311, line 142: Since this function doesn't use any of Map's member variables/functions, it could be made `static` and then called in `Dataloader.cpp` via `Map::load_region_colours`.

Jarturog: oops, I hadn't even thought about it, thanks for the feedback :)

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -666,6 +666,7 @@ bool Dataloader::_load_events(GameManager& game_manager) {
 
 bool Dataloader::_load_map_dir(GameManager& game_manager) const {
 	static constexpr std::string_view map_directory = "map/";
+	static constexpr std::string_view interface_directory = "interface/"; /* TODO - use the same variable from _load_interface_files */
```
Hop311, line 669: Since nothing else in this function will use the `interface` directory and we're not going to change the region colours filepath as we need to match what Vic2 uses, might as well remove this line and just make `region_colours` hold the full path.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -682,6 +683,7 @@ bool Dataloader::_load_map_dir(GameManager& game_manager) const {
 	static constexpr std::string_view default_region_sea = "region_sea.txt"; /* Some empty province sets (unused). */
 	static constexpr std::string_view default_province_flag_sprite = "province_flag_sprites"; /* Canal sprite/model names. */
 	static constexpr std::string_view climate_file = "climate.txt";
+	static constexpr std::string_view region_colours = "colors.txt"; /* File from interface_directory */
```
Hop311, line 686:
```suggestion

	static constexpr std::string_view region_colours = "interface/colors.txt"; /* This file is not in map_directory */
```
As explained above, might as well just put the whole path in this variable. Also adding a newline between this and the filenames above would help make clear that it's not in the `map` directory.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -735,9 +737,20 @@ bool Dataloader::_load_map_dir(GameManager& game_manager) const {
 		ret = false;
 	}
 
-	if (!map.load_region_file(parse_defines(lookup_file(append_string_views(map_directory, region))).get_file_node())) {
-		Logger::error("Failed to load region file!");
-		ret = false;
+	{
+		std::vector<colour_t> colours;
+		if (!Map::load_region_colours(parse_defines(lookup_file(append_string_views(interface_directory, region_colours))).get_file_node(), colours)) {
```
Hop311, line 742:
```suggestion
		if (!Map::load_region_colours(parse_defines(lookup_file(region_colours)).get_file_node(), colours)) {
```
Using a single variable for the region colours filepath saves us an unnecessary `append_string_views`, which currently allocates memory for an intermediate `std::string` holding the combined path.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -735,9 +737,20 @@ bool Dataloader::_load_map_dir(GameManager& game_manager) const {
...
+			Logger::error("Failed to load region colours file!");
+			ret = false;
+		}
+
+		using namespace OpenVic::colour_literals;
+		colours.push_back(0xFFFFFF_rgb);
```
Hop311, line 748:
```suggestion
		colours.push_back(0xFFFFFF_rgb); /* This ensures there is always at least one region colour. */
```
A comment here would be nice, makes the code seem less arbitrary.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -735,9 +737,20 @@ bool Dataloader::_load_map_dir(GameManager& game_manager) const {
...
+
+		if (!map.load_region_file(parse_defines(lookup_file(append_string_views(map_directory, region))).get_file_node(), colours)) {
```
Hop311, line 750:
```suggestion
		if (!map.load_region_file(
			parse_defines(lookup_file(append_string_views(map_directory, region))).get_file_node(), colours
		)) {
```
One part of our (not always perfectly enforced) style guide (which I still need to update 🤦‍♂️) is a maximum line length of 128 columns. VSCode has a "ruler" feature which helps show this visually, it can be enabled by adding `"editor.rulers": [ 128 ]` to your User Settings json file (opened by pressing `F1` then searching for `Preferences: Open User Settings (JSON)`).

src/openvic-simulation/map/Map.cpp
```diff
@@ -562,15 +562,26 @@ bool Map::load_province_positions(BuildingTypeManager const& building_type_manag
 	})(root);
 }
 
-bool Map::load_region_file(ast::NodeCPtr root) {
+bool Map::load_region_colours(ast::NodeCPtr root, std::vector<colour_t>& colours) {
+	return expect_dictionary(
```
Hop311, line 566:
```suggestion
	return expect_dictionary_reserve_length(
		colours,
```
Currently this code will need to reallocate `colours` every time it runs out of space, this can be avoided by reserving space in the `std::vector`. `expect_dictionary_reserve_length` will reserve as much space as there are entries in the dictionary, so in this case one per colour.

Hop311 (commented):
Left some more comments, just some simple bits of cleanup.
Once these are addressed, could you please squash this PR down to a single commit before we merge it. We do this to keep `master` clean, with every commit adding a functioning new feature, while still using a regular merge (rather than Github's squash merge) so that there's a link to the PR in the commit history.

Hop311 (approved):
Great work! 👍

PR #151, opened 2024-03-02
Cleaned up tech loading and added tech lists for areas

Hop311:


conboy137 (approved):
Perfect il put this to use

PR #153, opened 2024-03-22
GFX objectTypes loading framework + Actor loading

Hop311:
- Add `GFXObject` framework for loading GFX file `objectTypes`. Currently only loads `EMFXActorType` but all other structures are broken down into their compnents via `expect_dictionary_keys` chains with `success_callback` placeholders.
- Reanmed `GFX` files to `GFXSprite`, to differentiate from `GFXObject` files.
- Updated `fixed_point_t::parse` to allow `f` or `F` suffix.


Spartan322 (approved):
Runs fine on Linux.

PR #161, opened 2024-06-03
Province const/mutable separation + State cleanup

Hop311:


src/openvic-simulation/GameManager.cpp
```diff
@@ -105,21 +104,30 @@ static constexpr colour_argb_t DEFAULT_COLOUR_WHITE = (0xFFFFFF_argb).with_alpha
...
+static constexpr auto get_colour_mapmode(T const*(P::*get_item)() const) {
+	return [get_item](Map const& map, ProvinceInstance const& province) -> Mapmode::base_stripe_t {
+		ProvinceDefinition const& province_definition = province.get_province_definition();
+
+		T const* item;
+		if constexpr (std::same_as<P, ProvinceDefinition>) {
+			item = (province_definition.*get_item)();
+		} else {
+			item = (province.*get_item)();
+		}
```
Spartan322, line 118:
```suggestion
		T const& item = [&]{
		  if constexpr (std::same_as<P, ProvinceDefinition>) {
			  return *(province_definition.*get_item)();
		  } else {
			  return *(province.*get_item)();
		  }
		}();
```
Unless the item pointer needs to be changed, initializing them this way is generally safer.

Spartan322 (approved):
Looks good, tested on Linux

PR #162, opened 2024-06-06
Map instance

Hop311:


Juliusfrank11:
Tested on windows.

Juliusfrank11 (approved):
Tested on Windows. Ran successfully, I did get warning messages about redundant definition of provinces, but I'm assuming that's from jank Paradox left in the final game

PR #163, opened 2024-06-07
Add GoodInstance

Hop311:


schombert:
I understand the motivation for these sorts of splits: there is some data that you don't expect to change while the game is in progress, and this split is a way of enforcing that it does not. However, this sort of implementation is probably not what you want, for two reasons. First, it will be un-ergonomic: frequently you will be working with a reference to an instance and will need data from the definition, or vice versa, requiring duplication of the lookup code. If you pass ids around, you will be constantly looking up both from two different managers; if you pass around references you will double the number of references you need to pass around. Secondly, it will annoying to refactor. Eventually you will want to make a property mutable that you had originally put in the definition, and that will new require changes in many places.

If you really want to have the compiler enforce the distinction, I would have all the properties stored in a single object without direct public access to any of them. You then expose the properties via one of many views. Each view exposes the getters and setters appropriate for its context (for example, you can have a view that allows setting of the properties that can change while the game is running while only allowing read access to the definitional properties. Then, you can enforce whatever correctness checks you want at a single point of failure: where the code requests a particular view of the data. This also makes refactoring easy because changing where a property can be viewed or changed is simply a matter of adding/removing the right functions from particular views.

Juliusfrank11 (approved):
Tested on Windows

Hop311:
@schombert The main reason behind splitting const defines data and mutable game state data is to make it easier to reset the game state (by destroying and recreating the top level `InstanceManager` defined in the next PR), rather than having specific code for resetting every mutable variable. Instances should be the main variants passed around, with access to their corresponding definition through a reference variable. I'll admit this approach doesn't produce the prettiest code, but it does help make clear which data need to go in saves, contribute to the checksum, etc. compared to using a combined class.

PR #169, opened 2024-07-05
Update OpenVicProject/OpenVic-Dataloader@4a49007

Spartan322:
See OpenVicProject/OpenVic-Dataloader#50

Update OpenVicProject/OpenVic-Dataloader@4a49007
Update .clang-format categories
Move type_name and SelfType functioanlity to utility/Getters

Average usr time execution appears the same on Linux, but average sys time execution appears to have doubled. (from 100 milliseconds to 200 milliseconds)

### DIFF
--- a/src/openvic-simulation/dataloader/NodeTools.cpp
+++ b/src/openvic-simulation/dataloader/NodeTools.cpp
@@ -1,34 +1,70 @@
 #include "NodeTools.hpp"
 
+#include <type_traits>
+
+#include <openvic-dataloader/detail/Utility.hpp>
+#include <openvic-dataloader/v2script/AbstractSyntaxTree.hpp>
+
+#include <dryad/node.hpp>
+
+#include <range/v3/iterator/operations.hpp>
+#include <range/v3/view/enumerate.hpp>
+
 #include "openvic-simulation/types/Colour.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
 
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
 template<typename T>
-static NodeCallback auto _expect_type(Callback<T const&> auto callback) {
+static NodeCallback auto _expect_type(Callback<T const*> auto callback) {
 	return [callback](ast::NodeCPtr node) -> bool {
 		if (node != nullptr) {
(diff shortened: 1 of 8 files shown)


src/openvic-simulation/dataloader/NodeTools.cpp
```diff
@@ -286,16 +333,19 @@ node_callback_t NodeTools::expect_length(callback_t<size_t> callback) {
...
-			for (ast::NodeUPtr const& sub_node : list_node._statements) {
-				ast::AssignNode const* assign_node = sub_node->cast_to<ast::AssignNode>();
-				if (assign_node != nullptr && assign_node->_name == key) {
+			for (auto sub_node : list) {
+				auto const* assign_node = dryad::node_try_cast<ast::AssignStatement>(sub_node);
+				std::string_view left;
+				if (!expect_identifier(assign_variable_callback_string(left))(assign_node->left())) {
+					continue;
+				}
+				if (assign_node != nullptr && left == key) {
```
Hop311, line 346: `assign_node->left()` is called before the `assign_node != nullptr` check, which could cause a segfault if `sub_node` is not successfully cast to an `ast::AssignStatement`.

src/openvic-simulation/dataloader/NodeTools.cpp
```diff
@@ -356,3 +406,7 @@ std::ostream& OpenVic::operator<<(std::ostream& stream, name_list_t const& name_
 callback_t<std::string_view> NodeTools::assign_variable_callback_string(std::string& var) {
 	return assign_variable_callback_cast<std::string_view>(var);
 }
+
+callback_t<std::string_view> NodeTools::assign_variable_callback_string(std::string_view& var) {
+	return assign_variable_callback_cast<std::string_view>(var);
+}
```
Hop311, line 412: `assign_variable_callback` works for this case, the other `assign_variable_callback_string` function was only needed due to the mismatch in parameters (`std::string_view` vs `std::string`).

Hop311 (approved):
Great work, only comments are a potential segfault if `_NodeStatementRange` can contain non-`AssignStatement`s and that `assign_variable_callback` works when the source and dest are both `std::string_view`s.

Hop311 (approved):

PR #172, opened 2024-07-06
Use rvalue references for NodeTools

Spartan322:
Saves between 500-550ms on my local Linux debug tests.


Spartan322:
Its almost certainly more efficient to replace std::function a move only function or function ref, like in https://github.com/Naios/function2 or https://github.com/iboB/itlib/blob/master/include/itlib/ufunction.hpp or any number of implementations from projects referenced in https://github.com/jamboree/CxxFunctionBenchmark

Hop311 (approved):
Tested on Windows 👍

Hop311 (approved):
LGTM 👍

PR #174, opened 2024-07-16
Add IndexedMap and use in low key count, high value density cases

Hop311:
There are probably more places we can switch to using IndexedMap, e.g. for a Pop's Reform/Issue support distribution (although their current setup using separate registries/indexing means they can't be stored in a single IndexedMap), but I'll leave those for future PRs.


Spartan322 (approved):
Works on Linux

PR #177, opened 2024-07-21
Fleshing out Country, State and Province instances + history

Hop311:
Some parts are still WIP/subject to change, particularly country instance variables and gamestate updating.

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.cpp
+++ b/src/openvic-simulation/country/CountryInstance.cpp
@@ -2,58 +2,155 @@
 
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/history/CountryHistory.hpp"
-#include "openvic-simulation/military/UnitInstanceGroup.hpp"
+#include "openvic-simulation/map/MapInstance.hpp"
+#include "openvic-simulation/politics/Ideology.hpp"
+#include "openvic-simulation/research/Invention.hpp"
+#include "openvic-simulation/research/Technology.hpp"
 
 using namespace OpenVic;
+
+static constexpr colour_t ERROR_COLOUR = colour_t::from_integer(0xFF0000);
 
 CountryInstance::CountryInstance(
 	CountryDefinition const* new_country_definition,
 	decltype(technologies)::keys_t const& technology_keys,
 	decltype(inventions)::keys_t const& invention_keys,
-	decltype(upper_house)::keys_t const& ideology_keys
-) : country_definition { new_country_definition },
-	primary_culture { nullptr },
-	religion { nullptr },
-	ruling_party { nullptr },
-	last_election {},
+	decltype(upper_house)::keys_t const& ideology_keys,
+	decltype(pop_type_distribution)::keys_t const& pop_type_keys
+) : /* Main attributes */
+	country_definition { new_country_definition },
+	colour { ERROR_COLOUR },
 	capital { nullptr },
-	government_type { nullptr },
-	plurality { 0 },
-	national_value { nullptr },
+	country_flags {},
 	civilised { false },
-	prestige { 0 },
-	upper_house { &ideology_keys },
+	releasable_vassal { true },
+	owned_provinces {},
+	controlled_provinces {},
+	core_provinces {},
+	states {},
+
+	/* Production */
+
+	/* Budget */
+	cash_stockpile { 0 },
+
+	/* Technology */
 	technologies { &technology_keys },
-	inventions { &invention_keys } {}
+	inventions { &invention_keys },
+	current_research { nullptr },
+	invested_research_points { 0 },
+	expected_completion_date {},
+	research_point_stockpile { 0 },
+	daily_research_points { 0 },
+	national_literacy { 0 },
+	tech_school { nullptr },
+
+	/* Politics */
+	national_value { nullptr },
+	government_type { nullptr },
+	last_election {},
+	ruling_party { nullptr },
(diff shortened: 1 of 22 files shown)


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -2,59 +2,158 @@
...
-			"Attempted to remove accepted culture ", culture_to_remove->get_identifier(), " from country ",
-			country_definition->get_identifier(), ": not present!"
+bool CountryInstance::clear_country_flag(std::string_view flag) {
+	if (flag.empty()) {
+		Logger::error("Attempted to clear empty country flag from country ", get_identifier());
+		return false;
+	}
+	if (country_flags.erase(flag) == 0) {
+		Logger::error(
+			"Attempted to clear country flag \"", flag, "\" from country ", get_identifier(), ": not set!"
```
wvpm, line 120: Clearing a flag that isn't in the set is not an error.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -176,20 +277,155 @@ bool CountryInstance::apply_history_to_country(CountryHistoryEntry const* entry)
...
+
+void CountryInstance::_update_population() {
+	total_population = 0;
+	national_literacy = 0;
+	national_consciousness = 0;
+	national_militancy = 0;
+	pop_type_distribution.clear();
+
+	for (auto const& state : states) {
+		total_population += state->get_total_population();
```
wvpm, line 337: I'd store `state->get_total_population()` in a variable `state_population` and reuse that instead of calling it 4 times.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -2,59 +2,158 @@
...
+	max_ship_supply { 0 },
+	leadership_points { 0 },
+	war_exhaustion { 0 } {}
 
 std::string_view CountryInstance::get_identifier() const {
-	return country_definition != nullptr ? country_definition->get_identifier() : "NULL";
+	return country_definition->get_identifier();
 }
 
-bool CountryInstance::add_accepted_culture(Culture const* new_accepted_culture) {
```
wvpm, line 33: Where did this method go?
I can't find it in the .cpp anymore.

src/openvic-simulation/map/State.cpp
```diff
@@ -1,23 +1,45 @@
...
-	std::vector<ProvinceInstance*>&& provinces, ProvinceInstance::colony_status_t colony_status
+	StateSet const& new_state_set, CountryInstance* owner, ProvinceInstance* capital,
+	std::vector<ProvinceInstance*>&& provinces, ProvinceInstance::colony_status_t colony_status,
+	decltype(pop_type_distribution)::keys_t const& pop_type_keys
 ) : state_set { new_state_set }, owner { owner }, capital { capital }, provinces { std::move(provinces) },
-	colony_status { colony_status } {}
+	colony_status { colony_status }, pop_type_distribution { &pop_type_keys } {}
+
+std::string State::get_identifier() const {
+	return StringUtils::append_string_views(state_set.get_region().get_identifier(), "_", owner->get_identifier());
```
wvpm, line 20: Note you can have a colonial state and non-colonial state with the same owner in the same area.
So maybe append `_c` or `_colony` if it is colonial.
When the colonial part becomes a state, they merge.

wvpm (approved):
Looks good in general with some minor comments

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -2,59 +2,158 @@
...
-			"Attempted to add accepted culture ", new_accepted_culture->get_identifier(), " to country ",
-			country_definition->get_identifier(), ": already present!"
+bool CountryInstance::set_country_flag(std::string_view flag) {
+	if (flag.empty()) {
+		Logger::error("Attempted to set empty country flag for country ", get_identifier());
+		return false;
+	}
+	if (!country_flags.emplace(flag).second) {
+		Logger::error(
+			"Attempted to set country flag \"", flag, "\" for country ", get_identifier(), ": already set!"
```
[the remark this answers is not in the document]
Hop311: This is for country flags set by history, event, decision, etc, currently only used for history. I'll add an extra `bool warn` argument so duplicate flags in country history produce a warning, as that can only happen from flags being unnecessarily repeated in the same file, but when set from other sources like events and decisions there won't be a warning.

src/openvic-simulation/country/CountryInstance.cpp
Hop311: Same as above

src/openvic-simulation/country/CountryInstance.cpp
Hop311: Sure, I'll do that both here and in `State.cpp` and I'll also change it from an `int32_t` conversion to a `fixed_point_t` conversion. Eventually I'm expecting we'll change `Pop::pop_size_t` to a fixed point type rather than `int64_t`, as I believe it'll be needed for keeping track of population growth (otherwise small pops would never grow).

src/openvic-simulation/country/CountryInstance.cpp
Hop311: It's generated by the `ADD_AND_REMOVE` macro in `CountryInstance.cpp` which makes the add and remove functions for owned provinces, controlled provinces, core provinces, states and accepted cultures.

wvpm (approved):

PR #181, opened 2024-08-01
Music and Sound Effect define loading

Nemrav:
This is the sim portion of the music and Sound defines pr, needs review (code style?).
Main changes are:
- Addition of SongChance and SoundEffect loaders
- Dataloader now has separated root setting, and define loading functions so that the title_theme soundtrack can be loaded before the defines. Other files like GameSingleton and main.cpp changed to match.

### DIFF
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -824,6 +824,41 @@ bool Dataloader::_load_map_dir(DefinitionManager& definition_manager) const {
 	return ret;
 }
 
+bool Dataloader::_load_song_chances(DefinitionManager& definition_manager) {
+	static constexpr std::string_view song_chance_file = "music/songs.txt";
+	const fs::path path = lookup_file(song_chance_file, false);
+
+	bool ret = true;
+	SongChanceManager& song_chance_manager = definition_manager.get_song_chance_manager();
+
+	if(path.empty()) {
+		Logger::info("No Songs.txt file to load");
+		song_chance_manager.lock_song_chances();
+		return true;
+	}
+
+	ret &= song_chance_manager.load_songs_file(parse_defines_cached(path).get_file_node());
+
+	song_chance_manager.lock_song_chances();
+
+	return ret;
+}
+
+bool Dataloader::_load_sound_effect_defines(DefinitionManager& definition_manager) const {
+	static constexpr std::string_view sfx_file = "interface/sound.sfx";
+	const fs::path path = lookup_file(sfx_file);
+
+	bool ret = true;
+	SoundEffectManager& sound_effect_manager = definition_manager.get_sound_effect_manager();
+
+	ret &= sound_effect_manager.load_sound_defines_file(parse_defines(path).get_file_node());
+
+	sound_effect_manager.lock_sound_effects();
+
+	return ret;
+
+}
+
 bool Dataloader::load_defines(DefinitionManager& definition_manager) {
 	if (roots.empty()) {
 		Logger::error("Cannot load defines - Dataloader has no roots!");
@@ -1043,6 +1078,14 @@ bool Dataloader::load_defines(DefinitionManager& definition_manager) {
 		Logger::error("Failed to load events!");
 		ret = false;
 	}
+	if (!_load_song_chances(definition_manager)) {
+		Logger::error("Error while loading Song chances!");
+		ret = false;
+	}
+	if (!_load_sound_effect_defines(definition_manager)) {
+		Logger::error("Failed to load sound effect defines");
+		ret = false;
+	}
 	if (!definition_manager.get_event_manager().load_on_action_file(
 		parse_defines(lookup_file(on_actions_file)).get_file_node()
 	)) {
@@ -1087,6 +1130,7 @@ bool Dataloader::parse_scripts(DefinitionManager& definition_manager) const {
 	PARSE_SCRIPTS("wargoal type", definition_manager.get_military_manager().get_wargoal_type_manager());
 	PARSE_SCRIPTS("decision", definition_manager.get_decision_manager());
 	PARSE_SCRIPTS("event", definition_manager.get_event_manager());
+	PARSE_SCRIPTS("song chance", definition_manager.get_song_chance_manager());
 	return ret;
 }
 
--- a/src/openvic-simulation/dataloader/Dataloader.hpp
+++ b/src/openvic-simulation/dataloader/Dataloader.hpp
@@ -21,7 +21,7 @@ namespace OpenVic {
 		using path_vector_t = std::vector<fs::path>;
 
 	private:
-		path_vector_t roots;
+		path_vector_t PROPERTY(roots);
 		std::vector<ovdl::v2script::Parser> cached_parsers;
 
 		bool _load_interface_files(UIManager& ui_manager) const;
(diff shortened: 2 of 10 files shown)


src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -824,6 +828,46 @@ bool Dataloader::_load_map_dir(DefinitionManager& definition_manager) const {
...
+	bool ret = true;
+	SongChanceManager& song_chance_manager = definition_manager.get_song_chance_manager();
+
+	if(song_chance_file.empty()){
+		Logger::info("No Songs.txt file to load");
+		song_chance_manager.lock_songChances();
+		return true;
+	}
+
+	ret &= song_chance_manager.load_songs_file(parse_defines_cached(song_chance_file).get_file_node());
```
Hop311, line 841:
```suggestion
bool Dataloader::_load_song_chances(DefinitionManager& definition_manager) {
	static constexpr std::string_view song_chance_file = "music/songs.txt";

	bool ret = true;
	SongChanceManager& song_chance_manager = definition_manager.get_song_chance_manager();

	const fs::path song_chance_path = lookup_file(song_chance_file, false);

	if (song_chance_path.empty()) {
		Logger::info("No ", song_chance_file, " file to load");
		song_chance_manager.lock_song_chances();
		return true;
	}

	ret &= song_chance_manager.load_songs_file(parse_defines_cached(song_chance_path).get_file_node());
```
Since we'll always be loading the same file we can just have it as a constant at the top of this function instead of an argument.
Since we want to check if the file exists before attempting to load it, we can create a local `fs::path` variable to store it in. Calling `lookup_file` with `false` as the second argument prevents it from reporting an error when the file isn't found, needed for this case where the file might not exist without causing any problems.
We still use `parse_defines_cached` as we'll need to hold onto the parsers so we can parse the chance scripts after everything else is loaded.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -824,6 +828,46 @@ bool Dataloader::_load_map_dir(DefinitionManager& definition_manager) const {
...
+	)) {
+		Logger::error("Error while loading sound defines file!");
+		ret = false;
+	}
+
+	sfx_manager.lock_sfx_defines();
+
+	return ret;
+
+}
```
Hop311, line 869:
```suggestion
bool Dataloader::_load_sfx_defines(DefinitionManager& definition_manager) const {
	static constexpr std::string_view sfx_file = "interface/sound.sfx";

	bool ret = true;
	SfxManager& sfx_manager = definition_manager.get_sfx_manager();

	ret &= sfx_manager.load_sound_defines_file(parse_defines(lookup_file(sfx_file)).get_file_node());

	sfx_manager.lock_sfx_defines();

	return ret;
}
```
This function can be const as we don't need to cache the parser (it can safely be freed once the load function ends). We can also provide the filepath as a constant, rather than an argument, as it'll always be the same.
There's no need for a check on the looked up filepath being empty, as `lookup_file` will print an error if it isn't found and the expect functions in `load_sound_defines_file` will handle the resulting null filenode properly.
We also don't need to print an error message here if `load_sound_defines_file` fails, as we print an identical message straight afterwards in `load_defines` if this function returns false.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -851,6 +895,8 @@ bool Dataloader::load_defines(DefinitionManager& definition_manager) {
 	static constexpr std::string_view static_modifiers_file = "common/static_modifiers.txt";
 	static constexpr std::string_view triggered_modifiers_file = "common/triggered_modifiers.txt";
 	static constexpr std::string_view on_actions_file = "common/on_actions.txt";
+	static constexpr std::string_view song_chance_file = "music/songs.txt";
+	static constexpr std::string_view sfx_file = "interface/sound.sfx";
```
Hop311, line 899: These can go inside the functions they're passed to.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -1043,6 +1089,14 @@ bool Dataloader::load_defines(DefinitionManager& definition_manager) {
...
 		ret = false;
 	}
+	if (!_load_song_chances(definition_manager, lookup_file(song_chance_file))){
+		Logger::error("Error while loading Song chances!");
+		ret = false;
+	}
+	if (!_load_sfx_defines(definition_manager, lookup_file(sfx_file))){
+		Logger::error("Failed to load sound effect defines");
+		ret = false;
+	}
```
Hop311, line 1099:
```suggestion
	if (!_load_song_chances(definition_manager)) {
		Logger::error("Error while loading Song chances!");
		ret = false;
	}
	if (!_load_sfx_defines(definition_manager)) {
		Logger::error("Failed to load sound effect defines");
		ret = false;
	}
```
These don't need the path arguments, and there should be spaces between the `)`s and `{`s.

Also in order to parse the song chance scripts, this line needs to be added to `Dataloader::parse_scripts` later in this file (with names updated to match any changes made to the struct/header and source names).
```
	PARSE_SCRIPTS("song chance", definition_manager.get_song_chance_manager());
```

Hop311 (commented):
Great work 👍 Mostly style comments but a few important bits including the code for parsing the loaded song chances

src/openvic-simulation/dataloader/Dataloader.hpp
```diff
@@ -86,6 +88,7 @@ namespace OpenVic {
 
 		/* In reverse-load order, so base defines first and final loaded mod last */
 		bool set_roots(path_vector_t const& new_roots);
+		path_vector_t get_roots() const;
```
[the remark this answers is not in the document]
Nemrav: Changed

src/openvic-simulation/misc/SoundEffect.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <openvic-simulation/types/fixed_point/FixedPoint.hpp>
+#include <openvic-simulation/types/IdentifierRegistry.hpp>
+
+namespace OpenVic {
+    /*For interface/Sound.sfx */
```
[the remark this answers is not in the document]
Nemrav: Resolved

src/openvic-simulation/misc/SoundEffect.hpp
```diff
@@ -0,0 +1,28 @@
...
+	struct SfxManager;
+	struct Sfx : HasIdentifier {
+		private:
+		friend struct SfxManager;
+		std::string PROPERTY(file);
+		fixed_point_t PROPERTY(volume);
+
+		public:
+		Sfx(std::string_view new_identifier, std::string_view new_file, fixed_point_t new_volume);
```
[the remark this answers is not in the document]
Nemrav: The style guide needs to be updated, because it explicitly says to put the public/private declarations on the same indentation level.

src/openvic-simulation/misc/MusicChance.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <openvic-simulation/types/IdentifierRegistry.hpp>
+#include <openvic-simulation/scripts/ConditionalWeight.hpp>
+
+namespace OpenVic {
+    /*For music/Songs.txt if it exists*/
+    struct SongChanceManager;
```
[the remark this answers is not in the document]
Nemrav: Changed to SongChance.*pp

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -824,6 +824,41 @@ bool Dataloader::_load_map_dir(DefinitionManager& definition_manager) const {
...
 }
 
+bool Dataloader::_load_song_chances(DefinitionManager& definition_manager) {
+	static constexpr std::string_view song_chance_file = "music/songs.txt";
+	const fs::path path = lookup_file(song_chance_file, false);
+
+	bool ret = true;
+	SongChanceManager& song_chance_manager = definition_manager.get_song_chance_manager();
+
+	if(song_chance_file.empty()) {
```
Hop311, line 834:
```suggestion
	if (path.empty()) {
```
This needs to check the lookup up path, not the constant path which will never be empty.

Hop311 (commented):
Just one small fix needed, then this can be squashed together and merged 👍

Hop311 (approved):

PR #184, opened 2024-08-04
Load font colour codes

Hop311:


Nemrav (approved):
Everything looks good.

PR #187, opened 2024-08-29
Remove unnecessary asserts + make `vec2_t(T val)` constructor explicit

Hop311:


src/openvic-simulation/types/Vector.hpp
```diff
@@ -14,7 +14,7 @@ namespace OpenVic {
 		T x, y;
 
 		constexpr vec2_t() = default;
-		constexpr vec2_t(T new_val) : x { new_val }, y { new_val } {}
+		explicit constexpr vec2_t(T new_val) : x { new_val }, y { new_val } {}
```
wvpm, line 17: Should this require existing calls to be changed to explicitly cast?

wvpm (approved):

src/openvic-simulation/types/Vector.hpp
Hop311: All existing calls are explicit, this just prevents accidental implicit casts eg `func(1)` instead of `func(ivcec2_t { 1 })`

PR #194, opened 2024-09-06
Add define caching to avoid post-load lookups

Hop311:


src/openvic-simulation/misc/Define.cpp
```diff
@@ -1,109 +1,249 @@
...
+	switch (type) {
+	case Date: return "date";
+	case Country: return "country";
+	case Economy: return "economy";
+	case Military: return "military";
+	case Diplomacy: return "diplomacy";
+	case Pops: return "pops";
+	case Ai: return "ai";
+	case Graphics: return "graphics";
+	case Unknown: [[fallthrough]];
```
wvpm, line 26: How does this work?

Hop311: When a switch case isn't ended by a `return` or `break` the program continues through to the next case. While this is well-defined and often useful behaviour, compilers may still warn about it in case it's accidental and the author forgot to add a `break`. The [fallthrough attribute](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) tells the compiler that the behaviour is intentional and suppresses any warnings.

In this case I could've just left out `case Unknown` and let it be handled by `default` directly, but IMO when using enums in switches it's better to be explicit and list every option, both to avoid ambiguity for readers and because compilers can warn about enum switch statements not handling every value (although that might only happen without a `default` statement in place).

src/openvic-simulation/misc/Define.cpp
```diff
@@ -1,109 +1,249 @@
...
-	return end_date ? *end_date : Date {};
+std::ostream& OpenVic::operator<<(std::ostream& os, Define::Type type) {
+	return os << Define::type_to_string(type);
 }
 
-bool DefineManager::in_game_period(Date date) const {
-	if (start_date && end_date) {
-		return date.in_range(*start_date, *end_date);
+template<typename T>
+bool DefineManager::load_define(T& value, Define::Type type, std::string_view name) const {
```
schombert, line 77: wouldn't optional<T> be more ergonomic here? Using an output parameter that is *maybe* set is error prone because it will be easy to forget to check the return value (unless you add `[[nodiscard]]`) and try to use a potentially uninitialized value.

src/openvic-simulation/misc/Define.cpp
```diff
@@ -1,109 +1,249 @@
...
+
 			}
 		})
 	)(root);
 
 	lock_defines();
 
+	// Date
+	ret &= load_define(start_date, Date, "start_date");
+	ret &= load_define(end_date, Date, "end_date");
```
schombert, line 232: Obviously this is both sensible and necessary: if you are going to be using the define during game play, you don't want to be doing a string lookup. But, doesn't this apply to *all* of the defines? And if it applies to all of them, such that you will eventually have one variable per define value, then you don't need the string map -> defines at all. You can just load them into their variables and omit the complexity of the map completely.

src/openvic-simulation/misc/Define.cpp
wvpm: Funny, I thought it might indicate fallthrough to another case but then I thought "Nah, if he wanted that he would've just omitted `case Unknown:` or the `[[fallthrough]]`".

Omitted `case Unknown:` seems best here. The fallthrough attribute is best suited for implicit else statements and such if you ask me.

src/openvic-simulation/misc/Define.cpp
Hop311: The return value is only used to contribute to `load_defines_file`'s return value, the define variables are default initialised in `DefineManager`'s constructor so no further fallback is required if `load_define` fails.

src/openvic-simulation/misc/Define.cpp
Hop311: Yeah I'm considering doing away with `DefineManager` altogether and directly loading the values into member variables of other appropriate managers, both to remove the unnecessary intermediate storage and to avoid needing to ensure simulation functions have access to yet another manager. I wasn't the original author or reviewer of this approach, and although I'm gonna stick with adaptating it and getting the work I've built on top of it merged for now, after that I'll look into a better repo-wide approach to constants and include the `defines.lua` ones in that.

wvpm (approved):

PR #195, opened 2024-09-08
Ranking System

Hop311:


src/openvic-simulation/map/State.cpp
```diff
@@ -40,13 +50,25 @@ void State::update_gamestate() {
...
 
 	if (total_population > 0) {
 		average_literacy /= total_population;
 		average_consciousness /= total_population;
 		average_militancy /= total_population;
 	}
+
+	// TODO - use actual values when State has factory data
+	const int32_t total_factory_levels_in_state = 0;
+	const int32_t potential_workforce_in_state = 0; // sum of worker pops, regardless of employment
```
wvpm, line 65: It's good to see mechanics implemented in code because it raises more questions.
In this case: "Which pops to sum here?"
Turns out the answer is pops with `can_work_factory = yes`
I assumed it would instead use the workforce requested by the actual factories but I was wrong.

I'll update the documentation.

src/openvic-simulation/military/UnitInstanceGroup.cpp
```diff
@@ -173,7 +173,7 @@ bool UnitInstanceManager::generate_unit_instance(
 		[&unit_deployment]() -> UnitInstanceBranched<Branch> {
 			if constexpr (Branch == UnitType::branch_t::LAND) {
 				// TODO - get pop from Province unit_deployment.get_home()
-				return { unit_deployment.get_name(), unit_deployment.get_type(), nullptr };
+				return { unit_deployment.get_name(), unit_deployment.get_type(), nullptr, false };
```
wvpm, line 176: Could you make this `false` a constant named "is_not_mobilised" or something similar?

src/openvic-simulation/misc/Define.cpp
```diff
@@ -150,15 +150,25 @@ bool DefineManager::load_define_years(Timespan& value, Define::Type type, std::s
 DefineManager::DefineManager()
   : // Date
 	start_date { 1836, 1, 1 },
-	end_date { 1936, 1, 1 }
+	end_date { 1936, 1, 1 },
 
 	// Country
+	great_power_rank { 8 },
```
wvpm, line 156: Why init the defines when you must overwrite this value with the value read from file?
If the value is missing from the files, we shouldn't fall back to some hardcoded value but instead error.

wvpm (approved):
Approved with comments

src/openvic-simulation/military/UnitInstanceGroup.cpp
Hop311: I'll add a comment next to the argument, and move the pop comment next to `nullptr` similarly.

src/openvic-simulation/misc/Define.cpp
Hop311: It's best practice to initialise all variables so nothing contains random data, and the current system will log an error if the define is missing but making the game not start due to load errors is a larger change beyond this PR.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -147,7 +153,18 @@ void ProvinceInstance::_update_pops() {
 	culture_distribution.clear();
 	religion_distribution.clear();
 
-	for (Pop const& pop : pops) {
+	max_supported_regiments = 0;
+
+	using enum colony_status_t;
+
+	const fixed_point_t min_pop_size_for_regiment_multiplier =
```
wvpm, line 160: It's not min pop size for regiment but rather pops per regiment.

src/openvic-simulation/misc/Define.cpp
```diff
@@ -150,15 +150,25 @@ bool DefineManager::load_define_years(Timespan& value, Define::Type type, std::s
...
+	lose_great_power_grace_days { Timespan::from_years(1) },
+	secondary_power_rank { 16 },
+	country_investment_industrial_score_factor { 1 },
 
 	// Economy
 
 	// Military
+	pop_size_per_regiment { 1000 },
```
wvpm, line 164: 3000 in vanilla. Not sure if you want to set everything to vanilla since it will be overwritten anyway...

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -125,6 +127,18 @@ void Pop::set_location(ProvinceInstance const& new_location) {
 	}
 }
 
+void Pop::update_gamestate(DefineManager const& define_manager, fixed_point_t const& min_pop_size_for_regiment_multiplier) {
+	if (type.get_can_be_recruited()) {
```
wvpm, line 131: There should be a check if the culture is recruitable. Unit types can be restricted to primary or accepted cultures.
If pop is not of an accepted culture and all unit types require that, then the pop can't be recruited

wvpm (dismissed):
Approve with comments

src/openvic-simulation/misc/Define.cpp
Hop311: I set arbitrary numbers just to have something there by default, the player can load vanilla numbers from their own copy of the base game if they wish, we don't provide that data ourselves.

src/openvic-simulation/pop/Pop.cpp
Hop311: Good catch, I've added a system for keeping track of which cultures are recruitable based on currently unlocked regiment types.

src/openvic-simulation/military/UnitType.hpp
```diff
@@ -94,7 +94,14 @@ namespace OpenVic {
 	struct UnitTypeBranched<UnitType::branch_t::LAND> : UnitType {
 		friend struct UnitTypeManager;
 
-		enum struct allowed_cultures_t { ALL_CULTURES, ACCEPTED_CULTURES, PRIMARY_CULTURE };
+		// Each value is a subset of its predecessor, so smaller values contain larger values
+		enum struct allowed_cultures_t { ALL_CULTURES, ACCEPTED_CULTURES, PRIMARY_CULTURE, NO_CULTURES };
```
wvpm, line 98: Why add NO_CULTURES?
That value is invalid and never occurs in Victoria 2 (including mods).

wvpm: Nevermind, figured it out. You use this purely as a default in the country instance so you can expand it when unlocking more types.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -245,12 +300,70 @@ template void CountryInstance::add_leader(LeaderBranched<UnitType::branch_t::NAV
...
+
+bool CountryInstance::is_primary_culture(Culture const& culture) const {
+	return &culture == primary_culture;
+}
+
+bool CountryInstance::is_accepted_culture(Culture const& culture) const {
+	return accepted_cultures.contains(&culture);
+}
+
+bool CountryInstance::can_recruit_culture(Culture const& culture) const {
```
wvpm, line 337: This feels misplaced as recruitment requires the same logic.
Shouldn't this be part of the allowed_cultures_t?
allowed_cultures_t seems like a stategy to me.

Something like allowed_cultures_t.can_recruit(country instance, pop)

wvpm (dismissed):
With comment

src/openvic-simulation/country/CountryInstance.cpp
Hop311: Good point, I'll move it alongside the `get_more_permissive_allowed_cultures` function

wvpm (approved):

PR #196, opened 2024-09-08
Simulation Side Rivers

BrickPi:
Loads the rivers bitmap into structure that the sim can eventually use for movement and combat checks. No graphics in this PR, probably going to shelve working on river graphics and focus on sim-side as I get more availability to contribute again.

Tested, loads 905 rivers successfully in vanilla.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@
 
 # VSCode
 .vscode/*
-!.vscode/launch.json
 !.vscode/tasks.json
 
 # Visual Studio
--- a/src/openvic-simulation/map/MapDefinition.cpp
+++ b/src/openvic-simulation/map/MapDefinition.cpp
@@ -1,9 +1,11 @@
 #include "MapDefinition.hpp"
 
+#include <cstdint>
 #include <vector>
 
 #include "openvic-simulation/types/Colour.hpp"
 #include "openvic-simulation/types/OrderedContainers.hpp"
+#include "openvic-simulation/types/Vector.hpp"
 #include "openvic-simulation/utility/BMP.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
 
@@ -11,6 +13,8 @@ using namespace OpenVic;
 using namespace OpenVic::NodeTools;
 
 MapDefinition::MapDefinition() : dims { 0, 0 }, max_provinces { ProvinceDefinition::MAX_INDEX } {}
+
+RiverSegment::RiverSegment(uint8_t new_size, std::vector<ivec2_t> new_points) : size { new_size }, points { std::move(new_points) } {}
 
 bool MapDefinition::add_province_definition(std::string_view identifier, colour_t colour) {
 	if (province_definitions.size() >= max_provinces) {
@@ -490,7 +494,7 @@ static constexpr colour_t colour_at(uint8_t const* colour_data, int32_t idx) {
 	return { colour_data[idx + 2], colour_data[idx + 1], colour_data[idx] };
 }
 
-bool MapDefinition::load_map_images(fs::path const& province_path, fs::path const& terrain_path, bool detailed_errors) {
+bool MapDefinition::load_map_images(fs::path const& province_path, fs::path const& terrain_path, fs::path const& rivers_path, bool detailed_errors) {
 	if (!province_definitions_are_locked()) {
 		Logger::error("Province index image cannot be generated until after provinces are locked!");
 		return false;
@@ -500,12 +504,14 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
 		return false;
 	}
 
+	static constexpr uint16_t expected_province_bpp = 24;
+	static constexpr uint16_t expected_terrain_rivers_bpp = 8;
+
 	BMP province_bmp;
 	if (!(province_bmp.open(province_path) && province_bmp.read_header() && province_bmp.read_pixel_data())) {
 		Logger::error("Failed to read BMP for compatibility mode province image: ", province_path);
 		return false;
 	}
-	static constexpr uint16_t expected_province_bpp = 24;
 	if (province_bmp.get_bits_per_pixel() != expected_province_bpp) {
 		Logger::error(
 			"Invalid province BMP bits per pixel: ", province_bmp.get_bits_per_pixel(), " (expected ", expected_province_bpp,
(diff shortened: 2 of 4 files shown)


src/openvic-simulation/map/MapDefinition.hpp
```diff
@@ -20,6 +21,16 @@ namespace OpenVic {
...
 	struct ModifierManager;
 
+	struct RiverSegment {
+		friend struct MapDefinition;
+
+	private:
+			const uint8_t PROPERTY(size);
+			const std::vector<ivec2_t> PROPERTY(points);
+
+			RiverSegment(uint8_t new_size, std::vector<ivec2_t> new_points);
```
Hop311, line 31:
```suggestion
		const uint8_t PROPERTY(size);
		std::vector<ivec2_t> PROPERTY(points);

		RiverSegment(uint8_t new_size, std::vector<ivec2_t>&& new_points);

	public:
		RiverSegment(RiverSegment&&) = default;
```
- Make `points` non-const so it can be moved (const vectors can only be copied).
- For `new_points` to be a r-value reference/moved.
- Add a public default move constructor
- Indent back a tab to be one past private.

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -12,6 +14,8 @@ using namespace OpenVic::NodeTools;
 
 MapDefinition::MapDefinition() : dims { 0, 0 }, max_provinces { ProvinceDefinition::MAX_INDEX } {}
 
+RiverSegment::RiverSegment(uint8_t new_size, std::vector<ivec2_t> new_points) : size { new_size }, points { std::move(new_points) } {}
```
Hop311, line 17:
```suggestion
RiverSegment::RiverSegment(uint8_t new_size, std::vector<ivec2_t>&& new_points)
	: size { new_size }, points { std::move(new_points) } {}
```
Make `new_points` an r-value reference and move the initialiser list to the next line

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -519,18 +525,33 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
-	if (terrain_bmp.get_bits_per_pixel() != expected_terrain_bpp) {
+	if (terrain_bmp.get_bits_per_pixel() != expected_terrain_rivers_bpp) {
+		Logger::error(
+			"Invalid terrain BMP bits per pixel: ", terrain_bmp.get_bits_per_pixel(), " (expected ", expected_terrain_rivers_bpp, ")"
+		);
+		return false;
+	}
+
+	BMP rivers_bmp;
+	if (!(rivers_bmp.open(rivers_path) && rivers_bmp.read_header() && rivers_bmp.read_palette() && rivers_bmp.read_pixel_data())) {
```
Hop311, line 536: Is `read_palette` necessary here? We're only using the indices in the pixel data from what I can see

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -637,6 +658,183 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
+		}
+		// check pixel to left
+		if (last_segment_end.x > 0 && last_segment_direction != 3) {
+			if (river_data[idx - 1] < 12) {
+				points.push_back({ last_segment_end.x - 1, last_segment_end.y });
+				direction = 4;
+			}
+		}
+
+		uint8_t size = river_data[points.front().x + points.front().y * rivers_bmp.get_width()] - 1; // size of river from 1 - 10 determined by colour
```
Hop311, line 727: This line could crash if `points` is empty (e.g. a standalone starting point). We need to detect that error here and fail gracefully. Also the `uint8_t direction` variable could contain random data unless it's explicitly initialised, safest to set it to 0 even if we're sure it'll get initialised by later code.

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -637,6 +658,183 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
+			if (limit == 4096) {
+				Logger::error("River segment starting at (", points.front().x, ", ", points.front().y, ") is longer than limit 4096, check for misplaced pixels or other definition errors!");
+				river_complete = true;
+				break;
+			}
+
+			idx = points.back().x + points.back().y * rivers_bmp.get_width();
+
+			ivec2_t merge_location;
+			bool merge;
```
Hop311, line 745: This variable could be randomly true or false if not explicitly initialised (with MSVS at least), so it needs to be `bool merge = false;`

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -637,6 +658,183 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
+			// no further points
+			if (merge) points.push_back(merge_location);
+			river_complete = true;
+			break;
+		}
+
+		// TODO: implement optimisation algorithm to remove irrelevant points. Ramer-Douglas-Peucker?
+
+		// add segment then recursively call if neeeded
+		river.push_back({ size, points });
```
Hop311, line 817:
```suggestion
		river.push_back({ size, std::move(points) });
```
Move the `points` vector.

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -637,6 +658,183 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
+
+		if (river_complete) return;
+		next_segment(new_point, direction, river);
+	};
+
+	// find every river source and then run the segment algorithm.
+	int temp = 0;
```
Hop311, line 824: Leftover `temp` variable to remove?

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -637,6 +658,183 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
+	for (int y = 0; y < rivers_bmp.get_height(); ++y) {
+		for (int x = 0; x < rivers_bmp.get_width(); ++x) {
+			if (river_data[x + y * rivers_bmp.get_width()] == START_COLOUR) { // start of a river
+				river_t river;
+
+				next_segment({ x, y }, 0, river);
+
+				rivers.push_back(river);
```
Hop311, line 832:
```suggestion
				rivers.push_back(std::move(river));
```
Move the `river`

Hop311 (commented):
Great work, some minor code comments but the algorithm looks good. Could you test edge cases/invalid river formations just in case there are any other potential crash bugs? In the case of invalid data it doesn't matter much about loading the rest, just gracefully exiting the function without segfaulting

Hop311:
Also rather than storing every single point of each river, we could just store the corner points (so replace the last point if `direction == last_segment_direction`). This would require us to store the start point of the river as well (which I don't think we store at all atm?), but would save us a lot of memory/later processing time.

Also I'm wondering if it's necessary to have separate code for finding the first pixel vs the later pixels of a river segment? Could we find a way to handle all pixel with the same "next pixel finding" code?

.gitignore
```diff
@@ -36,7 +36,7 @@
 
 # VSCode
 .vscode/*
-!.vscode/launch.json
+.vscode/launch.json
```
Hop311, line 39: Did you intend to change this? The `!` means `.vscode/launch.json` isn't ignored despite the `.vscode/*` line above. If we should be ignoring the file then the line should be removed altogether as the line above will already catch it.

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -637,6 +659,196 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
+
+			// no further points
+			if (merge) points.push_back(merge_location);
+			river_complete = true;
+			break;
+		}
+
+		// save memory & simplify by storing only start, corner, and end points.
+		const auto is_corner_point = [](ivec2_t previous, ivec2_t current, ivec2_t next) {
+			return ((current.x - previous.x) * (next.y - current.y)) != ((current.y - previous.y) * (next.x - current.x)); // slope is fun
```
Hop311, line 821: Nice!

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -637,6 +659,196 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
+		next_segment(new_point, direction, river);
+	};
+
+	// find every river source and then run the segment algorithm.
+	for (int y = 0; y < rivers_bmp.get_height(); ++y) {
+		for (int x = 0; x < rivers_bmp.get_width(); ++x) {
+			if (river_data[x + y * rivers_bmp.get_width()] == START_COLOUR) { // start of a river
+				river_t river;
+
+				next_segment({ x, y }, 0, river);
```
Hop311, line 844: If I understand correctly the start point `{ x, y }` itself isn't included in the river, is it not needed for drawing the river or calculating crossings? If it is needed please include it in the river object

Hop311 (commented):
Very elegent corner detection 😃

.gitignore
BrickPi: Intended to change to prevent placing my local launch.json under source control, I'll remove the line.

src/openvic-simulation/map/MapDefinition.cpp
BrickPi: V2 doesn't appear to include the start pixel, 
image
image

Hop311 (approved):

PR #198, opened 2024-09-12
Calculate country and province modifier sums

Hop311:


src/openvic-simulation/InstanceManager.cpp
```diff
@@ -39,10 +39,16 @@ void InstanceManager::update_gamestate() {
 
 	Logger::info("Update: ", today);
 
+	map_instance.update_modifier_sums(today, definition_manager.get_modifier_manager().get_static_modifier_cache());
```
wvpm, line 42: Get the manager once and get the static modifier cache once. You can store them in a local variable and reuse it.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -118,8 +123,8 @@ namespace OpenVic {
 		IndexedMap<GovernmentType, GovernmentType const*> PROPERTY(government_flag_overrides);
 		GovernmentType const* PROPERTY(flag_government_type);
 		fixed_point_t PROPERTY(suppression_points);
-		fixed_point_t PROPERTY(infamy);
-		fixed_point_t PROPERTY(plurality);
+		fixed_point_t PROPERTY(infamy); // in 0-25 range
```
wvpm, line 126: Infamy isn't limited to 25

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -851,8 +857,10 @@ void CountryInstance::_update_military(DefineManager const& define_manager, Unit
 		max_supported_regiment_count += state->get_max_supported_regiments();
 	}
 
-	// TODO - apply country/tech modifiers to supply consumption
-	supply_consumption = 1;
+	supply_consumption = fixed_point_t::_1();
+	if (modifier_effect_cache.get_supply_consumption() != nullptr) {
```
wvpm, line 861: Store modifier_effect_cache.get_supply_consumption() in local variable instead of calling it twice.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -894,7 +902,11 @@ void CountryInstance::_update_military(DefineManager const& define_manager, Unit
 	military_power = military_power_from_land + military_power_from_sea + military_power_from_leaders;
 
 	// Mobilisation calculations
-	mobilisation_impact = 0; // TODO - apply ruling party's war policy
+	if (modifier_effect_cache.get_mobilization_impact() != nullptr) {
```
wvpm, line 905: variable (see above)

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -927,7 +939,119 @@ bool CountryInstance::update_rule_set() {
...
+	if (static_modifier_cache.get_disarming() != nullptr && is_disarmed()) {
+		base_modifier_sum.add_modifier(*static_modifier_cache.get_disarming(), this);
+	}
+	if (static_modifier_cache.get_war_exhaustion() != nullptr && war_exhaustion > 0) {
+		base_modifier_sum.add_modifier(*static_modifier_cache.get_war_exhaustion(), this, war_exhaustion);
+	}
+	if (static_modifier_cache.get_infamy() != nullptr && infamy > 0) {
+		base_modifier_sum.add_modifier(*static_modifier_cache.get_infamy(), this, infamy);
+	}
+	if (static_modifier_cache.get_literacy() != nullptr && infamy > 0) {
```
wvpm, line 991: infamy > 0 makes no sense here

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -927,7 +939,119 @@ bool CountryInstance::update_rule_set() {
...
+		base_modifier_sum.add_modifier(*static_modifier_cache.get_literacy(), this, national_literacy);
+	}
+	if (static_modifier_cache.get_plurality() != nullptr && infamy > 0) {
```
wvpm, line 994: infamy > 0 makes no sense here

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -927,7 +939,119 @@ bool CountryInstance::update_rule_set() {
...
+		base_modifier_sum.add_modifier(*static_modifier_cache.get_plurality(), this, plurality);
+	}
+	// TODO - difficulty modifiers, war, peace, debt_default_to, bad_debter, generalised_debt_default,
+	//        total_occupation, total_blockaded, in_bankrupcy
+
+	// TODO - handle triggered modifiers
+
+	if (ruling_party != nullptr) {
+		for (Issue const* issue : ruling_party->get_policies()) {
+			if (issue != nullptr) {
```
wvpm, line 1004: How can issue even be a nullptr here?
Does get_policy return bogus pointers?

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -927,7 +939,119 @@ bool CountryInstance::update_rule_set() {
...
+
+	if (national_value != nullptr) {
+		base_modifier_sum.add_modifier(*national_value, this);
+	}
+
+	if (tech_school != nullptr) {
+		base_modifier_sum.add_modifier(*tech_school, this);
+	}
+
+	for (Technology const& technology : *unlocked_technologies.get_keys()) {
```
wvpm, line 1024: unlocked_technologies is a poor name for the map since it contains all technologies unlocked or not.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -927,7 +939,119 @@ bool CountryInstance::update_rule_set() {
...
+		if (is_technology_unlocked(technology)) {
+			base_modifier_sum.add_modifier(technology, this);
+		}
+	}
+
+	for (Invention const& invention : *unlocked_inventions.get_keys()) {
```
wvpm, line 1030: same thing as with unlocked_technologies

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -927,7 +939,119 @@ bool CountryInstance::update_rule_set() {
...
+		}
+	}
+
+	resultant_modifier_sum = base_modifier_sum;
+
+	// Add province base modifiers (with local province modifier effects removed)
+	using enum ModifierEffect::target_t;
+	static constexpr ModifierEffect::target_t NOT_PROVINCE = ALL_TARGETS & ~PROVINCE;
+
+	for (ProvinceInstance* province : owned_provinces) {
```
wvpm, line 1042: Why not derefence the province once instead of per method call?

src/openvic-simulation/economy/BuildingType.cpp
```diff
@@ -112,18 +112,27 @@ bool BuildingTypeManager::load_buildings_file(
 	)(root);
 	lock_building_types();
 
-	for (BuildingType const& building_type : building_types.get_items()) {
+	IndexedMap<BuildingType, ModifierEffectCache::building_type_effects_t>& building_type_effects =
+		modifier_manager.modifier_effect_cache.building_type_effects;
+
+	building_type_effects.set_keys(&get_building_types());
+
+	for (BuildingType const& building_type : get_building_types()) {
```
wvpm, line 120: Didn't you change this to hardcoded forts, naval bases and railroads?

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +191,73 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
 	}
 }
 
+void ProvinceInstance::update_modifier_sum(Date today, StaticModifierCache const& static_modifier_cache) {
```
wvpm, line 194: Shouldn't the province update the owner?
Now we're looping through all provinces twice per day tick. Once via map and once via country.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +191,73 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+
+	for (BuildingInstance const& building : buildings.get_items()) {
+		base_modifier_sum.add_modifier(building.get_building_type(), this);
+	}
+
+	if (crime != nullptr) {
+		base_modifier_sum.add_modifier(*crime, this);
+	}
+
+	if (province_definition.get_continent() != nullptr) {
```
wvpm, line 242: Can continent even be null?

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +191,73 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+		base_modifier_sum.add_modifier(*province_definition.get_continent(), this);
+	}
+
+	if (province_definition.get_climate() != nullptr) {
```
wvpm, line 246: Can climate even be null?

src/openvic-simulation/modifier/ModifierEffect.hpp
```diff
@@ -0,0 +1,53 @@
...
+	struct ModifierManager;
+
+	struct ModifierEffect : HasIdentifier {
+		friend struct ModifierManager;
+
+		enum class format_t : uint8_t {
+			PROPORTION_DECIMAL,	/* An unscaled fraction/ratio, with 1 being "full"/"whole" */
+			PERCENTAGE_DECIMAL,	/* A fraction/ratio scaled so that 100 is "full"/"whole" */
+			RAW_DECIMAL,		/* A continuous quantity, e.g. attack strength */
+			INT					/* A discrete quantity, e.g. building count limit */
```
wvpm, line 19:
```suggestion
			INT			/* A discrete quantity, e.g. building count limit */
```

src/openvic-simulation/modifier/ModifierEffectCache.cpp
wvpm, line 1: Nice solution

src/openvic-simulation/modifier/ModifierSum.cpp
```diff
@@ -19,38 +21,46 @@ bool ModifierSum::has_effect(ModifierEffect const& effect) const {
 	return value_sum.has_effect(effect);
 }
 
-void ModifierSum::add_modifier(Modifier const& modifier, fixed_point_t multiplier) {
-	modifiers[&modifier] += multiplier;
-	value_sum.multiply_add(modifier, multiplier);
+void ModifierSum::add_modifier(
+	Modifier const& modifier, modifier_source_t source, fixed_point_t multiplier, ModifierEffect::target_t targets
+) {
+	modifiers.push_back({ &modifier, multiplier, source, targets });
```
wvpm, line 27: Why use push_back here instead of emplace_back?
(I'm still learning when to use which)

src/openvic-simulation/modifier/ModifierValue.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+ModifierValue& ModifierValue::operator+=(ModifierValue const& right) {
+	for (effect_map_t::value_type const& value : right.values) {
+		values[value.first] += value.second;
+	}
+	return *this;
+}
+
+ModifierValue ModifierValue::operator+(ModifierValue const& right) const {
+	ModifierValue copy = *this;
```
wvpm, line 63: Thanks for clarity

wvpm (changes requested):
Loop through all provinces once per day, not twice.

src/openvic-simulation/economy/BuildingType.cpp
Hop311: No, that was just the rules controlling whether they can be built. Min/max level modifiers still exist for every building type (we can restrict them if necessary in a later PR).

src/openvic-simulation/InstanceManager.cpp
Hop311: Yeah it might be worth switching over to that approach for arguments that are always the same, however the current approach is used all over the codebase so it's better to stay consistent here and make the change everywhere in one go in a separate PR if we decide to go that way.

src/openvic-simulation/country/CountryInstance.cpp
Hop311: Party policies are stored in an `IndexedMap` so there's an entry for every `IssueGroup` regardless of whether it's actually set. We'd usually expect parties to have all policies set but it's possible for a mod to miss some in which case there would be a `nullptr` in the policy map.

src/openvic-simulation/map/ProvinceInstance.cpp
Hop311: Yes, provinces' continents are defined by lists of province IDs for each continent, if a province isn't in any of those lists then it's continent will be `nullptr`. For example: water provinces usually have no continent, but we still need to calculate their modifiers for stuff naval attrition.

src/openvic-simulation/map/ProvinceInstance.cpp
Hop311: Same as above, if a province isn't present in any climate's list of provinces then it will have `nullptr` for its climate.

src/openvic-simulation/modifier/ModifierEffect.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "openvic-simulation/types/EnumBitfield.hpp"
+#include "openvic-simulation/types/HasIdentifier.hpp"
+
+namespace OpenVic {
+	struct ModifierManager;
+
+	struct ModifierEffect : HasIdentifier {
+		friend struct ModifierManager;
+
+		enum class format_t : uint8_t {
```
wvpm, line 15: Great comments explaining the formats

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -927,7 +932,110 @@ bool CountryInstance::update_rule_set() {
...
+	static constexpr ModifierEffect::target_t NOT_PROVINCE = ALL_TARGETS & ~PROVINCE;
+
+	for (ProvinceInstance const* province : owned_provinces) {
+		modifier_sum.add_modifier_sum_filter_targets(province->get_modifier_sum(), NOT_PROVINCE);
+	}
+
+	// This has to be done after adding all province modifiers to the country's sum, otherwise provinces earlier in the list
+	// wouldn't be affected by modifiers from provinces later in the list.
+	for (ProvinceInstance* province : owned_provinces) {
+		province->contribute_country_modifier_sum(modifier_sum);
```
wvpm, line 1021: This still feels misplaced.

I prefer the following

1. loop through provinces to calculate their modifier sum (using province modifiers and such)
2. calculate country modifier sum by looping through all owned (and controlled?) provinces
3. perform daily updates such as pop growth and production by summing the province modifier sum + country modifier sum

1. is currently implemented in the map instance loop in this PR.
2. is currently implemented in the country instance in this PR (line 1015)

That just leaves this code to be moved to the province responsibility.
As we discussed on Discord, you want to try approaches, so I'll approve anyway.
This comment is mainly for clarity and documentation.

wvpm (approved):

src/openvic-simulation/country/CountryInstance.cpp
wvpm: I like the new names.

src/openvic-simulation/country/CountryInstance.cpp
wvpm: I like the new names.

src/openvic-simulation/country/CountryInstance.cpp
wvpm: No longer and issue

src/openvic-simulation/country/CountryInstance.cpp
wvpm: No longer applies

src/openvic-simulation/country/CountryInstance.cpp
wvpm: no longer applies

src/openvic-simulation/country/CountryInstance.cpp
wvpm: No longer applies

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -80,6 +81,10 @@ namespace OpenVic {
 		ordered_set<ProvinceInstance*> PROPERTY(core_provinces);
 		ordered_set<State*> PROPERTY(states);
 
+		ModifierSum PROPERTY(base_modifier_sum);
+		ModifierSum PROPERTY(resultant_modifier_sum);
```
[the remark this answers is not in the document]
Hop311: These are now changed and their replacement variables have comments explaining what they are and what they're used for.

src/openvic-simulation/map/ProvinceInstance.cpp
Hop311: I handling the province-to-owner contributions on the province side, however there's an issue: we'd still need to clear each country's modifier sum before adding any province contributions. Still, that may be worth it to avoid a loop over every owned province.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +191,73 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+		base_modifier_sum.add_modifier(*province_definition.get_climate(), this);
+	}
+
+	if (terrain_type != nullptr) {
```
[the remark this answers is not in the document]
Hop311: `terrain_type` should always be non-null, with a `default_terrain_type` calculated for every province on game load. However it's still safer to double check as it's possible for the `default_terrain_type` calculation to fail for a province, which would result in a null `terrain_type` if it wasn't changed in province history.  Equally, if something fails in the province history loading and applying process then that could result in a province's `terrain_type` being set to `nullptr`.

src/openvic-simulation/modifier/ModifierSum.cpp
Hop311: Usually `push_back` either copies (when given `value_type const&`) or moves (when given `value_type&&`) an already-constructed object into a container.
In contrast, `emplace_back` takes a list of arguments which are used to construct a new object, so `emplace_back(a, b, c)` is equivalent to `push_back(value_type { a, b, c })`. In some cases either function can be used, in particular via the copy or move constructor on the `emplace_back` side.

This specific case has changed to `modifiers.emplace_back(&modifier, multiplier, source, targets);` as we can use `emplace_back` to construct the object directly. Throughout the repo there are other objects which we can't normally construct using `emplace_back` as their constructors are private/protected and can only be called from a friend struct. This is because `emplace_back(a, b, c)` results in a constructor call from inside `emplace_back`, while `emplace_back({ a, b, c })` calls the constructor from the parent function before passing the object on to be move constructed.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +190,120 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
 }
 
+void ProvinceInstance::update_modifier_sum(Date today, StaticModifierCache const& static_modifier_cache) {
+	// Update sum of direct province modifiers
+	modifier_sum.clear();
+
+	const ModifierSum::modifier_source_t province_source { this };
+
+	// Erase expired event modifiers and add non-expired ones to the sum
+	std::erase_if(event_modifiers, [this, today, &province_source](ModifierInstance const& modifier) -> bool {
```
wvpm, line 200: Why `this` captured?

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +190,120 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+	modifier_sum.add_modifier_sum_exclude_source(owner_modifier_sum, this);
+}
+
+fixed_point_t ProvinceInstance::get_modifier_effect_value(ModifierEffect const& effect) const {
+	if constexpr (ADD_OWNER_CONTRIBUTION) {
+		return modifier_sum.get_effect(effect);
+	} else {
+		using enum ModifierEffect::target_t;
+
+		if (owner != nullptr) {
```
wvpm, line 250: Personally I prefer returning the happy flow at the end and treating other cases as short circuits.

```cpp
if (owner == nullptr) {
	return modifier_sum.get_effect(effect);
}

if ((effect.get_targets() & PROVINCE) == NO_TARGETS) {
	// Non-province targeted effects are already added to the country modifier sum
	return owner->get_modifier_effect_value(effect);
} else {
	// Province-targeted effects aren't passed to the country modifier sum
	return owner->get_modifier_effect_value(effect) + modifier_sum.get_effect(effect);
}
```

Alternatively you could keep the if else structure and replace the return statements with assinging a result variable.
Multiple returns (excluding short circuits) can be confusing for the reader.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +190,120 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+				return owner->get_modifier_effect_value(effect) + modifier_sum.get_effect(effect);
+			}
+		} else {
+			return modifier_sum.get_effect(effect);
+		}
+	}
+}
+
+fixed_point_t ProvinceInstance::get_modifier_effect_value_nullcheck(ModifierEffect const* effect) const {
+	if (effect != nullptr) {
```
wvpm, line 265: Similar to above, I'd prefer treating null as a short circuit here.
```cpp
	if (effect != nullptr) {
		return fixed_point_t::_0();
	}

	return get_modifier_effect_value(*effect);
```

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +190,120 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+			if ((effect.get_targets() & PROVINCE) == NO_TARGETS) {
+				// Non-province targeted effects are already added to the country modifier sum
+				owner->push_contributing_modifiers(effect, contributions);
+			} else {
+				// Province-targeted effects aren't passed to the country modifier sum
+				modifier_sum.push_contributing_modifiers(effect, contributions);
+				owner->push_contributing_modifiers(effect, contributions);
+			}
+		} else {
+			modifier_sum.push_contributing_modifiers(effect, contributions);
```
wvpm, line 290: This should only happen when `(effect.get_targets() & PROVINCE) != NO_TARGETS`.
Consider splitting the checks.

```cpp
if (owner != nullptr) {
   owner->push_contributing_modifiers(effect, contributions);
}

if ((effect.get_targets() & PROVINCE) != NO_TARGETS) {
   modifier_sum.push_contributing_modifiers(effect, contributions);
}

File: src/openvic-simulation/map/ProvinceInstance.cpp [hunk 26]
```diff
@@ -185,6 +190,120 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+		}
+	}
+}
+
+std::vector<ModifierSum::modifier_entry_t> ProvinceInstance::get_contributing_modifiers(ModifierEffect const& effect) const {
+	if constexpr (ADD_OWNER_CONTRIBUTION) {
+		return modifier_sum.get_contributing_modifiers(effect);
+	} else {
+		std::vector<ModifierSum::modifier_entry_t> contributions;
```
wvpm, line 299: Shouldn't you initialise it?

wvpm (dismissed):

src/openvic-simulation/map/ProvinceInstance.cpp
Hop311: All struct and class variables get initialised automatically (if they have a default constructor). It's only primitives that always need explicit initialisation (so basic integer and float types, enums, pointers, references, etc.), and even then that's mainly because of MSVC, I believe other compilers will auto initialise even primitive variables.

src/openvic-simulation/modifier/ModifierEffectCache.cpp
```diff
@@ -196,7 +200,7 @@ ModifierEffectCache::ModifierEffectCache()
 	/* Military Modifier Effects */
 	attack { nullptr },
```
wvpm, line 201: attack_leader for consistency

src/openvic-simulation/research/Invention.cpp
```diff
@@ -74,15 +79,19 @@ bool InventionManager::load_inventions_file(
...
-			bool ret = modifier_manager.expect_modifier_value_and_keys(move_variable_callback(loose_modifiers),
+			bool ret = modifier_manager.expect_modifier_value_and_keys(
+				move_variable_callback(loose_modifiers),
+				INVENTION,
 				"news", ZERO_OR_ONE, expect_bool(assign_variable_callback(news)),
 				"limit", ONE_EXACTLY, limit.expect_script(),
 				"chance", ONE_EXACTLY, chance.expect_conditional_weight(ConditionalWeight::BASE),
 				"effect", ZERO_OR_ONE, modifier_manager.expect_modifier_value_and_keys(
 					move_variable_callback(modifiers),
+					INVENTION,
```
wvpm, line 90: There is no special invention thing here, this is just a country modifier.

src/openvic-simulation/politics/NationalValue.cpp
```diff
@@ -19,11 +21,14 @@ bool NationalValueManager::load_national_values_file(ModifierManager const& modi
 	bool ret = expect_dictionary_reserve_length(
 		national_values,
 		[this, &modifier_manager](std::string_view national_value_identifier, ast::NodeCPtr value) -> bool {
+			using enum Modifier::modifier_type_t;
+
 			ModifierValue modifiers;
 
-			bool ret = modifier_manager.expect_modifier_value(move_variable_callback(modifiers))(value);
+			bool ret = modifier_manager.expect_modifier_value(move_variable_callback(modifiers), NATIONAL_VALUE)(value);
```
wvpm, line 28: There is no special national value modifier, this is just a country modifier.

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -223,6 +231,7 @@ bool IssueManager::_load_reform(
 
 	bool ret = modifier_manager.expect_modifier_value_and_keys(
 		move_variable_callback(values),
+		REFORM,
```
wvpm, line 234: There is no distinction between reform modifiers and issue modifiers.

wvpm (approved):

src/openvic-simulation/modifier/ModifierEffectCache.cpp
Hop311: IMO it's more consistent to use `attack` - these variable names correspond to the `ModiferEffect`s' identifier strings, so `attack` has identifier `"attack"` and `defence_leader` has identifier `"defence leader"`. This makes it clear exactly what the variable corresponds to, including the fact that the `attack` effect could technically appear in places other than leader traits while `defence_leader` is guarunteed to only come from leader traits (we may want to enforce similar restrictions for effects like `attack`, but that can come in a later PR which can change the variable names too if necessary).

src/openvic-simulation/research/Invention.cpp
```diff
@@ -74,15 +79,19 @@ bool InventionManager::load_inventions_file(
 			ConditionScript limit { scope_t::COUNTRY, scope_t::COUNTRY, scope_t::NO_SCOPE };
 			ConditionalWeight chance { scope_t::COUNTRY, scope_t::COUNTRY, scope_t::NO_SCOPE };
 
-			bool ret = modifier_manager.expect_modifier_value_and_keys(move_variable_callback(loose_modifiers),
+			bool ret = modifier_manager.expect_modifier_value_and_keys(
+				move_variable_callback(loose_modifiers),
+				INVENTION,
```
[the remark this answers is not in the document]
Hop311: The modifier type enum is used for modifier variants in some cases, but it's also used for UI logic (e.g. some tooltips group all tech contributions into a single "From tech" line). Although the `INVENTION` argument doesn't do anything special here, we need something to fill the argument slot (and to let the function know this isn't a modifier using effect variants) so might as well use the specific type rather than a less descriptive generic null value.

src/openvic-simulation/modifier/ModifierEffectCache.cpp
wvpm: My comment wasn't about the origin of the modifier, it is about the effect of the modifier.
attack means a boost to the roll from the leader. This is how it shows up in the UI. It's unrelated to attack stats for units.
Thus attack technically can't appear in places other than leader background/traits as it is scoped to the leader.

Hop311: I mentioned the modifier's origin because that's what the modifier effect cache variable names are currently based on, so they provide useful information about what the variable actually means (IMO the origin information is more useful because that's dependent on the code and is difficult to manually determine, whereas the use information is fixed regardless of the game's code/implementation).
And while it's correct that the `attack` modifier effect shouldn't appear anywhere other than leader traits, the current implementation doesn't force that, meaning `attack` effects in country and province modifiers will be parsed and included in modifier sums. Whether these potential misuses of the effect actually get used in battle logic depends on its implementation, but we should probably restrict the effect's use regardless. When we have such a restriction we could change the variable's name to represent the fact that it definitely only appears in leader traits, but until then the current name serves as a warning that the variable just represents the `attack` effect without any context restrictions.

wvpm: I think there could be arguments for allowing attack modifiers from other origins:

- +1 attack when fighting on core (similar to hoi4)
- military academy building
- school of the attack invention
- military industrial complex tech school

Just some ideas that might be interesting for a mod.
The point is that attack in all these cases means attack bonus from leader in the UI, not gas attack or some other modifier.

PR #201, opened 2024-09-18
ModifierEffect and Rule localisation keys

Hop311:
There are a few modifiers which don't have localisation keys, e.g. `min_build_***` and some `***_RGO_SIZE` modifiers.

### DIFF
--- a/src/openvic-simulation/economy/GoodDefinition.cpp
+++ b/src/openvic-simulation/economy/GoodDefinition.cpp
@@ -88,29 +88,46 @@ bool GoodDefinitionManager::load_goods_file(ast::NodeCPtr root) {
 }
 
 bool GoodDefinitionManager::generate_modifiers(ModifierManager& modifier_manager) const {
+	using enum ModifierEffect::format_t;
+
 	bool ret = true;
 
-	const auto good_modifier = [this, &modifier_manager, &ret](std::string_view name, bool is_positive_good) -> void {
+	const auto good_modifier = [this, &modifier_manager, &ret](
+		std::string_view name, bool is_positive_good, auto make_localisation_suffix
+	) -> void {
 		ret &= modifier_manager.register_complex_modifier(name);
+
 		for (GoodDefinition const& good : get_good_definitions()) {
 			ret &= modifier_manager.add_modifier_effect(
-				ModifierManager::get_flat_identifier(name, good.get_identifier()), is_positive_good
+				ModifierManager::get_flat_identifier(name, good.get_identifier()), is_positive_good, PROPORTION_DECIMAL,
+				make_localisation_suffix(good.get_identifier())
 			);
 		}
 	};
 
-	good_modifier("artisan_goods_input", false);
-	good_modifier("artisan_goods_output", true);
-	good_modifier("artisan_goods_throughput", true);
-	good_modifier("factory_goods_input", false);
-	good_modifier("factory_goods_output", true);
-	good_modifier("factory_goods_throughput", true);
-	good_modifier("rgo_goods_output", true);
-	good_modifier("rgo_goods_throughput", true);
-	good_modifier("rgo_size", true);
+	const auto make_production_localisation_suffix = [](std::string_view localisation_suffix) -> auto {
+		return [localisation_suffix](std::string_view good_identifier) -> std::string {
+			return StringUtils::append_string_views("$", good_identifier,  "$ $", localisation_suffix, "$");
+		};
+	};
+
+	good_modifier("artisan_goods_input", false, make_production_localisation_suffix("TECH_INPUT"));
+	good_modifier("artisan_goods_output", true, make_production_localisation_suffix("TECH_OUTPUT"));
+	good_modifier("artisan_goods_throughput", true, make_production_localisation_suffix("TECH_THROUGHPUT"));
+	good_modifier("factory_goods_input", false, make_production_localisation_suffix("TECH_INPUT"));
+	good_modifier("factory_goods_output", true, make_production_localisation_suffix("TECH_OUTPUT"));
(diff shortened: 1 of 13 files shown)


src/openvic-simulation/politics/Rule.cpp
```diff
@@ -183,16 +191,6 @@ bool RuleManager::setup_rules(BuildingTypeManager const& building_type_manager)
 		}
 	}
 
-	for (std::string const& type : building_type_manager.get_building_type_types()) {
```
wvpm, line 186: Why remove this?
Now we lack naval_base & fort rules.

src/openvic-simulation/utility/StringUtils.hpp
```diff
@@ -160,6 +160,22 @@ namespace OpenVic::StringUtils {
 		return std::equal(lhs.begin(), lhs.end(), rhs.begin(), rhs.end(), ichar_equals);
 	}
 
+	inline constexpr std::string string_tolower(std::string_view str) {
```
wvpm, line 163: Wouldn't to_lower suffice?

src/openvic-simulation/utility/StringUtils.hpp
```diff
@@ -160,6 +160,22 @@ namespace OpenVic::StringUtils {
...
+		std::string result { str };
+		std::transform(result.begin(), result.end(), result.begin(),
+			[](unsigned char c) -> unsigned char { return std::tolower(c); }
+		);
+		return result;
+	}
+
+	inline constexpr std::string string_toupper(std::string_view str) {
```
wvpm, line 171: Wouldn't to_upper suffice?

src/openvic-simulation/economy/GoodDefinition.cpp
```diff
@@ -88,29 +88,46 @@ bool GoodDefinitionManager::load_goods_file(ast::NodeCPtr root) {
...
-	good_modifier("artisan_goods_input", false);
-	good_modifier("artisan_goods_output", true);
-	good_modifier("artisan_goods_throughput", true);
-	good_modifier("factory_goods_input", false);
-	good_modifier("factory_goods_output", true);
-	good_modifier("factory_goods_throughput", true);
-	good_modifier("rgo_goods_output", true);
-	good_modifier("rgo_goods_throughput", true);
-	good_modifier("rgo_size", true);
+	const auto make_production_localisation_suffix = [](std::string_view localisation_suffix) -> auto {
```
wvpm, line 108: This doesn't distinguish between factory, rgo or artisan goods modifier.

wvpm (changes requested):
Distinguish between factory, rgo or artisan goods modifiers & keep support for naval_base & fort.

src/openvic-simulation/utility/StringUtils.hpp
Hop311: Do you mean `godot::String`'s `to_lower` function? That's not available in the SIM submodule, all we have access to here is `std::tolower` which only converts a single character

src/openvic-simulation/economy/GoodDefinition.cpp
Hop311: Vic2 doesn't distinguish between them, the tooltips all say `<good> production <input/output/throughput>` regardless of whether it's affecting factory, rgo or artisan production

src/openvic-simulation/utility/StringUtils.hpp
```diff
@@ -160,6 +160,22 @@ namespace OpenVic::StringUtils {
 		return std::equal(lhs.begin(), lhs.end(), rhs.begin(), rhs.end(), ichar_equals);
 	}
 
+	inline constexpr std::string string_tolower(std::string_view str) {
+		std::string result { str };
+		std::transform(result.begin(), result.end(), result.begin(),
+			[](unsigned char c) -> unsigned char { return std::tolower(c); }
```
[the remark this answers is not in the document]
Hop311: True, `godot::String`'s `to_lower`/`to_upper` functions are better for general case conversion. These work fine for our current uses in the SIM submodule (modifier localisation keys), and none of the new modifiers mods can create (e.g. for new goods) are passed through these case functions

src/openvic-simulation/politics/Rule.cpp
Hop311: Vic2 doesn't support naval_base and fort building rules and has no localisation for them.
Although the fort and naval base button tooltips say "Economic policy allows building" similar to the railroad button, adding `build_fort = no` or `build_naval_base = no` to an economic policy has no effect - they don't show up in the tooltip at all, not even as unlocalised rules.

src/openvic-simulation/economy/GoodDefinition.cpp
schombert: Yes, but does the game still distinguish them internally? I believe that it does

src/openvic-simulation/politics/Rule.cpp
schombert: what about the extra buildings that mods add?

src/openvic-simulation/economy/GoodDefinition.cpp
Hop311: We do distinguish them internally, this is just the tooltips he's talking about

wvpm: Confusing. We should support different tooltips. Perhaps we can fallback to the generalised one?

Hop311: We could for custom mode, but for compat mode I don't think it's worth the extra complication this would cause to the modifier localisation system

wvpm (approved):

PR #203, opened 2024-09-18
Add ModifierSum and improve ModifierValue

Hop311:


src/openvic-simulation/misc/Modifier.cpp
```diff
@@ -67,6 +67,14 @@ bool ModifierValue::has_effect(ModifierEffect const* effect) const {
 	return values.contains(effect);
 }
 
+void ModifierValue::set_effect(ModifierEffect const* effect, fixed_point_t value) {
```
wvpm, line 70: Can't you use a reference instead of a pointer here?
That would save the nullcheck.

src/openvic-simulation/misc/Modifier.cpp
```diff
@@ -99,6 +107,32 @@ ModifierValue ModifierValue::operator-(ModifierValue const& right) const {
...
+ModifierValue& ModifierValue::operator*=(fixed_point_t const& right) {
+	for (auto value : mutable_iterator(values)) {
+		value.second *= right;
+	}
+	return *this;
+}
+
+ModifierValue ModifierValue::operator*(fixed_point_t const& right) const {
+	ModifierValue ret = *this;
+	return ret *= right;
```
wvpm, line 119: I'm struggling to understand this.
You dereference the pointer to get ret and then enter the above operator which modifies the value and returns *this.
Wouldn't the original value be changed then?

src/openvic-simulation/misc/Modifier.cpp
```diff
@@ -110,8 +144,63 @@ bool TriggeredModifier::parse_scripts(DefinitionManager const& definition_manage
...
+	modifiers.clear();
+	value_sum.clear();
+}
+
+bool ModifierSum::empty() {
+	return modifiers.empty();
+}
+
+fixed_point_t ModifierSum::get_effect(ModifierEffect const* effect, bool* successful) const {
+	return value_sum.get_effect(effect, successful);
```
wvpm, line 160: Why would get_effect not be successful?
Is this some sort of try catch for errors?

src/openvic-simulation/misc/Modifier.hpp
```diff
@@ -105,10 +112,35 @@ namespace OpenVic {
...
-		Modifier const& PROPERTY(modifier);
+		Modifier const* PROPERTY(modifier); // We can assume this is never null
 		Date PROPERTY(expiry_date);
 
-		ModifierInstance(Modifier const& modifier, Date expiry_date);
+	public:
+		ModifierInstance(Modifier const& new_modifier, Date new_expiry_date);
+	};
+
+	struct ModifierSum {
```
wvpm, line 122: Modifier is pretty big. It might be better to move the different structs to different files.
Modifiers can be moved from /misc to /modfiers then

wvpm (commented):
Looks okay but I lack the usage to put things into perspective.
Also left a few comments.

src/openvic-simulation/misc/Modifier.cpp
Hop311: Good point, I'll change that for this and similar functions in `ModifierValue` and `ModifierSum`

src/openvic-simulation/misc/Modifier.cpp
Hop311: `ModifierValue ret = *this;` creates a copy of `*this` and stores it in `ret`. We need to do this as `a * b` returns a new value rather than modifying `a` or `b`.

src/openvic-simulation/misc/Modifier.hpp
Hop311: Yeah I was thinking that too, I'll do that in an extra commit on this PR

src/openvic-simulation/misc/Modifier.cpp
Hop311: It's to detect if the lookup is successful, i.e. if there is a value for the effect in the `ModifierSum` or `ModifierValue`. This allows us to tell the difference between a modifier with an effect explicitly listed with value 0 (e.g. tech schools include every research type modifier, even if the value is 0) vs a modifier having no entry for the effect at all.

schombert: optional would be a better fit for that IMO instead of an out parameter (which you will then have to check for null)

Hop311: Null is allowed, it's the default value and just means we don't set the success flag. In most cases we don't care if the value is an explicit or implicit 0, and using optional would complicate those cases

src/openvic-simulation/misc/Modifier.cpp
wvpm: Could you rename ret to copy for clarity?

src/openvic-simulation/misc/Modifier.cpp
wvpm: Could you rename it to something like is_present or has_value?
Successful sounds like a try catch.

src/openvic-simulation/modifier/Modifier.cpp
```diff
@@ -49,8 +49,8 @@ bool ModifierValue::empty() const {
 	return values.empty();
 }
 
-fixed_point_t ModifierValue::get_effect(ModifierEffect const* effect, bool* successful) {
-	const effect_map_t::const_iterator it = values.find(effect);
+fixed_point_t ModifierValue::get_effect(ModifierEffect const& effect, bool* successful) const {
```
wvpm, line 52: Rename `successful` to `has_value` or `is_present` or `has_effect` to clarify it's a lookup instead of a try catch scenario.

src/openvic-simulation/modifier/Modifier.cpp
```diff
@@ -99,6 +103,32 @@ ModifierValue ModifierValue::operator-(ModifierValue const& right) const {
...
 
+ModifierValue& ModifierValue::operator*=(fixed_point_t const& right) {
+	for (auto value : mutable_iterator(values)) {
+		value.second *= right;
+	}
+	return *this;
+}
+
+ModifierValue ModifierValue::operator*(fixed_point_t const& right) const {
+	ModifierValue ret = *this;
```
wvpm, line 114: Rename `ret` to `copy` for clarity.

wvpm (approved):

src/openvic-simulation/misc/Modifier.cpp
schombert: > Null is allowed, it's the default value and just means we don't set the success flag. In most cases we don't care if the value is an explicit or implicit 0, and using optional would complicate those cases

That is what the `value_or` member function is for ( https://en.cppreference.com/w/cpp/utility/optional/value_or )
If you don't care whether the value is present you would do `get_effect(...).value_or(0)` which is ergonomic and avoids easy-to-misuse out parameters

wvpm (dismissed):

PR #204, opened 2024-09-20
Fix modifiers

Hop311:
- Fix unit modifier localisation keys
- Move national focus specific effects to custom variables rather than general modifier effects (this gets around issues with their localisation too)

### DIFF
--- a/src/openvic-simulation/military/UnitType.cpp
+++ b/src/openvic-simulation/military/UnitType.cpp
@@ -322,11 +322,11 @@ bool UnitTypeManager::generate_modifiers(ModifierManager& modifier_manager) cons
 		std::string_view identifier, UnitType::branch_t branch
 	) -> void {
 		const auto stat_modifier = [&modifier_manager, &ret, &identifier](
-			std::string_view suffix, bool is_positive_good, ModifierEffect::format_t format
+			std::string_view suffix, bool is_positive_good, ModifierEffect::format_t format, std::string_view localisation_key
 		) -> void {
 			ret &= modifier_manager.add_modifier_effect(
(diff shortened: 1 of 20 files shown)


src/openvic-simulation/military/UnitType.cpp
```diff
@@ -322,41 +322,41 @@ bool UnitTypeManager::generate_modifiers(ModifierManager& modifier_manager) cons
...
 		case LAND:
-			stat_modifier("reconnaissance", true, RAW_DECIMAL);
-			stat_modifier("discipline", true, PROPORTION_DECIMAL);
-			stat_modifier("support", true, PROPORTION_DECIMAL);
-			stat_modifier("maneuver", true, INT);
-			stat_modifier("siege", true, RAW_DECIMAL);
+			stat_modifier("reconnaissance", true, RAW_DECIMAL, "RECONAISSANCE");
+			stat_modifier("discipline", true, PROPORTION_DECIMAL, "DISCIPLINE");
+			stat_modifier("support", true, PROPORTION_DECIMAL, "SUPPORT");
+			stat_modifier("maneuver", true, INT, "Maneuver");
```
wvpm, line 349:
```suggestion
			stat_modifier("maneuver", true, INT, "MANUEVER");
```

wvpm (approved):
Code looks good apart from MANUEVER.
It's a shame this restricts capitalist behaviour modifiers to national foci though.

src/openvic-simulation/military/UnitType.cpp
Hop311: I checked in game, the localisation key is `Maneuver`, it shows that even if you delete every reference to it in the csv files. The non-English versions are in `housedivided2_2.csv` (under the key `Maneuver`).

PR #205, opened 2024-09-20
Modifier types

Hop311:
These will be useful for organising modifiers in tooltips, e.g. collecting together tech into a single "From tech" line


BrickPi (approved):
LGTM, such a paradox thing to do having a dozen different modifier tooltip types under the hood.

PR #206, opened 2024-09-22
Prepare for rgo

wvpm:
1. trade_goods from history are converted to actual RGO instances.
2. pops are converted to equivalents if applicable (farmer <> labourer).
3. RGO size is calculated.
4. Province rgo can be changed (this will convert pops and update the rgo).
5. When initialising a new game RGO hires workers, produces good and gives income to owners and employees.

### DIFF
--- a/src/headless/main.cpp
+++ b/src/headless/main.cpp
@@ -1,7 +1,10 @@
-#include <cstring>
-
+#include <openvic-simulation/country/CountryInstance.hpp>
 #include <openvic-simulation/dataloader/Dataloader.hpp>
+#include <openvic-simulation/economy/GoodDefinition.hpp>
+#include <openvic-simulation/economy/production/ProductionType.hpp>
+#include <openvic-simulation/economy/production/ResourceGatheringOperation.hpp>
 #include <openvic-simulation/GameManager.hpp>
+#include <openvic-simulation/pop/Pop.hpp>
 #include <openvic-simulation/testing/Testing.hpp>
 #include <openvic-simulation/utility/Logger.hpp>
 
@@ -16,6 +19,40 @@ static void print_help(std::ostream& stream, char const* program_name) {
 		<< "    -s : Use the following path as a hint to search for a base directory.\n"
 		<< "Any following paths are read as mod directories, with priority starting at one above the base directory.\n"
 		<< "(Paths with spaces need to be enclosed in \"quotes\").\n";
+}
+
+static void print_rgo(ProvinceInstance const& province) {
+	ResourceGatheringOperation const& rgo = province.get_rgo();
+	ProductionType const* const production_type_nullable = rgo.get_production_type_nullable();
+	if (production_type_nullable == nullptr) {
+		Logger::error(
+			"\n    ", province.get_identifier(),
+			" - production_type: nullptr"
+		);
+	} else {
+		ProductionType const& production_type = *production_type_nullable;
+		GoodDefinition const& output_good = production_type.get_output_good();
+		std::string text = StringUtils::append_string_views(
+			"\n\t", province.get_identifier(),
+			" - good: ", output_good.get_identifier(),
+			", production_type: ", production_type.get_identifier(),
+			", size_multiplier: ", rgo.get_size_multiplier().to_string(3),
+			", output_quantity_yesterday: ", rgo.get_output_quantity_yesterday().to_string(3),
+			", revenue_yesterday: ", rgo.get_revenue_yesterday().to_string(3),
+			", total owner income: ", rgo.get_total_owner_income_cache().to_string(3),
+			", total employee income: ", rgo.get_total_employee_income_cache().to_string(3),
+			"\n\temployees:"
+		);
+
+		auto const& employee_count_per_type_cache=rgo.get_employee_count_per_type_cache();
+		for (PopType const& pop_type : *employee_count_per_type_cache.get_keys()) {
+			const Pop::pop_size_t employees_of_type = employee_count_per_type_cache[pop_type];
+			if (employees_of_type > 0) {
+				text += StringUtils::append_string_views("\n\t\t", std::to_string(employees_of_type), " ", pop_type.get_identifier());
+			}
+		}
+		Logger::info("", text);
+	}
 }
 
 static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests) {
@@ -72,9 +109,21 @@ static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests)
 		CountryInstanceManager const& country_instance_manager =
 			game_manager.get_instance_manager()->get_country_instance_manager();
 
-		print_ranking_list("Great Powers", country_instance_manager.get_great_powers());
+		std::vector<CountryInstance*> const& great_powers = country_instance_manager.get_great_powers();
+		print_ranking_list("Great Powers", great_powers);
 		print_ranking_list("Secondary Powers", country_instance_manager.get_secondary_powers());
 		print_ranking_list("All countries", country_instance_manager.get_total_ranking());
+
+		Logger::info("===== RGO test... =====");
+		for (size_t i = 0; i < std::min<size_t>(3, great_powers.size()); ++i) {
+			CountryInstance const& great_power = *great_powers[i];
+			ProvinceInstance const* const capital_province = great_power.get_capital();
+			if (capital_province == nullptr) {
+				Logger::warning(great_power.get_identifier(), " has no capital ProvinceInstance set.");
+			} else {
+				print_rgo(*capital_province);
+			}
+		}
 	} else {
 		Logger::error("Instance manager not available!");
 		ret = false;
--- a/src/openvic-simulation/InstanceManager.cpp
+++ b/src/openvic-simulation/InstanceManager.cpp
@@ -38,32 +38,7 @@ void InstanceManager::update_gamestate() {
 	currently_updating_gamestate = true;
 
 	Logger::info("Update: ", today);
-
-	if constexpr (ProvinceInstance::ADD_OWNER_CONTRIBUTION) {
-		// Calculate local province modifier sums first, then national country modifier sums, then loop over owned provinces
-		// adding their contributions to the owner country's modifier sum and loop over them again to add the country's total
-		// (including province contributions) to the provinces' modifier sum. This results in every country and province
-		// having a full copy of all the modifiers affecting them in their modifier sum.
-		map_instance.update_modifier_sums(
-			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
-		);
-		country_instance_manager.update_modifier_sums(
-			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
-		);
-	} else {
-		// Calculate national country modifier sums first, then local province modifier sums, adding province contributions
-		// to owner countries' modifier sums if each province has an owner. This results in every country having a full copy
-		// of all the modifiers affecting them in their modifier sum, but provinces only having their directly/locally applied
-		// modifiers in their modifier sum, hence requiring both province and owner country modifier effect values to be looked
-		// up and added together to get the full effect on the province.
-		country_instance_manager.update_modifier_sums(
-			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
-		);
-		map_instance.update_modifier_sums(
-			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
-		);
-	}
-
+	update_modifier_sums();
 	// Update gamestate...
 	map_instance.update_gamestate(today, definition_manager.get_define_manager());
 	country_instance_manager.update_gamestate(
@@ -163,6 +138,11 @@ bool InstanceManager::load_bookmark(Bookmark const* new_bookmark) {
 		map_instance, definition_manager.get_pop_manager().get_pop_types()
 	);
 
+	if (ret) {
+		update_modifier_sums();
+		map_instance.initialise_for_new_game(definition_manager.get_modifier_manager().get_modifier_effect_cache());
+	}
+
 	return ret;
 }
 
@@ -203,4 +183,31 @@ bool InstanceManager::expand_selected_province_building(size_t building_index) {
 		return false;
 	}
 	return province->expand_building(building_index);
+}
+
+void InstanceManager::update_modifier_sums() {
+	if constexpr (ProvinceInstance::ADD_OWNER_CONTRIBUTION) {
+		// Calculate local province modifier sums first, then national country modifier sums, then loop over owned provinces
+		// adding their contributions to the owner country's modifier sum and loop over them again to add the country's total
+		// (including province contributions) to the provinces' modifier sum. This results in every country and province
+		// having a full copy of all the modifiers affecting them in their modifier sum.
+		map_instance.update_modifier_sums(
+			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
+		);
+		country_instance_manager.update_modifier_sums(
+			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
+		);
+	} else {
+		// Calculate national country modifier sums first, then local province modifier sums, adding province contributions
+		// to owner countries' modifier sums if each province has an owner. This results in every country having a full copy
+		// of all the modifiers affecting them in their modifier sum, but provinces only having their directly/locally applied
+		// modifiers in their modifier sum, hence requiring both province and owner country modifier effect values to be looked
+		// up and added together to get the full effect on the province.
+		country_instance_manager.update_modifier_sums(
+			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
+		);
+		map_instance.update_modifier_sums(
+			today, definition_manager.get_modifier_manager().get_static_modifier_cache()
+		);
+	}
 }
--- a/src/openvic-simulation/country/CountryInstance.hpp
+++ b/src/openvic-simulation/country/CountryInstance.hpp
@@ -139,7 +139,7 @@ namespace OpenVic {
 		// TODO - population change over last 30 days
 		fixed_point_t PROPERTY(national_consciousness);
 		fixed_point_t PROPERTY(national_militancy);
-		IndexedMap<PopType, fixed_point_t> PROPERTY(pop_type_distribution);
+		IndexedMap<PopType, Pop::pop_size_t> PROPERTY(pop_type_distribution);
 		size_t PROPERTY(national_focus_capacity)
 		// TODO - national foci
 
--- a/src/openvic-simulation/economy/production/Employee.hpp
+++ b/src/openvic-simulation/economy/production/Employee.hpp
@@ -1,0 +1,13 @@
+#pragma once
+
+#include "openvic-simulation/pop/Pop.hpp"
+
+namespace OpenVic {
+	struct Employee {
+	private:
+		Pop::pop_size_t PROPERTY_RW(size);
+	public:
+		Pop& pop;
+		Employee(Pop& new_pop, const Pop::pop_size_t new_size);
+	};
+}
--- a/src/openvic-simulation/economy/production/ProductionType.cpp
+++ b/src/openvic-simulation/economy/production/ProductionType.cpp
@@ -9,33 +9,33 @@ Job::Job(
 	PopType const* new_pop_type,
 	effect_t new_effect_type,
 	fixed_point_t new_effect_multiplier,
-	fixed_point_t new_desired_workforce_share
+	fixed_point_t new_amount
 ) : pop_type { new_pop_type },
 	effect_type { new_effect_type },
 	effect_multiplier { new_effect_multiplier },
-	desired_workforce_share { new_desired_workforce_share } {}
+	amount { new_amount } {}
 
 ProductionType::ProductionType(
-	std::string_view new_identifier,
-	std::optional<Job> new_owner,
+	const std::string_view new_identifier,
+	const std::optional<Job> new_owner,
 	std::vector<Job>&& new_jobs,
-	template_type_t new_template_type,
-	Pop::pop_size_t new_base_workforce_size,
+	const template_type_t new_template_type,
+	const Pop::pop_size_t new_base_workforce_size,
 	GoodDefinition::good_definition_map_t&& new_input_goods,
-	GoodDefinition const* new_output_goods,
-	fixed_point_t new_base_output_quantity,
+	GoodDefinition const& new_output_good,
+	const fixed_point_t new_base_output_quantity,
 	std::vector<bonus_t>&& new_bonuses,
 	GoodDefinition::good_definition_map_t&& new_maintenance_requirements,
-	bool new_is_coastal,
-	bool new_is_farm,
-	bool new_is_mine
+	const bool new_is_coastal,
+	const bool new_is_farm,
+	const bool new_is_mine
 ) : HasIdentifier { new_identifier },
 	owner { new_owner },
 	jobs { std::move(new_jobs) },
 	template_type { new_template_type },
 	base_workforce_size { new_base_workforce_size },
 	input_goods { std::move(new_input_goods) },
-	output_goods { new_output_goods },
+	output_good { new_output_good },
 	base_output_quantity { new_base_output_quantity },
 	bonuses { std::move(new_bonuses) },
 	maintenance_requirements { std::move(new_maintenance_requirements) },
@@ -51,7 +51,9 @@ bool ProductionType::parse_scripts(DefinitionManager const& definition_manager) 
 	return ret;
 }
 
-ProductionTypeManager::ProductionTypeManager() : rgo_owner_sprite { 0 } {}
+ProductionTypeManager::ProductionTypeManager() :
+	good_to_rgo_production_type { nullptr },
+	rgo_owner_sprite { 0 } {}
 
 node_callback_t ProductionTypeManager::_expect_job(
 	GoodDefinitionManager const& good_definition_manager, PopManager const& pop_manager, callback_t<Job&&> callback
@@ -92,19 +94,19 @@ node_callback_t ProductionTypeManager::_expect_job_list(
 }
 
 bool ProductionTypeManager::add_production_type(
-	std::string_view identifier,
+	const std::string_view identifier,
 	std::optional<Job> owner,
 	std::vector<Job>&& jobs,
-	ProductionType::template_type_t template_type,
-	Pop::pop_size_t base_workforce_size,
+	const ProductionType::template_type_t template_type,
+	const Pop::pop_size_t base_workforce_size,
 	GoodDefinition::good_definition_map_t&& input_goods,
-	GoodDefinition const* output_goods,
-	fixed_point_t base_output_quantity,
+	GoodDefinition const* const output_good,
+	const fixed_point_t base_output_quantity,
 	std::vector<ProductionType::bonus_t>&& bonuses,
 	GoodDefinition::good_definition_map_t&& maintenance_requirements,
-	bool is_coastal,
-	bool is_farm,
-	bool is_mine
+	const bool is_coastal,
+	const bool is_farm,
+	const bool is_mine
 ) {
 	if (identifier.empty()) {
 		Logger::error("Invalid production type identifier - empty!");
@@ -121,7 +123,7 @@ bool ProductionTypeManager::add_production_type(
 		return false;
 	}
 
-	if (output_goods == nullptr) {
+	if (output_good == nullptr) {
 		Logger::error("Output good for production type ", identifier, " was null!");
 		return false;
 	}
@@ -167,13 +169,25 @@ bool ProductionTypeManager::add_production_type(
 	}
 
 	const bool ret = production_types.add_item({
-		identifier, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), output_goods,
+		identifier, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), *output_good,
 		base_output_quantity, std::move(bonuses), std::move(maintenance_requirements), is_coastal, is_farm, is_mine
 	});
+
+	if (ret && (template_type == RGO)) {
+		ProductionType const& production_type = production_types.get_items().back();
+		ProductionType const*& current_rgo_pt = good_to_rgo_production_type[*output_good];
+		if (current_rgo_pt == nullptr || (is_farm && !current_rgo_pt->is_farm())) {
+			// first rgo pt or farms are preferred (over mines) in V2
+			current_rgo_pt = &production_type;
+		}
+		//else ignore, we already have an rgo pt
+	}
+
 	if (rgo_owner_sprite <= 0 && ret && template_type == RGO && owner.has_value() && owner->get_pop_type() != nullptr) {
 		/* Set rgo owner sprite to that of the first RGO owner we find. */
 		rgo_owner_sprite = owner->get_pop_type()->get_sprite();
 	}
+
 	return ret;
 }
 
@@ -224,7 +238,10 @@ bool ProductionTypeManager::load_production_types_file(
 		}
 	)(parser.get_file_node());
 
+
 	/* Pass #3: actually load production types */
+	good_to_rgo_production_type.set_keys(&good_definition_manager.get_good_definitions());
+
 	reserve_more_production_types(expected_types);
 	ret &= expect_dictionary(
 		[this, &good_definition_manager, &pop_manager, &template_target_map, &template_node_map](
@@ -238,7 +255,7 @@ bool ProductionTypeManager::load_production_types_file(
 			std::optional<Job> owner;
 			std::vector<Job> jobs;
 			ProductionType::template_type_t template_type { FACTORY };
-			GoodDefinition const* output_goods = nullptr;
+			GoodDefinition const* output_good = nullptr;
 			Pop::pop_size_t base_workforce_size = 0;
 			GoodDefinition::good_definition_map_t input_goods, maintenance_requirements;
 			fixed_point_t base_output_quantity = 0;
@@ -271,7 +288,7 @@ bool ProductionTypeManager::load_production_types_file(
 				"input_goods", ZERO_OR_ONE,
 					good_definition_manager.expect_good_definition_decimal_map(move_variable_callback(input_goods)),
 				"output_goods", ZERO_OR_ONE,
-					good_definition_manager.expect_good_definition_identifier(assign_variable_callback_pointer(output_goods)),
+					good_definition_manager.expect_good_definition_identifier(assign_variable_callback_pointer(output_good)),
 				"value", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(base_output_quantity)),
 				"efficiency", ZERO_OR_ONE, good_definition_manager.expect_good_definition_decimal_map(
 					move_variable_callback(maintenance_requirements)
@@ -300,9 +317,10 @@ bool ProductionTypeManager::load_production_types_file(
 			ret &= parse_node(node);
 
 			ret &= add_production_type(
-				key, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), output_goods,
+				key, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), output_good,
 				base_output_quantity, std::move(bonuses), std::move(maintenance_requirements), is_coastal, is_farm, is_mine
 			);
+
 			return ret;
 		}
 	)(parser.get_file_node());
--- a/src/openvic-simulation/economy/production/ProductionType.hpp
+++ b/src/openvic-simulation/economy/production/ProductionType.hpp
@@ -20,13 +20,13 @@ namespace OpenVic {
 		PopType const* PROPERTY(pop_type);
 		effect_t PROPERTY(effect_type);
 		fixed_point_t PROPERTY(effect_multiplier);
-		fixed_point_t PROPERTY(desired_workforce_share);
+		fixed_point_t PROPERTY(amount);
 
 		Job(
 			PopType const* new_pop_type,
 			effect_t new_effect_type,
 			fixed_point_t new_effect_multiplier,
-			fixed_point_t new_desired_workforce_share
+			fixed_point_t new_amount
 		);
 
 	public:
@@ -47,7 +47,7 @@ namespace OpenVic {
 		const Pop::pop_size_t PROPERTY(base_workforce_size);
 
 		GoodDefinition::good_definition_map_t PROPERTY(input_goods);
-		GoodDefinition const* PROPERTY(output_goods);
+		GoodDefinition const& PROPERTY(output_good);
 		const fixed_point_t PROPERTY(base_output_quantity);
 		std::vector<bonus_t> PROPERTY(bonuses);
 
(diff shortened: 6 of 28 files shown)


wvpm:
I know I still have to squash everything...
I left the separate commits for now as they are linked in discord discussion.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -185,6 +201,51 @@ void ProvinceInstance::_update_pops(DefineManager const& define_manager) {
...
+	//TODO: if is_farm add farm size bonus
+	//TODO: if is_mine add mine size bonus
+	return ((total_worker_count_in_province / base_workforce_size).ceil() * fixed_point_t::_1_50()).floor();
+}
+
+void ProvinceInstance::convert_rgo_worker_pops_to_equivalent(ProductionType const& production_type) {
+	std::vector<Job> const& jobs = production_type.get_jobs();
+	fixed_point_t total_worker_count_in_province = 0; //not counting equivalents
+	for(Pop& pop : pops) {
+		bool is_worker_pop_type = false;
```
schombert, line 231: You never change the value of this variable, so the if condition below will never trigger. If this didn't show up in your warnings, you should probably turn it on.

wvpm: What do you mean?
`is_worker_pop_type` is set to true in `if(job.get_pop_type() == pop.get_type())`
`total_worker_count_in_province` is filled in
```cpp
if(is_worker_pop_type) {
	total_worker_count_in_province += pop.get_size();
}
```

wvpm: Right, that's a copy paste mistake from calculate rgo size. THanks for pointing it out. As for warnings, no clue who/what would warn here. Vs code barely helps with cpp programming.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,127 @@
...
 
 ResourceGatheringOperation::ResourceGatheringOperation(
-	ProductionType const& new_production_type,
+	ProductionType const* new_production_type,
 	fixed_point_t new_size_multiplier,
 	fixed_point_t new_revenue_yesterday,
 	fixed_point_t new_output_quantity_yesterday,
 	fixed_point_t new_unsold_quantity_yesterday,
-	ordered_map<Pop*, Pop::pop_size_t>&& new_employees
+	ordered_map<Pop const*, Pop::pop_size_t>&& new_employees
```
wvpm, line 18: I had to change this to const as province instance doesn't expose mutable pops.
We'll have to change province instance to expose them as they must be mutable, so the RGO can pay them.

wvpm: Turned out I have to use a different iterator:
`for (plf::colony<Pop>::iterator it = pops.begin(); it != pops.end(); ++it)`

src/openvic-simulation/map/ProvinceInstance.cpp
schombert: Warnings come from your compiler, not your IDE. For example, here is a selection of things you can turn on for clang, if that if your C++ compiler: https://clang.llvm.org/docs/DiagnosticsReference.html

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -35,6 +39,22 @@ ProvinceInstance::ProvinceInstance(
...
 
+GoodDefinition const* ProvinceInstance::get_rgo_good() const {
+	if(!rgo.is_valid()) { return nullptr; }
+	return &(rgo.get_production_type()->get_output_good());
+}
+void ProvinceInstance::set_rgo_production_type(ProductionType const* rgo_production_type_nullable) {
+	if(rgo_production_type_nullable != nullptr) {
+		ProductionType const& rgo_production_type = *rgo_production_type_nullable;
+		if(rgo_production_type.get_template_type() != ProductionType::template_type_t::RGO) {
+			//error
```
Hop311, line 50: Make set_rgo_production_type return a bool which is false if this error occurs and true otherwise and pass it up via whatever functions call this, and Logger::error a message here too.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -114,6 +114,16 @@ void Pop::setup_pop_test_values(IssueManager const& issue_manager) {
 	luxury_needs_fulfilled = test_range();
 }
 
+void Pop::convert_to_equivalent() {
+	PopType const* const equivalent = get_type()->get_equivalent();
+	if(equivalent == nullptr) {
+		//error?
```
Hop311, line 120: Make this function return a bool: false if this error occurs, true otherwise, and `Logger::error` a message. Pass the return value of this function up through the functions that call it.

src/openvic-simulation/map/State.cpp
```diff
@@ -6,6 +6,7 @@
 #include "openvic-simulation/map/ProvinceInstance.hpp"
 #include "openvic-simulation/map/Region.hpp"
 #include "openvic-simulation/utility/StringUtils.hpp"
+#include "types/fixed_point/FixedPoint.hpp"
```
Hop311, line 9:
```suggestion
#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
```

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -135,6 +140,9 @@ namespace OpenVic {
 		bool setup(BuildingTypeManager const& building_type_manager);
 		bool apply_history_to_province(ProvinceHistoryEntry const& entry, CountryInstanceManager& country_manager);
 
+		void initialise_for_new_game();
+
 		void setup_pop_test_values(IssueManager const& issue_manager);
+		std::vector<Pop*> get_mutable_pops();
```
Hop311, line 146: There's no need to create a temporary vector of pops, you can just add a mutable getter like so:
```
plf::colony<Pop>& ProvinceInstance::get_mutable_pops() {
	return pops;
}
```
and then iterate through it normally:
```
for (Pop& pop : province.get_mutable_pops()) {
	// ...
}
```

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
-	ProductionType const& new_production_type, fixed_point_t new_size_multiplier
-) : ResourceGatheringOperation { new_production_type, new_size_multiplier, 0, 0, 0, {} } {}
+ResourceGatheringOperation::ResourceGatheringOperation() : ResourceGatheringOperation {
+	nullptr, fixed_point_t::_0(),
+	 fixed_point_t::_0(), fixed_point_t::_0()
+	 , fixed_point_t::_0(), {}
+} {}
+
+void ResourceGatheringOperation::initialise_for_new_game(ProvinceInstance& location) {
+	if(this->production_type == nullptr) {
```
Hop311, line 41: No need for `this->`, you can just use `production_type` directly, same throughout the rest of this file.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
+		output_quantity_yesterday = 0;
+		revenue_yesterday = 0;
+		return;
+	}
+
+	ProductionType const& production_type = *(this->production_type);
```
Hop311, line 47: Best not to create another variable with the same name as the member variable `production_type` as it makes the code ambiguous, either give it a different name or just use the pointer directly since you already know it's not null. Same throughout the rest of this file.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
+		TerrainType const* const terrain_type = location.get_terrain_type();
+		if(terrain_type != nullptr) {
+			ModifierValue const& terrain_modifiers = terrain_type->get_modifier();
+			if(production_type.is_farm()) {
+				//base_workforce_size *= terrain_modifiers["farm_rgo_size"];
+			}
+			if(production_type.is_mine()) {
+				//base_workforce_size *= terrain_modifiers["mine_rgo_size"];
+			}
+		}
```
Hop311, line 85: When I've finished implementing the resultant modifier system you won't need to go via the terrain type, the province will have a `ModiferSum` variable which you can look up the farm or mine size modifier from.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
+		size_multiplier = ((total_worker_count_in_province / base_workforce_size).ceil() * fixed_point_t::_1_50()).floor();
+	}
+	return total_worker_count_in_province;
+}
+
+void ResourceGatheringOperation::hire(ProvinceInstance& location, Pop::pop_size_t available_worker_count) {
+	total_employees_count_cache = 0;
+	total_paid_employees_count_cache=0;
+	if(this->production_type == nullptr) {
+		employees.empty();
```
Hop311, line 95: This is a bool checking if the `plf::colony` is empty, if you want to make it empty use `employees.clear()`.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
+		PopType const& pop_type = *(pop.get_type());
+		for(Job const& job : jobs) {
+			if(job.get_pop_type() == &pop_type) {
+				is_worker_pop_type = true;
+				break;
+			}
+		}
+
+		if(is_worker_pop_type) {
+			const Pop::pop_size_t pop_size_to_hire = (proportion_to_hire * pop.get_size()).floor().to_int32_t();
```
Hop311, line 132:
```suggestion
			const Pop::pop_size_t pop_size_to_hire = static_cast<Pop::pop_size_t>((proportion_to_hire` * pop.get_size()).floor());
```
Better to cast to `Pop::pop_size_t` rather than hardcoding it to use `int32_t`, in case we change the pop size type in the future.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
+
+fixed_point_t ResourceGatheringOperation::produce(ProvinceInstance& location) {
+	if(this->production_type == nullptr) {
+		return fixed_point_t::_0();
+	}
+	
+	ProductionType const& production_type = *(this->production_type);
+	const fixed_point_t base_workforce_size = production_type.get_base_workforce_size();
+	//TODO include modifiers, job effects, etc
+	return production_type.get_base_output_quantity() * total_employees_count_cache / base_workforce_size;
```
Hop311, line 150: Check `base_workforce_size > 0`? Unless we already guaruntee that when loading production types.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
+		constexpr fixed_point_t upper_limit = fixed_point_t::_0_50();
+		if(owner_share > upper_limit) {
+			owner_share = upper_limit;
+		}
+
+		for(Pop* owner_pop_nullable : owner_pops) {
+			if(owner_pop_nullable == nullptr) {
+				Logger::error("owner_pop became null somehow after adding it to vector.");
+				return;
+			}
```
Hop311, line 214: This won't ever happen, it's just wasted CPU cycles to check here.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -14,8 +24,228 @@ ResourceGatheringOperation::ResourceGatheringOperation(
...
+		}
+		revenue_left *= (fixed_point_t::_1() - owner_share);
+	}
+
+	if(total_paid_employees_count_cache > 0) {
+		for (std::pair<Pop*, Pop::pop_size_t> employee_pair : employees) {
+			if(employee_pair.first == nullptr) {
+				Logger::error("employee became null somehow after adding it to ordered_map.");
+				return;
+			}
```
Hop311, line 229: Again, this is impossible so it's just wasting CPU cycles to check.

src/headless/main.cpp
```diff
@@ -72,9 +102,27 @@ static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests)
...
 			game_manager.get_instance_manager()->get_country_instance_manager();
 
-		print_ranking_list("Great Powers", country_instance_manager.get_great_powers());
+		std::vector<CountryInstance*> const& great_powers = country_instance_manager.get_great_powers();
+		print_ranking_list("Great Powers", great_powers);
 		print_ranking_list("Secondary Powers", country_instance_manager.get_secondary_powers());
 		print_ranking_list("All countries", country_instance_manager.get_total_ranking());
+
+		Logger::info("===== RGO test... =====");
+		for (int i = 0; i < 3; ++i) {
```
Hop311, line 111:
```suggestion
		for (int i = 0; i < std::min(3, great_powers.size()); ++i) {
```
Check against the number of great powers just in case there are fewer than 3, to avoid a crash.

src/headless/main.cpp
```diff
@@ -72,9 +102,27 @@ static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests)
...
+			CountryInstance const* const great_power = great_powers[i];
+			if(great_power == nullptr) {
+				Logger::warning("Great power ",std::to_string(i), " is null.");
```
Hop311, line 114: No need to use `std::to_string` inside `Logger` messages, `int`s are converted automatically.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,9 +1,19 @@
...
+#include <vector>
+
+#include "openvic-simulation/economy/production/ProductionType.hpp"
+#include "openvic-simulation/pop/Pop.hpp"
+#include "openvic-simulation/map/ProvinceInstance.hpp"
+#include "openvic-simulation/map/State.hpp"
+#include "openvic-simulation/map/TerrainType.hpp"
+#include "openvic-simulation/modifier/Modifier.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "utility/Logger.hpp"
```
Hop311, line 11:
```suggestion
#include "openvic-simulation/utility/Logger.hpp"
```

Hop311 (commented):
Great work, impressive to see you working over such a wide range of files, I'm glad to have others familiarising themselves with the SIM repo 😃.
Here are a bunch of comments on the PR so far, I'll aim to get the resultant modifier work done in the next few days so this can be done properly in a single PR

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -35,6 +38,18 @@ ProvinceInstance::ProvinceInstance(
 	religion_distribution {},
 	max_supported_regiments { 0 } {}
 
+GoodDefinition const* ProvinceInstance::get_rgo_good() const {
+	if(!rgo.is_valid()) { return nullptr; }
+	return &(rgo.get_production_type()->get_output_good());
+}
+void ProvinceInstance::set_rgo_production_type(ProductionType const& rgo_production_type) {
+	if(rgo_production_type.get_template_type() != ProductionType::template_type_t::RGO) {
+		//error
```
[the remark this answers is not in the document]
wvpm: Sure, it's a scenario that should never happen though.
When changing rgo via event, the event parser should catch this error instead of the rgo.
Same thing with start dates.
I'll make it return a bool and log error anyway (this code was before I realised we have a global logger).

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: Only terrain (& continent?) is used for rgo size calculation. Country modifiers, province modifiers, tech, issues, etc are not used iirc.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: base_workforce_size is only valid when > 0
I could add checks anyway, but as you pointed out, wasted cpu.
I only write such code when weird crashes make me paranoid.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: This is why I love is_<boolean> or has_<boolean> prefixes.
`empty()` is a verb and a function, that's confusing. But it's out of our hands due to library :/

src/openvic-simulation/map/ProvinceInstance.hpp
wvpm: Great, it also saves me the null checks :)

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -98,7 +102,16 @@ bool ProvinceHistoryMap::_load_history_entry(
...
 		"trade_goods", ZERO_OR_ONE,
-			good_definition_manager.expect_good_definition_identifier(assign_variable_callback_pointer_opt(entry.rgo)),
+			good_definition_manager.expect_good_definition_identifier_or_string(
+				[&definition_manager, &entry](GoodDefinition const& rgo_good) ->bool {
+					entry.rgo_production_type = definition_manager.get_economy_manager().get_production_type_manager().get_good_to_rgo_production_type()[rgo_good];
+					return entry.rgo_production_type != nullptr;
+					//we expect the good to have an rgo production type
+					//Victoria 2 treats this as null, but clearly the modder wanted there to be a good
+				},
+				allow_empty_true, //could be explicitly setting trade_goods to null
```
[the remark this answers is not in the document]
wvpm: Good point, I'll change it to optional to distinguish between no value specific in entry vs explicit null.

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -98,7 +102,16 @@ bool ProvinceHistoryMap::_load_history_entry(
...
+				do_warn //could be typo in good identifier
```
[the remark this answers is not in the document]
wvpm: If there is no trade_goods entry, okay.
If there is a trade_goods entry with invalid goods identifier, treat as null + warning.
If there is a trade_goods entry with valid goods identifier that doesn't have rgo production type, error.
If there is a trade_goods entry with valid goods identifier that does have rgo production type, okay + change rgo.

src/openvic-simulation/map/MapInstance.cpp
```diff
@@ -93,12 +94,14 @@ bool MapInstance::apply_history_to_provinces(
...
 			if (history_map != nullptr) {
 				ProvinceHistoryEntry const* pop_history_entry = nullptr;
+				ProductionType const* rgo_production_type = nullptr;
 
 				for (auto const& [entry_date, entry] : history_map->get_entries()) {
 					if (entry_date > date) {
 						break;
 					}
 
+					rgo_production_type = entry->get_rgo_production_type();
```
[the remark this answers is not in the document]
wvpm: I went with optional

src/headless/main.cpp
```diff
@@ -72,9 +112,24 @@ static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests)
...
 
-		print_ranking_list("Great Powers", country_instance_manager.get_great_powers());
+		std::vector<CountryInstance*> const& great_powers = country_instance_manager.get_great_powers();
+		print_ranking_list("Great Powers", great_powers);
 		print_ranking_list("Secondary Powers", country_instance_manager.get_secondary_powers());
 		print_ranking_list("All countries", country_instance_manager.get_total_ranking());
+
+		Logger::info("===== RGO test... =====");
+		for (int i = 0; i < std::min<int>(3, great_powers.size()); ++i) {
+			CountryInstance const& great_power = *(great_powers[i]);
```
Hop311, line 122:
```suggestion
			CountryInstance const& great_power = *great_powers[i];
```
Brackets not required as `var[...]` has higher precedence than `*var`

src/headless/main.cpp
```diff
@@ -72,9 +112,24 @@ static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests)
...
+			ProvinceInstance const* const capital_province = great_power.get_capital();
+			if(capital_province == nullptr) {
+				Logger::warning(great_power.get_identifier(), " has no capital ProvinceInstance set.");
+			}
+			else {
+				print_rgo(*capital_province);
+			}
+		}
+		ProvinceInstance *const& new_orleans = game_manager.get_instance_manager()->get_map_instance().get_province_instance_by_identifier("139");
```
Hop311, line 131:
```suggestion
		ProvinceInstance const* new_orleans = game_manager.get_instance_manager()->get_map_instance().get_province_instance_by_identifier("139");
```
Better to use a `const*`, and no need to make it a `const&` (the only reasons to take a reference to a pointer is if you want to change the value of a pointer variable somewhere else or have access to the value of a pointer somewhere else that might change, in all other cases it just adds an unnecessary extra level of indirection).
Also best to check if the `new_orleans` pointer is null before dereferencing it below, just so the program doesn't crash when tested with mods that don't have such a province (e.g. a test mod with the bare minimum define needed to run the game).

src/openvic-simulation/InstanceManager.cpp
```diff
@@ -163,6 +138,11 @@ bool InstanceManager::load_bookmark(Bookmark const* new_bookmark) {
 		map_instance, definition_manager.get_pop_manager().get_pop_types()
 	);
 
+	if(ret) {
```
Hop311, line 141:
```suggestion
	if (ret) {
```
I'm not sure if we should be `if`-gating the code below, even if there are some issues IMO it's better to try and press ahead. If we don't initialise the map for a new game then we're bascially deciding the game shouldn't proceed, whereas the approach in the rest of the codebase is to press ahead and do as much as we can. If we do decide to abort a game altogether then that decision should be made higher up in the code, from a place where we can half the program altogether and exit in a graceful manner.
If you do keep this `if` statement then it needs a space before the bracket

src/openvic-simulation/economy/production/Employee.hpp
```diff
@@ -0,0 +1,13 @@
...
+
+#include "openvic-simulation/pop/Pop.hpp"
+
+namespace OpenVic {
+	struct Employee {
+	private:
+		Pop::pop_size_t PROPERTY_RW(size);
+	public:
+		Pop& pop;
+		Employee(Pop& new_pop, Pop::pop_size_t const new_size);
```
Hop311, line 11: If you want a `const` for `new_size` then put it before the type rather than after (same in the C++ file), although `const` qualifiers for non-pointer or reference function arguments in declarations don't do anything, you could remove the `const` from the header or from the cpp file and the other one would still work.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -139,7 +139,7 @@ namespace OpenVic {
 		// TODO - population change over last 30 days
 		fixed_point_t PROPERTY(national_consciousness);
 		fixed_point_t PROPERTY(national_militancy);
-		IndexedMap<PopType, fixed_point_t> PROPERTY(pop_type_distribution);
+		IndexedMap<PopType, Pop::pop_size_t> PROPERTY(pop_type_distribution);
```
Hop311, line 142: The reason for this using `fixed_point_t` was partly because pie charts require their data to be `fixed_point`s. We'll need a corresponding main repo PR to add support for non-`fixed_point_t` value types for pie chart data.

src/openvic-simulation/economy/production/ProductionType.cpp
```diff
@@ -167,13 +169,25 @@ bool ProductionTypeManager::add_production_type(
...
+		identifier, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), *output_good,
 		base_output_quantity, std::move(bonuses), std::move(maintenance_requirements), is_coastal, is_farm, is_mine
 	});
+
+	if(ret && (template_type == RGO)) {
+		ProductionType const& production_type = production_types.get_items().back();
+		ProductionType const*& current_rgo_pt = good_to_rgo_production_type[*output_good];
+		if (current_rgo_pt == nullptr || (is_farm && !current_rgo_pt->is_farm())) {
+			// first rgo pt or farms are preferred (over mines) in V2
+			good_to_rgo_production_type[*output_good] = &production_type;
```
Hop311, line 181:
```suggestion
			current_rgo_pt = &production_type;
```
Use `current_rgo_pt` here, I assume that's why it's a reference

src/openvic-simulation/economy/production/ProductionType.hpp
```diff
@@ -83,6 +83,7 @@ namespace OpenVic {
 	private:
 		IdentifierRegistry<ProductionType> IDENTIFIER_REGISTRY(production_type);
 		PopType::sprite_t PROPERTY(rgo_owner_sprite);
+		IndexedMap<GoodDefinition, ProductionType const*> PROPERTY(good_to_rgo_production_type);
```
Hop311, line 86: IMO it's better to store this as a variable in each `GoodDefinition` rather than a map here, that way we can simply get the production type from the good rather than need access to the `ProductionTypeManager`

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,353 @@
 #include "ResourceGatheringOperation.hpp"
 
+#include <vector>
+
+#include "openvic-simulation/economy/production/Employee.hpp"
+#include "openvic-simulation/economy/production/ProductionType.hpp"
+#include "openvic-simulation/pop/Pop.hpp"
```
Hop311, line 7: This is out of alphabetical order

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,353 @@
...
+) : production_type_nullable { new_production_type_nullable },
 	revenue_yesterday { new_revenue_yesterday },
 	output_quantity_yesterday { new_output_quantity_yesterday },
 	unsold_quantity_yesterday { new_unsold_quantity_yesterday },
 	size_multiplier { new_size_multiplier },
-	employees { std::move(new_employees) } {}
+	employees { std::move(new_employees) },
+	max_employee_count_cache {},
+	total_employees_count_cache {},
+	total_paid_employees_count_cache { 0},
```
Hop311, line 32:
```suggestion
	max_employee_count_cache { 0 },
	total_employees_count_cache { 0 },
	total_paid_employees_count_cache { 0 },
```
The top two should be explicitly zero-initialised (they're primitive types so we don't want nonsense data under any circumstances, even if they'll be overwritten before being used) and the bottom one needs a space after the `0`

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,353 @@
...
+	set_production_type_nullable(production_type_nullable);
+}
 
-ResourceGatheringOperation::ResourceGatheringOperation(
-	ProductionType const& new_production_type, fixed_point_t new_size_multiplier
-) : ResourceGatheringOperation { new_production_type, new_size_multiplier, 0, 0, 0, {} } {}
+ResourceGatheringOperation::ResourceGatheringOperation(decltype(employee_count_per_type_cache)::keys_t const& pop_type_keys) : ResourceGatheringOperation {
+	nullptr, fixed_point_t::_0(),
+	 fixed_point_t::_0(), fixed_point_t::_0()
+	 , fixed_point_t::_0(), {}, pop_type_keys
```
Hop311, line 43: Comma on the wrong line

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,353 @@
...
+} {}
+
+//The location ensures only valid RGO production types are passed.
+void ResourceGatheringOperation::set_production_type_nullable(ProductionType const* new_production_type_nullable){
+	production_type_nullable = new_production_type_nullable;
+}
```
Hop311, line 49: Why not just make the definition of `production_type_nullable` use `PROPERTY_RW` (which generates a setter identical to this except it's `constexpr` so the function call can be optimised away during compilation) and get rid of this + its header declaration?

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -98,7 +102,16 @@ bool ProvinceHistoryMap::_load_history_entry(
 			expect_identifier(expect_mapped_string(colony_status_map, assign_variable_callback(entry.colonial))),
 		"is_slave", ZERO_OR_ONE, expect_bool(assign_variable_callback(entry.slave)),
 		"trade_goods", ZERO_OR_ONE,
-			good_definition_manager.expect_good_definition_identifier(assign_variable_callback_pointer_opt(entry.rgo)),
+			good_definition_manager.expect_good_definition_identifier_or_string(
+				[&definition_manager, &entry](GoodDefinition const& rgo_good) ->bool {
+					entry.rgo_production_type_nullable = definition_manager.get_economy_manager().get_production_type_manager().get_good_to_rgo_production_type()[rgo_good];
+					return entry.rgo_production_type_nullable != nullptr;
```
Hop311, line 108: There should be a `Logger::error` message if `entry.rgo_production_type_nullable` is `nullptr`, every single place where we return false during the loading process needs an associated log message explaining why

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -29,7 +33,7 @@ ProvinceInstance::ProvinceInstance(
 	event_modifiers {},
 	slave { false },
 	crime { nullptr },
-	rgo { nullptr },
+	rgo { { pop_type_keys} },
```
Hop311, line 36:
```suggestion
	rgo { pop_type_keys },
```
No need for the double `{}`, in fact that forces the creation of a temporary intermediate object which is then either move or copy constructed into the actual variable.

Hop311 (commented):
Left some initial comments, I still need to properly read through the RGO hiring amd production logic.
Also there are various formatting things I didn't leave specific comments for, e.g. spaces after `if` and `for` and around operators like `=`, indenting function arguments on a new line, not squashing `if` statements onto a single line, not going onto a newline for `else` statements (so `} else {` rather than `}\nelse {`), removing unnecessary brackets, etc. If you could clean up this stuff wherever you see it that would be great, but it's not the end of the world if some slips through, better to get the PR merged and think about formatting later

src/headless/main.cpp
wvpm: This line will just go as it was a leftover from adhoc rgo testing

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: Alphabet hard it seems

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: You're right. I thought I had to recalculate the rgo size multiplier or something but PROPERTY_RW is fine.

src/openvic-simulation/economy/production/ProductionType.cpp
wvpm: Great catch. I think I stopped halfway changing a const* const to a const*&.

src/openvic-simulation/economy/production/ProductionType.hpp
wvpm: That doesn't work as the identifier registry only exposes consts.
I'd have to extend the registry macro to expose mutable methods.

If that is the desired approach, it should be done as a follow up PR.

src/openvic-simulation/history/ProvinceHistory.cpp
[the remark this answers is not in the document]
Hop311: - `ZERO_OR_ONE` means no `trade_goods` entry is fine
- `allow_empty_true` also means `trade_goods = ""` produces no error and doesn't call the callback at all
- any non-empty string which isn't a valid good identifier won't call the callback either, and because `do_warn` is true this will produce a warning rather than an error
- a valid trade good with no production type will indeed produce an error, as the callback will be called and the `if` statement within it will be triggered
- a valid trade good with a non-null production type will work correctly, setting the rgo production type and not producing an error

So compared to what you state above, the only discrepancy is with the invalid good identifier case: an invalid trade good won't be treated as null, it will have no effect on `entry.rgo_production_type_nullable` at all (the callback can't be called unless there's a valid good to use for its argument). If the invalid trade good is due to an empty string then there'll be no warning, if it's from a non-empty string then there will be a warning. If you want a warning from empty strings too then replace `allow_empty_true` with a false value. If you want to force `entry.rgo_production_type_nullable` to be null when given an invalid trade good string (empty or not) then you'll need to use a regular `expect_identifier_or_string` and try find a `GoodDefinition` with that identifier in the callback, either prodeeding as normal if you find it or manually setting `entry.rgo_production_type_nullable` to `nullptr` if you don't.

src/openvic-simulation/economy/production/ProductionType.hpp
```diff
@@ -97,19 +98,19 @@ namespace OpenVic {
...
-			std::vector<Job>&& employees,
-			ProductionType::template_type_t template_type,
-			Pop::pop_size_t workforce,
-			GoodDefinition::good_definition_map_t&& input_goods,
-			GoodDefinition const* output_goods,
-			fixed_point_t value,
+			std::vector<Job>&& jobs,
+			const ProductionType::template_type_t template_type,
+			const Pop::pop_size_t base_workforce_size,
+			GoodDefinition::good_definition_map_t const&& input_goods,
```
Hop311, line 106: All instances of `const&&` in this file and in ProductionType.cpp should be changed to just `&&` - this represents an r-value reference, or a "moved value". When moving something like a vector or a map, you "steal" its internal memory pointer and replace it will null so that when the original value is destroyed it won't free that memory. You can't do that if the value is const, as its internal pointer can't be changed to null, forcing you to make a copy instead and defeating the whole purpose of moving the object.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,343 @@
...
+		output_quantity_yesterday = 0;
+		revenue_yesterday = 0;
+		return;
+	}
+
+	ProductionType const& production_type = *production_type_nullable;
+	const fixed_point_t size_modifier = calculate_size_modifier(location, modifier_effect_cache);
+	const Pop::pop_size_t total_worker_count_in_province = update_size_and_return_total_worker_count(location, modifier_effect_cache, size_modifier);
+	hire(location, total_worker_count_in_province);
+	Pop::pop_size_t total_owner_count_in_state_cache {};
```
Hop311, line 55:
```suggestion
	Pop::pop_size_t total_owner_count_in_state_cache = 0;
```
Better to explicitly initialise this to `0` as it's really just a primitive integer type behind the alias, both for clarity and absolute certainty that it's correctly initialised. Things like vectors and fixed points are structs under the hood, so they have constructors that we can be sure trigger upon `{}` initialisation.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,343 @@
...
+	const fixed_point_t size_modifier
+) {
+	if (production_type_nullable == nullptr) {
+		size_multiplier = fixed_point_t::_0();
+		max_employee_count_cache = fixed_point_t::_0();
+		return fixed_point_t::_0();
+	}
+	
+	Pop::pop_size_t total_worker_count_in_province = 0; //not counting equivalents
+	ProductionType const& production_type = *this->production_type_nullable;
```
Hop311, line 74:
```suggestion
	ProductionType const& production_type = *production_type_nullable;
```
It's never necessary to use `this` to access member variables within a struct's member function. The only potential case is when there are variable name conflicts, but the solution to that is to rename the variables rather than using `this` unnecessarily.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,343 @@
...
+	for (PopType const& pop_type : *employee_count_per_type_cache.get_keys()) {
+		Pop::pop_size_t const& employees_of_type = employee_count_per_type_cache[pop_type];
+		
+		for(Job const& job : production_type.get_jobs()) {
+			if (job.get_pop_type() != &pop_type) {
+				continue;
+			}
+
+			fixed_point_t const& effect_multiplier = job.get_effect_multiplier();
+			fixed_point_t relative_to_workforce = employees_of_type * fixed_point_t::_1() / max_employee_count_cache;
```
Hop311, line 257: Is the `* 1` necessary? Can't you just do `employees_of_type  / max_employee_count_cache`, maybe casting `employees_of_type` to fixed point if necessary?

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -1,21 +1,343 @@
...
+			fixed_point_t const& amount = job.get_amount();
```
Hop311, line 258: No need for `const&` here, `get_amount()` will return a copy of the value in the job rather than a reference to it.

src/openvic-simulation/economy/production/ResourceGatheringOperation.hpp
```diff
@@ -1,25 +1,64 @@
...
+			Pop::pop_size_t& total_owner_count_in_state_cache,
+			ModifierEffectCache const& modifier_effect_cache,
+			const fixed_point_t size_modifier
+		);
+		void pay_employees(
+			ProvinceInstance& location,
+			const fixed_point_t revenue,
+			const Pop::pop_size_t total_worker_count_in_province,
+			std::vector<Pop*>& owner_pops_cache,
+			Pop::pop_size_t const& total_owner_count_in_state_cache
```
Hop311, line 45: `Pop::pop_size_t` is a primitive integer type which is smaller than a pointer value, so it's far cheaper to just use it by value rather than using a `const&` reference to it.

Hop311 (commented):
Left some comments, also this needs to be rebased onto master before it can be merged

src/openvic-simulation/economy/production/ProductionType.hpp
wvpm: Good advice, thanks.
I still struggle with const or no const and & vs * vs nothing.
In my mind, I think of them as functional aspects of a variable.
So in my mind const means "read only".
& I think of as "guaranteed instance".
* I treat as nullable.

This way of thinking helps me in some cases but fails to properly optimise code. I'll change it to just &&.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: Sorry, nasty javascript habbit I picked up.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: A cast is better yeah. This is a common C# practice as C# has suffixes for real number types (1M, 1D or 1F).

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: Right. Same thing for effect multiplier I think. Since they are just numbers.

Hop311 (approved):

src/openvic-simulation/history/ProvinceHistory.cpp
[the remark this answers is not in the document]
wvpm: It works as intended then. If someone wants to use an invalid non-empty identifier for null and we somehow think that is fine, we can always change it later.

PR #208, opened 2024-10-20
"defence leader", no, it's "defence"

Nivaturimika:


Hop311:
We split certain modifier effects which work differently in different places in separate variant effects, e.g. `combat_width` is an additive integer when found in tech and inventions but it's a decimal/fractional percentage change when found in terrain modifiers.
For `defence` we need to differentiate between leader (add to defence of units in army) and terrain (subtract from attacking units' defence in battles in that province) variants, which we do based on the `modifier_type_t` provided to the modifier parsing function (in this case `LEADER` which causes `defence` to be interpreted as the `defence leader` modifier effect). The modifier whitelist contains identifiers of the modifier effects that can be parsed (after variant substitution), not necessarily the same as the raw text that appears in the defines file's text.
If you try running the headless sim executable with this change you'll get a lot this error (because the whitelist will only allow `defence`, not the `defence leader` variant):
```
[ERROR] ModifierManager.cpp(709): Failed to validate modifier effect: defence leader
```

PR #210, opened 2024-10-22
Contextual modifier parsing

wvpm:
Parse modifier effects based on the context in which they are defined. The context dictates which keys are valid and how to map them to functional modifier effects.

The contexts are:
- leader (isolated)
- unit_terrain (isolated)
- shared_tech_country (fallback for all other modifiers)
- technology (falls back to shared_tech_country)
- base_country (falls back to shared_tech_country)
- base_province (falls back to base_country)
- province_event (falls back to base_province (also used for country event modifiers, using the fallbacks))
- terrain (falls back to base_province)

See also https://discord.com/channels/1063392556160909312/1063416834650554398/1297862256947892265

### DIFF
--- a/src/openvic-simulation/map/Crime.cpp
+++ b/src/openvic-simulation/map/Crime.cpp
@@ -1,5 +1,6 @@
 #include "Crime.hpp"
 
+#include "openvic-simulation/dataloader/NodeTools.hpp"
 #include "openvic-simulation/modifier/ModifierManager.hpp"
 
 using namespace OpenVic;
@@ -29,16 +30,13 @@ bool CrimeManager::load_crime_modifiers(ModifierManager const& modifier_manager,
 	const bool ret = expect_dictionary_reserve_length(
 		crime_modifiers,
 		[this, &modifier_manager](std::string_view key, ast::NodeCPtr value) -> bool {
-			using enum Modifier::modifier_type_t;
-
 			ModifierValue modifier_value;
 			IconModifier::icon_t icon = 0;
 			ConditionScript trigger { scope_t::PROVINCE, scope_t::NO_SCOPE, scope_t::NO_SCOPE };
 			bool default_active = false;
 
-			bool ret = modifier_manager.expect_modifier_value_and_keys(
-				move_variable_callback(modifier_value),
-				CRIME,
+			bool ret = NodeTools::expect_dictionary_keys_and_default(
+				modifier_manager.expect_base_province_modifier(modifier_value),
 				"icon", ZERO_OR_ONE, expect_uint(assign_variable_callback(icon)),
 				"trigger", ONE_EXACTLY, trigger.expect_script(),
 				"active", ZERO_OR_ONE, expect_bool(assign_variable_callback(default_active))
--- a/src/openvic-simulation/map/MapDefinition.cpp
+++ b/src/openvic-simulation/map/MapDefinition.cpp
@@ -3,6 +3,7 @@
 #include <cstdint>
 #include <vector>
 
+#include "openvic-simulation/dataloader/NodeTools.hpp"
 #include "openvic-simulation/modifier/ModifierManager.hpp"
 #include "openvic-simulation/types/Colour.hpp"
 #include "openvic-simulation/types/OrderedContainers.hpp"
@@ -955,11 +956,11 @@ bool MapDefinition::load_climate_file(ModifierManager const& modifier_manager, a
 			bool ret = true;
 			Climate* cur_climate = climates.get_item_by_identifier(identifier);
 			if (cur_climate == nullptr) {
-				using enum Modifier::modifier_type_t;
-
 				ModifierValue values;
 
-				ret &= modifier_manager.expect_modifier_value(move_variable_callback(values), CLIMATE)(node);
+				ret &= NodeTools::expect_dictionary(
+					modifier_manager.expect_base_province_modifier(values)
+				)(node);
 
 				ret &= climates.add_item({ identifier, std::move(values), Modifier::modifier_type_t::CLIMATE });
 			} else {
@@ -1004,7 +1005,6 @@ bool MapDefinition::load_continent_file(ModifierManager const& modifier_manager,
 	bool ret = expect_dictionary_reserve_length(
 		continents,
 		[this, &modifier_manager](std::string_view identifier, ast::NodeCPtr node) -> bool {
-			using enum Modifier::modifier_type_t;
 
 			if (identifier.empty()) {
 				Logger::error("Invalid continent identifier - empty!");
@@ -1013,9 +1013,8 @@ bool MapDefinition::load_continent_file(ModifierManager const& modifier_manager,
 
 			ModifierValue values;
 			std::vector<ProvinceDefinition const*> prov_list;
-			bool ret = modifier_manager.expect_modifier_value_and_keys(
-				move_variable_callback(values),
-				CONTINENT,
+			bool ret = NodeTools::expect_dictionary_keys_and_default(
+				modifier_manager.expect_base_province_modifier(values),
 				"provinces", ONE_EXACTLY, expect_list_reserve_length(prov_list, expect_province_definition_identifier(
 					[&prov_list](ProvinceDefinition const& province) -> bool {
 						if (province.continent == nullptr) {
--- a/src/openvic-simulation/map/State.cpp
+++ b/src/openvic-simulation/map/State.cpp
@@ -5,6 +5,7 @@
 #include "openvic-simulation/map/MapInstance.hpp"
 #include "openvic-simulation/map/ProvinceInstance.hpp"
 #include "openvic-simulation/map/Region.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
 #include "openvic-simulation/utility/StringUtils.hpp"
 
 using namespace OpenVic;
@@ -65,10 +66,16 @@ void State::update_gamestate() {
 	const int32_t potential_workforce_in_state = 0; // sum of worker pops, regardless of employment
 	const int32_t potential_employment_in_state = 0; // sum of (factory level * production method base_workforce_size)
 
-	industrial_power = total_factory_levels_in_state * std::clamp(
-		(fixed_point_t { potential_workforce_in_state } / 100).floor() * 400 / potential_employment_in_state,
-		fixed_point_t::_0_20(), fixed_point_t::_4()
-	);
+	fixed_point_t unclamped_score_per_factory = fixed_point_t::_0();
+
+	if (potential_employment_in_state > 0) {
+		unclamped_score_per_factory = (fixed_point_t { potential_workforce_in_state } / fixed_point_t::_100()).floor() * 400 / potential_employment_in_state;
+	} else {
+		industrial_power = total_factory_levels_in_state * std::clamp(
+			unclamped_score_per_factory,
+			fixed_point_t::_0_20(), fixed_point_t::_4()
+		);
+	}
 }
 
 /* Whether two provinces in the same region should be grouped into the same state or not.
--- a/src/openvic-simulation/map/TerrainType.cpp
+++ b/src/openvic-simulation/map/TerrainType.cpp
@@ -1,5 +1,7 @@
 #include "TerrainType.hpp"
 
+#include <string_view>
+
 #include "openvic-simulation/modifier/ModifierManager.hpp"
 #include "openvic-simulation/types/Colour.hpp"
 
@@ -16,6 +18,38 @@ TerrainTypeMapping::TerrainTypeMapping(
 	index_t new_priority, bool new_has_texture
 ) : HasIdentifier { new_identifier }, type { new_type }, terrain_indices { std::move(new_terrain_indicies) },
 	priority { new_priority }, has_texture { new_has_texture } {}
+
+bool TerrainTypeManager::generate_modifiers(ModifierManager& modifier_manager) const {
+	using enum ModifierEffect::format_t;
+	IndexedMap<TerrainType, ModifierEffectCache::unit_terrain_effects_t>& unit_terrain_effects =
+		modifier_manager.modifier_effect_cache.unit_terrain_effects;
+
+	unit_terrain_effects.set_keys(&get_terrain_types());
+
+	constexpr bool has_no_effect = true;
+	bool ret = true;
+	for (TerrainType const& terrain_type : get_terrain_types()) {
+		const std::string_view identifier = terrain_type.get_identifier();
+		ModifierEffectCache::unit_terrain_effects_t& this_unit_terrain_effects = unit_terrain_effects[terrain_type];
+		ret &= modifier_manager.register_unit_terrain_modifier_effect(
+			this_unit_terrain_effects.attack, ModifierManager::get_flat_identifier("attack", identifier), true,
+			PROPORTION_DECIMAL, "UA_ATTACK", has_no_effect
+		);
+		ret &= modifier_manager.register_unit_terrain_modifier_effect(
+			this_unit_terrain_effects.defence, ModifierManager::get_flat_identifier("defence", identifier), true,
+			PROPORTION_DECIMAL, "UA_DEFENCE", has_no_effect
+		);
+		ret &= modifier_manager.register_unit_terrain_modifier_effect(
+			this_unit_terrain_effects.attrition, ModifierManager::get_flat_identifier("attrition", identifier), false,
+			RAW_DECIMAL, "UA_ATTRITION", has_no_effect
+		);
+		ret &= modifier_manager.register_unit_terrain_modifier_effect(
+			this_unit_terrain_effects.movement, ModifierManager::get_flat_identifier("movement", identifier), true,
+			PROPORTION_DECIMAL, "UA_MOVEMENT"
+		);
+	}
+	return ret;
+}
 
 bool TerrainTypeManager::add_terrain_type(
 	std::string_view identifier, colour_t colour, ModifierValue&& values, bool is_water
@@ -72,15 +106,12 @@ node_callback_t TerrainTypeManager::_load_terrain_type_categories(ModifierManage
 	return [this, &modifier_manager](ast::NodeCPtr root) -> bool {
 		const bool ret = expect_dictionary_reserve_length(terrain_types,
 			[this, &modifier_manager](std::string_view type_key, ast::NodeCPtr type_node) -> bool {
-				using enum Modifier::modifier_type_t;
-
 				ModifierValue values;
 				colour_t colour = colour_t::null();
 				bool is_water = false;
 
-				bool ret = modifier_manager.expect_modifier_value_and_keys(
-					move_variable_callback(values),
-					TERRAIN,
+				bool ret = NodeTools::expect_dictionary_keys_and_default(
+					modifier_manager.expect_terrain_modifier(values),
 					"color", ONE_EXACTLY, expect_colour(assign_variable_callback(colour)),
 					"is_water", ZERO_OR_ONE, expect_bool(assign_variable_callback(is_water))
 				)(type_node);
--- a/src/openvic-simulation/military/LeaderTrait.cpp
+++ b/src/openvic-simulation/military/LeaderTrait.cpp
@@ -1,5 +1,6 @@
 #include "LeaderTrait.hpp"
 
+#include "openvic-simulation/dataloader/NodeTools.hpp"
 #include "openvic-simulation/modifier/ModifierManager.hpp"
 
 using namespace OpenVic;
@@ -32,17 +33,10 @@ bool LeaderTraitManager::load_leader_traits_file(ModifierManager const& modifier
 		return expect_dictionary_reserve_length(
 			leader_traits,
 			[this, &modifier_manager, type](std::string_view trait_identifier, ast::NodeCPtr value) -> bool {
-				using enum Modifier::modifier_type_t;
-
-				static const string_set_t allowed_modifiers = {
-					"attack", "defence leader", "morale", "organisation", "reconnaissance",
-					"speed", "attrition", "experience", "reliability"
-				};
-
 				ModifierValue modifiers;
 
-				bool ret = modifier_manager.expect_whitelisted_modifier_value(
-					move_variable_callback(modifiers), LEADER, allowed_modifiers
+				bool ret = NodeTools::expect_dictionary(
+					modifier_manager.expect_leader_modifier(modifiers)
 				)(value);
 
 				ret &= add_leader_trait(trait_identifier, type, std::move(modifiers));
--- a/src/openvic-simulation/military/UnitType.cpp
+++ b/src/openvic-simulation/military/UnitType.cpp
@@ -1,6 +1,7 @@
 #include "UnitType.hpp"
 
 #include "openvic-simulation/country/CountryInstance.hpp"
+#include "openvic-simulation/dataloader/NodeTools.hpp"
 #include "openvic-simulation/map/TerrainType.hpp"
 #include "openvic-simulation/modifier/ModifierManager.hpp"
 
@@ -247,11 +248,9 @@ bool UnitTypeManager::load_unit_type_file(
 			TerrainType const* terrain_type = terrain_type_manager.get_terrain_type_by_identifier(default_key);
 
 			if (terrain_type != nullptr) {
-				using enum Modifier::modifier_type_t;
-
-				// TODO - restrict what modifier effects can be used here
-				return modifier_manager.expect_modifier_value(
-					map_callback(unit_args.terrain_modifier_values, terrain_type), UNIT_TERRAIN
+				ModifierValue& modifier_value = unit_args.terrain_modifier_values[terrain_type];
+				return expect_dictionary(
+					modifier_manager.expect_unit_terrain_modifier(modifier_value, terrain_type->get_identifier())
 				)(default_value);
 			}
 
@@ -336,14 +335,13 @@ bool UnitTypeManager::generate_modifiers(ModifierManager& modifier_manager) cons
 		std::derived_from<ModifierEffectCache::unit_type_effects_t> auto unit_type_effects, std::string_view identifier
 	) -> void {
 		using enum ModifierEffect::format_t;
-		using enum ModifierEffect::target_t;
 
 		const auto stat_modifier = [&modifier_manager, &ret, &identifier](
 			ModifierEffect const*& effect_cache, std::string_view suffix, bool is_positive_good,
 			ModifierEffect::format_t format, std::string_view localisation_key
 		) -> void {
-			ret &= modifier_manager.add_modifier_effect(
-				effect_cache, ModifierManager::get_flat_identifier(identifier, suffix), is_positive_good, format, COUNTRY,
+			ret &= modifier_manager.register_technology_modifier_effect(
+				effect_cache, ModifierManager::get_flat_identifier(identifier, suffix), is_positive_good, format,
 				StringUtils::append_string_views("$", identifier, "$: $", localisation_key, "$")
 			);
 		};
--- a/src/openvic-simulation/modifier/Modifier.hpp
+++ b/src/openvic-simulation/modifier/Modifier.hpp
@@ -12,6 +12,7 @@ namespace OpenVic {
 
 	struct Modifier : HasIdentifier, ModifierValue {
 		friend struct ModifierManager;
+		friend struct StaticModifierCache;
 		friend struct UnitType;
 
 		enum struct modifier_type_t : uint8_t {
--- a/src/openvic-simulation/modifier/ModifierEffectCache.cpp
+++ b/src/openvic-simulation/modifier/ModifierEffectCache.cpp
@@ -2,6 +2,7 @@
 
 #include "openvic-simulation/economy/BuildingType.hpp"
 #include "openvic-simulation/economy/GoodDefinition.hpp"
+#include "openvic-simulation/map/TerrainType.hpp"
 #include "openvic-simulation/politics/Rebel.hpp"
 #include "openvic-simulation/research/Technology.hpp"
 
@@ -47,6 +48,12 @@ ModifierEffectCache::ship_type_effects_t::ship_type_effects_t()
 	fire_range { nullptr },
 	evasion { nullptr },
 	torpedo_attack { nullptr } {}
+
+ModifierEffectCache::unit_terrain_effects_t::unit_terrain_effects_t()
+  : attack  { nullptr },
+	defence  { nullptr },
+	attrition { nullptr },
+	movement { nullptr } {}
 
 ModifierEffectCache::strata_effects_t::strata_effects_t()
   : income_modifier { nullptr },
@@ -113,7 +120,8 @@ ModifierEffectCache::ModifierEffectCache()
 	leadership { nullptr },
 	leadership_modifier { nullptr },
 	literacy_con_impact { nullptr },
-	loan_interest { nullptr },
+	loan_interest_base { nullptr },
+	loan_interest_foreign { nullptr },
 	max_loan_modifier { nullptr },
 	max_military_spending { nullptr },
 	max_national_focus { nullptr },
@@ -170,8 +178,11 @@ ModifierEffectCache::ModifierEffectCache()
 	boost_strongest_party { nullptr },
 	combat_width_percentage_change { nullptr },
 	defence_terrain { nullptr },
-	farm_rgo_eff { nullptr },
-	farm_rgo_size { nullptr },
+	farm_rgo_throughput_global { nullptr },
+	farm_rgo_output_global { nullptr },
+	farm_rgo_output_local { nullptr },
+	farm_rgo_size_global { nullptr },
+	farm_rgo_size_local { nullptr },
 	immigrant_attract { nullptr },
 	immigrant_push { nullptr },
 	life_rating { nullptr },
@@ -187,22 +198,28 @@ ModifierEffectCache::ModifierEffectCache()
 	local_ruling_party_support { nullptr },
 	local_ship_build { nullptr },
 	max_attrition { nullptr },
-	mine_rgo_eff { nullptr },
-	mine_rgo_size { nullptr },
+	mine_rgo_throughput_global { nullptr },
+	mine_rgo_output_global { nullptr },
+	mine_rgo_output_local { nullptr },
+	mine_rgo_size_global { nullptr },
+	mine_rgo_size_local { nullptr },
 	movement_cost_base { nullptr },
 	movement_cost_percentage_change { nullptr },
+	movement_cost_percentage_change_fake { nullptr },
 	number_of_voters { nullptr },
 	pop_consciousness_modifier { nullptr },
 	pop_militancy_modifier { nullptr },
 	population_growth { nullptr },
-	supply_limit { nullptr },
+	supply_limit_global_percentage_change { nullptr },
+	supply_limit_local_base { nullptr },
 
 	/* Military Modifier Effects */
-	attack { nullptr },
+	attack_leader { nullptr },
 	attrition { nullptr },
 	defence_leader { nullptr },
 	experience { nullptr },
-	morale { nullptr },
+	morale_global { nullptr },
+	morale_leader { nullptr },
 	organisation { nullptr },
 	reconnaissance { nullptr },
 	reliability { nullptr },
@@ -219,6 +236,7 @@ ModifierEffectCache::ModifierEffectCache()
 	regiment_type_effects { nullptr },
 	navy_base_effects {},
 	ship_type_effects { nullptr },
+	unit_terrain_effects { nullptr },
 
 	/* Rebel Effects */
 	rebel_org_gain_all { nullptr },
--- a/src/openvic-simulation/modifier/ModifierManager.cpp
+++ b/src/openvic-simulation/modifier/ModifierManager.cpp
@@ -1,11 +1,35 @@
 #include "ModifierManager.hpp"
 
+#include <utility>
+
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "openvic-simulation/modifier/Modifier.hpp"
+#include "openvic-simulation/modifier/ModifierEffect.hpp"
+
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
+using namespace OpenVic::StringUtils;
 
-bool ModifierManager::add_modifier_effect(
-	ModifierEffect const*& effect_cache, std::string_view identifier, bool positive_good, ModifierEffect::format_t format,
-	ModifierEffect::target_t targets, std::string_view localisation_key
+using enum ModifierEffect::target_t;
+
+void ModifierManager::lock_all_modifier_except_base_country_effects() {
+	lock_leader_modifier_effects();
+	lock_unit_terrain_modifier_effects();
+	lock_shared_tech_country_modifier_effects();
+	lock_technology_modifier_effects();
+	lock_base_province_modifier_effects();
+	lock_terrain_modifier_effects();
+}
+
+bool ModifierManager::_register_modifier_effect(
+	modifier_effect_registry_t& registry,
+	ModifierEffect::target_t targets,
+	ModifierEffect const*& effect_cache,
+	const std::string_view identifier,
+	const bool is_positive_good,
+	const ModifierEffect::format_t format,
+	const std::string_view localisation_key,
+	const bool has_no_effect
 ) {
 	using enum ModifierEffect::target_t;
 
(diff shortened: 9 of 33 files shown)


src/openvic-simulation/map/State.cpp
wvpm, line 1: Unrelated change but required to fix divide by zero.
If desired, I can make a separate PR for this.

src/openvic-simulation/military/UnitType.cpp
```diff
@@ -247,12 +247,12 @@ bool UnitTypeManager::load_unit_type_file(
 			TerrainType const* terrain_type = terrain_type_manager.get_terrain_type_by_identifier(default_key);
 
 			if (terrain_type != nullptr) {
-				using enum Modifier::modifier_type_t;
-
-				// TODO - restrict what modifier effects can be used here
-				return modifier_manager.expect_modifier_value(
-					map_callback(unit_args.terrain_modifier_values, terrain_type), UNIT_TERRAIN
-				)(default_value);
+				//TODO WVPM
```
wvpm, line 250: I need some examples for the UNIT_TERRAIN to properly implement this.

src/openvic-simulation/modifier/ModifierEffectCache.cpp
wvpm, line 1: There were several modifier effects that have different meanings based on the context. I hope their new names are clear.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -514,74 +586,27 @@ std::string ModifierManager::get_flat_identifier(
...
 	}
 
 	return event_modifiers.add_item(
-		{ identifier, std::move(values), EVENT, icon }, duplicate_warning_callback
+		{ identifier, std::move(values), origin, icon }, duplicate_warning_callback
 	);
 }
 
-bool ModifierManager::load_event_modifiers(ast::NodeCPtr root) {
+bool ModifierManager::load_event_modifiers(ast::NodeCPtr const& root) {
```
Hop311, line 601: These shouldn't be `const&`, `ast::NodeCPtr` is an alias of a pointer type so a reference to it would be equivalent to a pointer to a pointer when we could just pass the first pointer by value.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -1,11 +1,32 @@
...
+	lock_terrain_modifier_effects();
+}
 
-bool ModifierManager::add_modifier_effect(
-	ModifierEffect const*& effect_cache, std::string_view identifier, bool positive_good, ModifierEffect::format_t format,
-	ModifierEffect::target_t targets, std::string_view localisation_key
+bool ModifierManager::_register_modifier_effect(
+	CaseInsensitiveIdentifierRegistry<ModifierEffect, RegistryStorageInfoDeque>& registry,
+	ModifierEffect::target_t targets,
+	ModifierEffect const*& effect_cache, std::string_view identifier,
```
Hop311, line 25:
```suggestion
	ModifierEffect const*& effect_cache,
	std::string_view identifier,
```
Make the argument formatting consistent

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -35,466 +56,517 @@ bool ModifierManager::add_modifier_effect(
...
+	bool has_no_effect
+) { return _register_modifier_effect(base_province_modifier_effects, ModifierEffect::target_t::PROVINCE, effect_cache, std::move(identifier), is_positive_good, format, localisation_key, has_no_effect); }
+bool ModifierManager::register_terrain_modifier_effect(
+	ModifierEffect const*& effect_cache,
+	std::string_view identifier,
+	bool is_positive_good,
+	ModifierEffect::format_t format,
+	std::string_view localisation_key,
+	bool has_no_effect
+) { return _register_modifier_effect(terrain_modifier_effects, ModifierEffect::target_t::PROVINCE, effect_cache, std::move(identifier), is_positive_good, format, localisation_key, has_no_effect); }
```
Hop311, line 123: Formatting + unnecessary code duplication, use a macro instead (and undefine it at the end).

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -1,11 +1,32 @@
...
+#include <utility>
+#include "dataloader/NodeTools.hpp"
+#include "modifier/Modifier.hpp"
+#include "modifier/ModifierEffect.hpp"
+
+#define FWD(...) static_cast<decltype(__VA_ARGS__)>(__VA_ARGS__)
 
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
+using namespace OpenVic::StringUtils;
```
Hop311, line 11: Remove this, there's nothing in this file that's dependent on it (and there are so few places using `StringUtils:...` that it's better to be explicit every time)

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -35,466 +56,517 @@ bool ModifierManager::add_modifier_effect(
...
-	ret &= add_modifier_effect(
-		modifier_effect_cache.artisan_throughput, "artisan_throughput", true, PROPORTION_DECIMAL, COUNTRY
+		PROPORTION_DECIMAL, ModifierEffect::make_default_modifier_effect_localisation_key("administrative_efficiency")
 	);
-	ret &= add_modifier_effect(modifier_effect_cache.badboy, "badboy", false, RAW_DECIMAL, COUNTRY);
-	ret &= add_modifier_effect(
-		modifier_effect_cache.cb_generation_speed_modifier, "cb_generation_speed_modifier", true, PROPORTION_DECIMAL, COUNTRY
+	ret &= register_technology_modifier_effect(modifier_effect_cache.artisan_input, "artisan_input", false, PROPORTION_DECIMAL,{},has_no_effect);
+	ret &= register_technology_modifier_effect(modifier_effect_cache.artisan_output, "artisan_output", true, PROPORTION_DECIMAL,{},has_no_effect);
+	ret &= register_technology_modifier_effect(	modifier_effect_cache.artisan_throughput, "artisan_throughput", true, PROPORTION_DECIMAL,{},has_no_effect);
```
Hop311, line 175:
```suggestion
	ret &= register_technology_modifier_effect(modifier_effect_cache.artisan_throughput, "artisan_throughput", true, PROPORTION_DECIMAL, {}, has_no_effect);
```
Throughout this function: remove stray tabs, add missing spaces, and split any lines `>= 128` column widths wide.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +674,105 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
 	return ret;
 }
 
-key_value_callback_t ModifierManager::_modifier_effect_callback(
-	ModifierValue& modifier, Modifier::modifier_type_t type, key_value_callback_t default_callback,
-	ModifierEffectValidator auto effect_validator
+bool ModifierManager::_add_flattened_modifier_cb(
+	ModifierValue& modifier_value,
+	std::string_view const& prefix, std::string_view const& key, ast::NodeCPtr const& value
```
Hop311, line 679: None of these should be `const&`, all of them are small enough that it's cheaper to pass by value.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +674,105 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
+	ModifierValue& modifier_value
+) const {
+	return _expect_modifier_effect_with_fallback(registry, modifier_value, key_value_invalid_callback);
+}
+NodeTools::key_value_callback_t ModifierManager::_expect_modifier_effect_with_fallback(
+	CaseInsensitiveIdentifierRegistry<ModifierEffect, RegistryStorageInfoDeque> const& registry,
+	ModifierValue& modifier_value,
+	NodeTools::key_value_callback_t const& fallback
+) const {
+	return [this, &registry, &modifier_value, fallback = FWD(fallback)](auto const& key, auto const& value) {
```
Hop311, line 709: I'm pretty sure you wouldn't need the `FWD` if `fallback` wasn't `const&`. And the lambda functions aruguments should be `std::string_view key, ast::NodeCPtr value`, not `auto const&`.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +674,105 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
+		if (dryad::node_has_kind<ast::ListValue>(value) && complex_modifiers.contains(key)) {
+			const key_value_callback_t callback = [this, &modifier_value, &key](auto const& inner_key, auto const& value) -> bool{
+				return _add_flattened_modifier_cb(modifier_value, key, inner_key, value);
+			};
+			return expect_dictionary(callback)(value);
```
Hop311, line 714: The lambda function should go directly in `expect_dictionary`, creating a local variable (especially a `const` one) creates unnecessary copies. And again, use the correct lambda arguments

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +674,105 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
 		}
-
-		return default_callback(key, value);
+		
+		ModifierEffect const* effect = registry.get_item_by_identifier(key);
+		if(effect == nullptr) { return fallback(key, value); }
```
Hop311, line 718:
```suggestion
		if (effect == nullptr) {
			return fallback(key, value);
		}
```
Formatting standards - space after `if`, never compress expressions like this onto a single line

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +674,105 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
+			ret &= _add_flattened_modifier_cb(modifier_value, key, faction_identifier, value_node);
 
-		ret &= modifier_callback(std::move(modifier));
+			return ret;
+		}
 
-		return ret;
+		return _expect_modifier_effect_with_fallback(technology_modifier_effects, modifier_value,
+		_expect_shared_tech_country_modifier_effect(modifier_value)
+		)(key, value);
```
Hop311, line 746:
```suggestion
		return _expect_modifier_effect_with_fallback(
			technology_modifier_effects, modifier_value,
			_expect_shared_tech_country_modifier_effect(modifier_value)
		)(key, value);
```
Formatting, and the same for the functions below

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +674,105 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
-		)(node);
-
-		ret &= check_key_map_counts(key_map);
-
-		return ret;
+NodeTools::key_value_callback_t ModifierManager::expect_province_event_modifier(ModifierValue& modifier_value) const {
+	return [this, &modifier_value](auto const& key, auto const& value) {
+		if(strings_equal_case_insensitive(key, "movement_cost")) {
+			return _add_modifier_cb(modifier_value, modifier_effect_cache.movement_cost_percentage_change, value);
+		}
```
Hop311, line 768: Why is this handled specially? Is it an effect that is valid in the same way across multiple modifier types?

src/openvic-simulation/modifier/ModifierManager.hpp
```diff
@@ -19,126 +16,136 @@ namespace OpenVic {
...
-		using effect_variant_map_t = ordered_map<Modifier::modifier_type_t, ModifierEffect const*>;
+		using effect_variant_map_t = ordered_map<Modifier::modifier_origin_t, ModifierEffect const*>;
 
 		/* Some ModifierEffects are generated mid-load, such as max/min count modifiers for each building, so
 		 * we can't lock it until loading is over. This means we can't rely on locking for pointer stability,
 		 * so instead we store the effects in a deque which doesn't invalidate pointers on insert.
 		 */
 	private:
-		CaseInsensitiveIdentifierRegistry<ModifierEffect, RegistryStorageInfoDeque> IDENTIFIER_REGISTRY(modifier_effect);
+		CaseInsensitiveIdentifierRegistry<ModifierEffect, RegistryStorageInfoDeque> IDENTIFIER_REGISTRY(leader_modifier_effect);
```
Hop311, line 26: Define a type alias for this registry type and use it throughout this file and the corresponding cpp file.
A lot of duplicate code in this file can be avoided using macros.

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,166 @@
...
-	great_power { nullptr },
-	secondary_power { nullptr },
-	civilised { nullptr },
-	uncivilised { nullptr },
-	literacy { nullptr },
-	plurality { nullptr },
-	generalised_debt_default { nullptr },
-	total_occupation { nullptr },
-	total_blockaded { nullptr },
+	very_easy_player {"very_easy_player", {}, STATIC},
```
Hop311, line 13:
```suggestion
	very_easy_player { "very_easy_player", {}, STATIC },
```
Formatting - all of these should have spaces inside the brackets

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,166 @@
...
+	set_static_country_modifier(uncivilised);
+	set_static_country_modifier(literacy);
+	set_static_country_modifier(plurality);
+	set_static_country_modifier(generalised_debt_default);
+	set_static_country_modifier(total_occupation);
+	set_static_country_modifier(total_blockaded);
+	
+	constexpr IconModifier::icon_t default_icon = 0;
+	set_country_event_modifier(bad_debtor,"bad_debter", default_icon); //paradox typo
+	set_country_event_modifier(in_bankruptcy,"in_bankrupcy", default_icon); //paradox typo
```
Hop311, line 142: Why store these specifically as events? Is there anything that can be done with them that is impossible to do with the others?

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,166 @@
...
+			key_map, modifier.get_identifier(), ONE_EXACTLY,
+			expect_dictionary(modifier_manager.expect_base_country_modifier(modifier))
+		);
+	};
 
-		if (mutable_modifier != nullptr) {
-			if (multiplier != fixed_point_t::_1()) {
-				*mutable_modifier *= multiplier;
+	const auto set_country_event_modifier = [&key_map, &modifier_manager, &ret](
+		Modifier const*& modifier_ptr, std::string_view const&& identifier, IconModifier::icon_t const& icon
```
Hop311, line 68: `icon` shouldn't be `const&`, and `identifier` definitely shouldn't be `const&&`

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,166 @@
...
+	) -> void {
+		ret &= add_key_map_entry(
+			key_map, identifier, ONE_EXACTLY,
+			[&modifier_manager, &modifier_ptr, &identifier, &icon]
+			(ast::NodeCPtr const& value)->bool {
```
Hop311, line 73:
```suggestion
			[&modifier_manager, &modifier_ptr, &identifier, &icon](ast::NodeCPtr value) -> bool {
```
Formatting - should be on one line, spaces around the `->`, and ofc no `const&`

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,166 @@
...
+				modifier_ptr = modifier_manager.get_event_modifier_by_identifier(identifier);
+				if(modifier_ptr != nullptr) {
+					return true; //an event modifier overrides a static modifier with the same identifier.
+				}
+
+				ModifierValue modifier_value {};
+				bool has_parsed_modifier = expect_dictionary(modifier_manager.expect_base_country_modifier(modifier_value))(value);
+				has_parsed_modifier &= modifier_manager.add_event_modifier(identifier, std::move(modifier_value), icon, STATIC);
+				if(has_parsed_modifier) {
+					modifier_ptr = modifier_manager.get_event_modifier_by_identifier(identifier);
```
Hop311, line 83: It would be faster to just get the back modifier from the event registry's internal data deque.

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,166 @@
...
+		key_map, base_modifier.get_identifier(), ONE_EXACTLY,
+		expect_dictionary_keys_and_default(
+			modifier_manager.expect_base_country_modifier(base_modifier),
+			"supply_limit", ZERO_OR_ONE, [this, &modifier_manager](auto const& value) -> bool {
+				return modifier_manager._add_modifier_cb(
+					base_modifier,
+					modifier_manager.get_modifier_effect_cache().get_supply_limit_global_base(),
+					value
+				);
+			}
```
Hop311, line 120: Why the special handling? Is this effect valid in the same way across multiple modifier types? Also use `ast::NodeCPtr` instead of `auto const&`.

src/openvic-simulation/modifier/StaticModifierCache.hpp
```diff
@@ -1,62 +1,65 @@
...
+		Modifier PROPERTY(core);
+		Modifier PROPERTY(has_siege);
+		Modifier PROPERTY(occupied);
+		Modifier PROPERTY(nationalism);
+		Modifier PROPERTY(infrastructure);
 
 		StaticModifierCache();
 
-		bool load_static_modifiers(ModifierManager& registry);
+		bool load_static_modifiers(ModifierManager& modifier_manager, ast::NodeCPtr const& root);
```
Hop311, line 62: `root` shouldn't be `const&`

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -178,19 +179,22 @@ bool IssueManager::_load_issue(
...
-		ISSUE,
+	bool ret = NodeTools::expect_dictionary_keys_and_default(
+		modifier_manager.expect_base_country_modifier(values),
 		"is_jingoism", ZERO_OR_ONE, expect_bool(assign_variable_callback(jingoism)),
-		"rules", ZERO_OR_ONE, rule_manager.expect_rule_set(move_variable_callback(rules))
+		"rules", ZERO_OR_ONE, rule_manager.expect_rule_set(move_variable_callback(rules)),
+		"war_exhaustion_effect", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(war_exhaustion_effect))
 	)(node);
 
+	if(ret && war_exhaustion_effect != fixed_point_t::_0()) {
```
Hop311, line 194: Rather than checking it isn't zero, it would be better to get rid of the variable and instead use a custom callback that prints the warning message whenever `war_exhaustion_effect` is found regardless of value.

src/openvic-simulation/politics/NationalFocus.cpp
```diff
@@ -132,10 +130,9 @@ bool NationalFocusManager::load_national_foci_file(
 						scope_t::PROVINCE | scope_t::COUNTRY, scope_t::PROVINCE | scope_t::COUNTRY, scope_t::NO_SCOPE
 					};
 
-					bool ret = modifier_manager.expect_modifier_value_and_keys_and_default(
-						move_variable_callback(modifiers),
-						NATIONAL_FOCUS,
-						[&good_definition_manager, &encourage_goods, &pop_manager, &encourage_pop_types](
+					auto const& expect_base_province_modifier_cb = modifier_manager.expect_base_province_modifier(modifiers);
```
Hop311, line 133: `const auto ...` rather than `auto const& ...`

src/openvic-simulation/politics/NationalFocus.cpp
```diff
@@ -148,6 +145,10 @@ bool NationalFocusManager::load_national_foci_file(
 								return expect_fixed_point(map_callback(encourage_pop_types, pop_type))(value);
 							}
 
+							if(expect_base_province_modifier_cb(key, value)) {
+								return true;
+							}
+
 							return key_value_invalid_callback(key, value);
```
Hop311, line 152:
```suggestion
							return expect_base_province_modifier_cb(key, value) || key_value_invalid_callback(key, value);
```

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -505,9 +505,8 @@ callback_t<std::string_view> ConditionManager::expect_parse_identifier(
 		EXPECT_CALL(BUILDING, building_type, definition_manager.get_economy_manager().get_building_type_manager(), "FACTORY");
 		EXPECT_CALL(CASUS_BELLI, wargoal_type, definition_manager.get_military_manager().get_wargoal_type_manager());
 		EXPECT_CALL(GOVERNMENT_TYPE, government_type, definition_manager.get_politics_manager().get_government_type_manager());
-		EXPECT_CALL(MODIFIER, event_modifier, definition_manager.get_modifier_manager());
-		EXPECT_CALL(MODIFIER, triggered_modifier, definition_manager.get_modifier_manager());
-		EXPECT_CALL(MODIFIER, static_modifier, definition_manager.get_modifier_manager());
```
Hop311, line 510: So are we not able to check if a country or province has a static modifier (other than those you've treated as event modifiers)?

src/openvic-simulation/scripts/Condition.hpp
```diff
@@ -63,12 +63,13 @@ namespace OpenVic {
...
 		CASUS_BELLI     = 1 << 21,
 		GOVERNMENT_TYPE = 1 << 22,
-		MODIFIER        = 1 << 23,
+		COUNTRY_EVENT_MODIFIER = 1 << 23,
 		NATIONAL_VALUE  = 1 << 24,
 		CULTURE_UNION   = 1 << 25, // same as COUNTRY_TAG but also accepts scope this_union
 		CONTINENT       = 1 << 26,
 		CRIME           = 1 << 27,
 		TERRAIN         = 1 << 28,
+		PROVINCE_EVENT_MODIFIER = 1 << 29
```
Hop311, line 72: Move this to just below `COUNTRY_EVENT_MODIFIER` and change the bitshift numbers to follow the new order. Also move `BUILD_STRING(PROVINCE_EVENT_MODIFIER)` to below `BUILD_STRING(COUNTRY_EVENT_MODIFIER)` in the print operator function below.

Hop311 (changes requested):

src/openvic-simulation/map/Crime.cpp
```diff
@@ -1,14 +1,15 @@
 #include "Crime.hpp"
 
 #include "openvic-simulation/modifier/ModifierManager.hpp"
+#include "dataloader/NodeTools.hpp"
```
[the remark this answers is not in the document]
wvpm: Good catch. I already did that for other files but this one slipped through it.

src/openvic-simulation/map/State.cpp
```diff
@@ -65,10 +66,17 @@ void State::update_gamestate() {
...
-	industrial_power = total_factory_levels_in_state * std::clamp(
-		(fixed_point_t { potential_workforce_in_state } / 100).floor() * 400 / potential_employment_in_state,
-		fixed_point_t::_0_20(), fixed_point_t::_4()
-	);
+	fixed_point_t unclamped_score_per_factory = fixed_point_t::_0();
+
+	if(potential_employment_in_state > 0) {
+		unclamped_score_per_factory = (fixed_point_t { potential_workforce_in_state } / fixed_point_t::_100()).floor() * 400 / potential_employment_in_state;
+	}
+	else {
```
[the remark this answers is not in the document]
wvpm: I added spaces between if & (.
Why add a space after , at the end of line?
I can regex replace it everywhere.

src/openvic-simulation/scripts/Condition.cpp
wvpm: Correct. Victoria 2 only treats certain static modifiers as event modifiers.
I confirmed bad_debter and in_bankrupcy. If we find others, we can add them to the code.

src/openvic-simulation/scripts/Condition.hpp
wvpm: I'm used to appending bit flag enums for backwards compatiblity but you're right. In this case we don't need that.

src/openvic-simulation/politics/NationalFocus.cpp
wvpm: Good catch.

src/openvic-simulation/politics/Issue.cpp
wvpm: That is a better solution. Thanks for the suggestion.

src/openvic-simulation/modifier/ModifierManager.cpp
wvpm: I'll rewrite it using a macro. This looks like a great use case for one.

src/openvic-simulation/map/State.cpp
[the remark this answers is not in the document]
Hop311: Only add spaces for `,`s that *aren't* at the end of the line. I mentioned the end of the line in case you used regex, so you don't accidentally add them after commas at the end of lines

src/openvic-simulation/modifier/ModifierManager.cpp
Hop311: Normal const doesn't do anything bad but isn't really necessary for function args

src/openvic-simulation/map/State.cpp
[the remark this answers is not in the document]
wvpm: Right, that makes sense. I must have misread.

src/openvic-simulation/modifier/ModifierEffectCache.cpp
wvpm: No complaints, so guess not.

src/openvic-simulation/modifier/ModifierManager.cpp
wvpm: I also rewrote the hpp using a macro.

src/openvic-simulation/modifier/ModifierManager.cpp
wvpm: I had to pass fallback without & in the capture scope as well. It works now.

src/openvic-simulation/modifier/ModifierManager.cpp
wvpm: This is a special case. It only has an ingame effect when applied via province event modifier. Otherwise it shows up but does nothing.

src/openvic-simulation/modifier/ModifierManager.hpp
wvpm: Good idea.

src/openvic-simulation/modifier/StaticModifierCache.cpp
wvpm: Regex replaced all cases

src/openvic-simulation/modifier/StaticModifierCache.cpp
wvpm: Yes, these can be checked as if they are normal event modifiers.
Also you if you add an event modifier with the same name, it will override the static modifier.

src/openvic-simulation/modifier/StaticModifierCache.cpp
wvpm: It feels wrong for the StaticModifierCache to access the internals of the modifier_manager's event_modifiers storage imo.

src/openvic-simulation/modifier/StaticModifierCache.cpp
wvpm: Ofcourse this is yet another special case, we all know paradox loves those.
This is the only context in which you can add to the global base supply limit.
In other contexts you either add to the local supply limit or modify the global supply limit.

src/openvic-simulation/modifier/ModifierManager.cpp
wvpm: Tabs before EOL removed.
How do you get the IDE to split automatically?

src/openvic-simulation/modifier/ModifierManager.cpp
wvpm: ModifierManager uses `strings_equal_case_insensitive` from StringUtils.

src/openvic-simulation/modifier/Modifier.hpp
```diff
@@ -12,20 +12,21 @@ namespace OpenVic {
 
 	struct Modifier : HasIdentifier, ModifierValue {
 		friend struct ModifierManager;
+		friend struct StaticModifierCache;
 		friend struct UnitType;
 
-		enum struct modifier_type_t : uint8_t {
+		enum struct modifier_origin_t : uint8_t {
```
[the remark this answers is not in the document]
wvpm: Sure, there shouldn't be any ambiguity left as the mapping types are now handled compile-time.

src/openvic-simulation/modifier/ModifierManager.cpp
Hop311: You could make it auto-format the file but that would change a lot of other stuff too (including bugging out when formatting the lambda functions and making a massive mess).

src/openvic-simulation/modifier/StaticModifierCache.cpp
Hop311: You don't have to use anything private, just `&modifier_manager.get_event_modifiers().back()`. It's the standard way we get a just-added element across the codebase.

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -9,6 +9,7 @@
 #include "openvic-simulation/types/Vector.hpp"
 #include "openvic-simulation/utility/BMP.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
+#include "dataloader/NodeTools.hpp"
```
Hop311, line 12: Incorrrectly formatted include

src/openvic-simulation/map/State.cpp
```diff
@@ -65,10 +66,17 @@ void State::update_gamestate() {
...
-	industrial_power = total_factory_levels_in_state * std::clamp(
-		(fixed_point_t { potential_workforce_in_state } / 100).floor() * 400 / potential_employment_in_state,
-		fixed_point_t::_0_20(), fixed_point_t::_4()
-	);
+	fixed_point_t unclamped_score_per_factory = fixed_point_t::_0();
+
+	if (potential_employment_in_state > 0) {
+		unclamped_score_per_factory = (fixed_point_t { potential_workforce_in_state } / fixed_point_t::_100()).floor() * 400 / potential_employment_in_state;
+	}
+	else {
```
Hop311, line 74:
```suggestion
	} else {
```
Formatting

src/openvic-simulation/map/TerrainType.cpp
```diff
@@ -1,4 +1,5 @@
 #include "TerrainType.hpp"
+#include <string_view>
```
Hop311, line 2:
```suggestion
#include "TerrainType.hpp"

#include <string_view>
```
Include formatting (space between groups)

src/openvic-simulation/military/LeaderTrait.cpp
```diff
@@ -1,6 +1,7 @@
 #include "LeaderTrait.hpp"
 
 #include "openvic-simulation/modifier/ModifierManager.hpp"
+#include "openvic-simulation/dataloader/NodeTools.hpp"
```
Hop311, line 4:
```suggestion
#include "openvic-simulation/dataloader/NodeTools.hpp"
#include "openvic-simulation/modifier/ModifierManager.hpp"
```
Include formatting (alphabetical order)

src/openvic-simulation/military/UnitType.cpp
```diff
@@ -1,5 +1,6 @@
 #include "UnitType.hpp"
 
+#include "openvic-simulation/dataloader/NodeTools.hpp"
 #include "openvic-simulation/country/CountryInstance.hpp"
```
Hop311, line 4:
```suggestion
#include "openvic-simulation/country/CountryInstance.hpp"
#include "openvic-simulation/dataloader/NodeTools.hpp"
```
Include formatting (alphabetical order)

src/openvic-simulation/modifier/ModifierEffectCache.cpp
```diff
@@ -48,6 +49,12 @@ ModifierEffectCache::ship_type_effects_t::ship_type_effects_t()
 	evasion { nullptr },
 	torpedo_attack { nullptr } {}
 
+ModifierEffectCache::unit_terrain_effects_t::unit_terrain_effects_t()
+  : attack  { nullptr },
+  	defence  { nullptr },
```
Hop311, line 54:
```suggestion
	defence  { nullptr },
```
Formatting (this line has two spaces before the tab, they should be removed)

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -1,11 +1,34 @@
 #include "ModifierManager.hpp"
+#include <string_view>
+#include <utility>
+#include "openvic-simulation/dataloader/NodeTools.hpp"
```
Hop311, line 4:
```suggestion
#include "ModifierManager.hpp"

#include <string_view>
#include <utility>

#include "openvic-simulation/dataloader/NodeTools.hpp"
```
Spaces between include groups

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +659,140 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
+			return key_value_invalid_callback(flat_identifier, value);
+		}
+		return _add_modifier_cb(modifier_value, effect, value);
+	};
+}
+NodeTools::key_value_callback_t ModifierManager::expect_base_country_modifier(ModifierValue& modifier_value) const {
+	return _expect_modifier_effect_with_fallback(
+		base_country_modifier_effects,
+		modifier_value,
+	_expect_shared_tech_country_modifier_effect(modifier_value)
```
Hop311, line 772: Formatting: this should be indented inline with the other arguments above it, same for all the other functions below

src/openvic-simulation/modifier/ModifierManager.hpp
```diff
@@ -18,127 +15,123 @@ namespace OpenVic {
...
-
-	public:
-		bool add_modifier_effect(
-			ModifierEffect const*& effect_cache,
-			std::string_view identifier,
-			bool positive_good,
-			ModifierEffect::format_t format,
+	
+		bool _register_modifier_effect(
+			CaseInsensitiveIdentifierRegistry<ModifierEffect, RegistryStorageInfoDeque>& registry,
```
Hop311, line 43: Use the `modifier_effect_registry_t` type alias here and everywhere else the registry type is used

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,166 @@
 #include "StaticModifierCache.hpp"
-
+#include <cstddef>
+#include <string_view>
 #include "openvic-simulation/modifier/ModifierManager.hpp"
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "modifier/Modifier.hpp"
```
Hop311, line 6:
```suggestion

#include <cstddef>
#include <string_view>

#include "openvic-simulation/dataloader/NodeTools.hpp"
#include "openvic-simulation/modifier/Modifier.hpp"
#include "openvic-simulation/modifier/ModifierManager.hpp"
```
Include formatting: newlines between groups, alphabetical order, full paths

src/openvic-simulation/modifier/StaticModifierCache.hpp
```diff
@@ -1,62 +1,65 @@
...
+		Modifier PROPERTY(great_power);
+		Modifier PROPERTY(secondary_power);
+		Modifier PROPERTY(civilised);
+		Modifier PROPERTY(uncivilised);
+		Modifier PROPERTY(literacy);
+		Modifier PROPERTY(plurality);
+		Modifier PROPERTY(total_occupation);
+		Modifier PROPERTY(total_blockaded);
+
+		 //country event modifiers
```
Hop311, line 40:
```suggestion
		// Country event modifiers
```
Remove the space before the `//` and change the comment to match the format of `// Country modifiers` above and `// Province modifiers` below

src/openvic-simulation/politics/Issue.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Issue.hpp"
 
 #include "openvic-simulation/modifier/ModifierManager.hpp"
+#include "types/fixed_point/FixedPoint.hpp"
```
Hop311, line 4: Include format

Hop311 (commented):
Some minor remaining format comments, and nice fix for the static event modifier problem

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -1,115 +1,172 @@
...
+	set_static_country_modifier(very_easy_ai);
+	set_static_country_modifier(easy_ai);
+	set_static_country_modifier(hard_ai);
+	set_static_country_modifier(very_hard_ai);
+
+	ret &= add_key_map_entry(
+		key_map, base_modifier.get_identifier(), ONE_EXACTLY,
+		expect_dictionary_keys_and_default(
+			modifier_manager.expect_base_country_modifier(base_modifier),
+			"supply_limit", ZERO_OR_ONE, [this, &modifier_manager](auto const& value) -> bool {
```
Hop311, line 112:
```suggestion
			"supply_limit", ZERO_OR_ONE, [this, &modifier_manager](ast::NodeCPtr value) -> bool {
```
Change to explicit type rather than auto const reference

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +661,140 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
-
-node_callback_t ModifierManager::expect_modifier_value_and_key_map(
-	callback_t<ModifierValue&&> modifier_callback, Modifier::modifier_type_t type, key_map_t&& key_map
+NodeTools::key_value_callback_t ModifierManager::expect_unit_terrain_modifier(
+	ModifierValue& modifier_value,
+	const std::string_view terrain_type_identifier
 ) const {
-	return expect_modifier_value_and_key_map_and_default(
-		modifier_callback, type, key_value_invalid_callback, std::move(key_map)
+	return [this, &modifier_value, &terrain_type_identifier](const std::string_view key, const ast::NodeCPtr value) -> bool {
```
Hop311, line 761:
```suggestion
	return [this, &modifier_value, terrain_type_identifier](const std::string_view key, const ast::NodeCPtr value) -> bool {
```
`terrain_type_identifier` should be captured by value rather than reference to avoid potential issues if the function is called multiple times before the results are used (similar to the issue you fixed earlier today)

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +661,140 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
-		modifier_callback, type, default_callback, [](ModifierEffect const&) -> bool {
-			return true;
+	return [this, &registry, &modifier_value, fallback]
+		(const std::string_view key, const ast::NodeCPtr value) {
+		if (dryad::node_has_kind<ast::ListValue>(value) && complex_modifiers.contains(key)) {
+			return expect_dictionary([this, &modifier_value, &key]
+				(const std::string_view inner_key, const ast::NodeCPtr inner_value) -> bool{
+				return _add_flattened_modifier_cb(modifier_value, key, inner_key, inner_value);
+			})(value);
 		}
```
Hop311, line 709:
```suggestion
	return [this, &registry, &modifier_value, fallback](const std::string_view key, const ast::NodeCPtr value) -> bool {
		if (dryad::node_has_kind<ast::ListValue>(value) && complex_modifiers.contains(key)) {
			return expect_dictionary([this, &modifier_value, key](
				const std::string_view inner_key, const ast::NodeCPtr inner_value
			) -> bool {
				return _add_flattened_modifier_cb(modifier_value, key, inner_key, inner_value);
			})(value);
		}
```
Formatting + capture `key` by value rather than reference + return type of outer lambda function

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -686,181 +661,140 @@ bool ModifierManager::parse_scripts(DefinitionManager const& definition_manager)
...
-		modifier_callback, type, default_callback,
-		[&whitelist](ModifierEffect const& effect) -> bool {
-			return whitelist.contains(effect.get_identifier());
-		}
-	);
+NodeTools::key_value_callback_t ModifierManager::expect_leader_modifier(ModifierValue& modifier_value) const {
+	return _expect_modifier_effect(leader_modifier_effects, modifier_value);
 }
+NodeTools::key_value_callback_t ModifierManager::expect_technology_modifier(ModifierValue& modifier_value) const {
+	return [this, &modifier_value](const std::string_view key, const ast::NodeCPtr value) {
```
Hop311, line 725:
```suggestion
	return [this, &modifier_value](const std::string_view key, const ast::NodeCPtr value) -> bool {
```
Explicit return type for lambda function, add this to any others missing it too (e.g. in `expect_province_event_modifier`)

src/openvic-simulation/modifier/ModifierManager.hpp
```diff
@@ -1,15 +1,12 @@
 #pragma once
 
+#include <string_view>
 #include "openvic-simulation/modifier/Modifier.hpp"
```
Hop311, line 4:
```suggestion
#include <string_view>

#include "openvic-simulation/modifier/Modifier.hpp"
```
Newline between include groups

src/openvic-simulation/modifier/ModifierManager.cpp
wvpm: Ugh it feels like I'm programming in nodepad, does the IDE do nothing for me?

Hop311 (approved):

PR #212, opened 2024-10-28
Add market placeholder

wvpm:
Since producers have no stockpile, they will try to sell all their produce every day.
They take whatever price is offered. The market will later handle price and price change restrictions.
For now it just uses the base price.

### DIFF
--- a/src/openvic-simulation/economy/GoodInstance.cpp
+++ b/src/openvic-simulation/economy/GoodInstance.cpp
@@ -3,8 +3,36 @@
 using namespace OpenVic;
 
 GoodInstance::GoodInstance(GoodDefinition const& new_good_definition)
-  : HasIdentifierAndColour { new_good_definition }, good_definition { new_good_definition },
-	price { new_good_definition.get_base_price() }, is_available { new_good_definition.get_is_available_from_start() } {}
+  : HasIdentifierAndColour { new_good_definition },
+	sell_lock { std::make_unique<std::mutex>() },
+  	good_definition { new_good_definition },
+	price { new_good_definition.get_base_price() },
+	is_available { new_good_definition.get_is_available_from_start() },
+	total_supply_yesterday { fixed_point_t::_0() },
+	market_sell_orders {}
+	{}
+
+void GoodInstance::add_market_sell_order(GoodMarketSellOrder&& market_sell_order) {
+	const std::lock_guard<std::mutex> lock {*sell_lock};
+	market_sell_orders.push_back(std::move(market_sell_order));
+}
+
+void GoodInstance::execute_orders() {
+	const fixed_point_t price = get_price();
+
+	fixed_point_t supply_running_total = fixed_point_t::_0();
+	for(GoodMarketSellOrder const& market_sell_order : market_sell_orders) {
+		const fixed_point_t market_sell_quantity = market_sell_order.get_quantity();
+		supply_running_total += market_sell_quantity;
+		market_sell_order.get_after_trade()({
+			market_sell_quantity,
+			market_sell_quantity * price
+		});
+	}
+
+	total_supply_yesterday = supply_running_total;
+	market_sell_orders.clear();
+}
 
 bool GoodInstanceManager::setup(GoodDefinitionManager const& good_definition_manager) {
 	if (good_instances_are_locked()) {
@@ -23,4 +51,12 @@ bool GoodInstanceManager::setup(GoodDefinitionManager const& good_definition_man
 	lock_good_instances();
 
 	return ret;
+}
+
+GoodInstance& GoodInstanceManager::get_good_instance_from_definition(GoodDefinition const& good) {
+	return good_instances.get_items()[good.get_index()];
+}
+
+GoodInstance const& GoodInstanceManager::get_good_instance_from_definition(GoodDefinition const& good) const {
+	return good_instances.get_items()[good.get_index()];
 }
--- a/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
+++ b/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
@@ -4,6 +4,7 @@
 
 #include "openvic-simulation/economy/production/Employee.hpp"
 #include "openvic-simulation/economy/production/ProductionType.hpp"
+#include "openvic-simulation/economy/trading/SellResult.hpp"
 #include "openvic-simulation/map/ProvinceInstance.hpp"
 #include "openvic-simulation/map/State.hpp"
 #include "openvic-simulation/modifier/ModifierEffectCache.hpp"
@@ -14,6 +15,7 @@
 using namespace OpenVic;
 
 ResourceGatheringOperation::ResourceGatheringOperation(
+	MarketInstance& new_market_instance,
 	ModifierEffectCache const& new_modifier_effect_cache,
 	ProductionType const* new_production_type_nullable,
 	fixed_point_t new_size_multiplier,
@@ -22,7 +24,8 @@ ResourceGatheringOperation::ResourceGatheringOperation(
 	fixed_point_t new_unsold_quantity_yesterday,
 	std::vector<Employee>&& new_employees,
 	decltype(employee_count_per_type_cache)::keys_t const& pop_type_keys
-) : modifier_effect_cache { new_modifier_effect_cache },
+) : market_instance { new_market_instance },
+	modifier_effect_cache { new_modifier_effect_cache },
 	location_ptr { nullptr },
 	production_type_nullable { new_production_type_nullable },
 	revenue_yesterday { new_revenue_yesterday },
@@ -39,9 +42,11 @@ ResourceGatheringOperation::ResourceGatheringOperation(
 { }
 
 ResourceGatheringOperation::ResourceGatheringOperation(
+	MarketInstance& new_market_instance,
 	ModifierEffectCache const& new_modifier_effect_cache,
 	decltype(employee_count_per_type_cache)::keys_t const& pop_type_keys
 ) : ResourceGatheringOperation {
+	new_market_instance,
 	new_modifier_effect_cache,
 	nullptr, fixed_point_t::_0(),
 	fixed_point_t::_0(), fixed_point_t::_0(),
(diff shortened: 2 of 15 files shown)


src/openvic-simulation/economy/trading/MarketInstance.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace OpenVic {
+	struct SellResult;
+
+	struct MarketInstance { 
+		private:
+			GoodInstanceManager const* PROPERTY(good_instance_manager);
+		public:
+			bool setup(GoodInstanceManager const& new_good_instance_manager);
+			SellResult sell(GoodDefinition const& good, const fixed_point_t quantity);
```
wvpm, line 15: Mutable as we will have to updated the GoodInstance price, demand, supply, etc at some point.

src/openvic-simulation/economy/trading/MarketInstance.cpp
```diff
@@ -0,0 +1,28 @@
...
+	const fixed_point_t base_price = good_instance->get_price();
+	return {
+		quantity,
+		fixed_point_t::_0(),
+		quantity * base_price
+	};
+}
+
+SellResult::SellResult(
+	const fixed_point_t new_quantity_sold,
```
wvpm, line 21: I don't know if we will need this as it = quantity - quantity_left.

src/openvic-simulation/economy/trading/MarketInstance.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+using namespace OpenVic;
+
+bool MarketInstance::setup(GoodInstanceManager const& new_good_instance_manager) {
+	good_instance_manager = &new_good_instance_manager;
+	return true;
+}
+
+SellResult MarketInstance::sell(GoodDefinition const& good, const fixed_point_t quantity) {
+	GoodInstance const* const good_instance = good_instance_manager->get_good_instance_by_identifier(good.get_identifier());
```
wvpm, line 11: I'm assuming setup was called before sell.
If desired, I can add a null check + log.

src/openvic-simulation/economy/GoodInstance.cpp
```diff
@@ -3,8 +3,23 @@
...
-	price { new_good_definition.get_base_price() }, is_available { new_good_definition.get_is_available_from_start() } {}
+  : HasIdentifierAndColour { new_good_definition },
+  	good_definition { new_good_definition },
+	price { new_good_definition.get_base_price() },
+	is_available { new_good_definition.get_is_available_from_start() },
+	market_sell_order_quantities {},
+	market_sell_order_callbacks {}
+	{}
+
+void GoodInstance::add_order(const fixed_point_t quantity, const std::function<void(SellResult)> callback) {
```
wvpm, line 14: Should we include country already?
That would allow sphere stuff.

src/openvic-simulation/economy/trading/MarketInstance.cpp
wvpm: I decided against it and only return quanity sold.

src/openvic-simulation/economy/trading/MarketInstance.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+void MarketInstance::place_market_sell_order(const MarketSellOrder market_sell_order) {
+	GoodDefinition const* const good = market_sell_order.get_good();
+	GoodInstance* const good_instance = good_instance_manager->get_good_instance_by_identifier(good->get_identifier());
+	good_instance->add_market_sell_order(market_sell_order.get_quantity(), market_sell_order.get_after_trade());
+}
+
+void MarketInstance::execute_orders() {
+	std::vector<GoodInstance>& good_instances = good_instance_manager->get_good_instances();
+	for (GoodInstance& good_instance : good_instances) {
```
wvpm, line 22: Move the contents of this loop to GoodInstance?

wvpm: I moved the contents. That way the vectors can remain private for the goodinstance.

src/openvic-simulation/economy/GoodInstance.cpp
```diff
@@ -3,8 +3,38 @@
...
+  	good_definition { new_good_definition },
+	price { new_good_definition.get_base_price() },
+	is_available { new_good_definition.get_is_available_from_start() },
+	supply_running_total { fixed_point_t::_0() },
+	total_supply_yesterday { fixed_point_t::_0() },
+	market_sell_order_quantities {},
+	market_sell_order_callbacks {}
+	{}
+
+void GoodInstance::add_market_sell_order(const fixed_point_t quantity, const std::function<void(SellResult)> callback) {
```
wvpm, line 16: Needs to be thread safe as it could be called from parallel code (processing pops, factories, rgos & governments in parallel for example).

src/openvic-simulation/map/MapInstance.cpp
```diff
@@ -159,18 +163,30 @@ void MapInstance::update_gamestate(const Date today, DefineManager const& define
 }
 
 void MapInstance::map_tick(const Date today) {
-	for (ProvinceInstance& province : province_instances.get_items()) {
-		province.province_tick(today);
-	}
+	std::vector<ProvinceInstance>& provinces = province_instances.get_items();
+	std::for_each(
+		std::execution::par_unseq,
```
wvpm, line 168: std::execution isn't working on linux and macos

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -146,15 +151,24 @@ void ResourceGatheringOperation::rgo_tick() {
...
-	);	
+	market_instance.place_market_sell_order({
+		production_type.get_output_good(),
+		output_quantity_yesterday,
+		[
+			this,
+			total_worker_count_in_province,
+			owner_pops_cache = std::move(owner_pops_cache),
+			total_owner_count_in_state_cache
+		](const SellResult sell_result) mutable -> void {
```
Hop311, line 162:
```suggestion
			owner_pops_cache,
			total_owner_count_in_state_cache
		](const SellResult sell_result) -> void {
```
This doesn't need to be `mutable` and `owner_pops_cache` doesn't need to be moved, it's a pointer not a vector itself so moving just copies it anyway (you couldn't move the value anyway even with `std::move(*owner_pops_cache)` as it's a `const*` to a `State`'s pop cache, and if it was made a non-const pointer then moving it would clear the `State`'s pop cache until the next gamestate update)

Hop311 (commented):
Nice work, left a few comments, in particular `MarketSellOrder`s and `GoodMarketSellOrder`s should be passed around as `&&` and `std::move`d between functions.

src/openvic-simulation/economy/trading/MarketSellOrder.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include "openvic-simulation/economy/GoodDefinition.hpp"
+#include "openvic-simulation/economy/trading/SellResult.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+namespace OpenVic {
+	struct GoodMarketSellOrder {
```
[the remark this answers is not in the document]
wvpm: I think your comment is purely about naming.
MarketSellOrder is the 'normal' thing you should be using. GoodMarketSellOrder is a special thing only GoodInstance may use.

I'm open to better names but I disagree with switching the names.

src/openvic-simulation/map/MapInstance.cpp
```diff
@@ -159,7 +161,8 @@ void MapInstance::update_gamestate(const Date today, DefineManager const& define
 }
 
 void MapInstance::map_tick(const Date today) {
-	for (ProvinceInstance& province : province_instances.get_items()) {
+	std::vector<ProvinceInstance>& provinces = province_instances.get_items();
```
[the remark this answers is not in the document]
wvpm: This was a leftover for the parallel code where you need to specify the begin & end.
I'll make it inline.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: Nice catch. It's a leftover from earlier code. Should've been cleaned up in [the state owner cache PR](https://github.com/OpenVicProject/OpenVic-Simulation/pull/224).

Also noticed `produce` doesn't need the cache and so I removed it. On top of that, I renamed it to `owner_pops_cache_nullable` for clarity.

src/openvic-simulation/economy/trading/MarketSellOrder.cpp
```diff
@@ -0,0 +1,19 @@
+#include "MarketSellOrder.hpp"
+
+using namespace OpenVic;
+
+GoodMarketSellOrder::GoodMarketSellOrder(
+	const fixed_point_t new_quantity,
+	const std::function<void(const SellResult)> new_after_trade
+):
+	quantity { new_quantity },
+	after_trade { new_after_trade }
```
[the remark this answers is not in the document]
wvpm: Doesn't work, see https://discord.com/channels/1063392556160909312/1183886810850529430/1308352302184267820

src/openvic-simulation/economy/GoodInstance.cpp
wvpm: Guess not.

src/openvic-simulation/economy/trading/MarketInstance.cpp
wvpm: Setup is replaced by constructor and we use references now instead of pointers so this no longer matters.

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -104,6 +105,7 @@ namespace OpenVic {
 		size_t PROPERTY(max_supported_regiments);
 
 		ProvinceInstance(
+			MarketInstance& new_market_instance,
```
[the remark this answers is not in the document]
wvpm: Let's leave type forwarding optimisations for a separate non-functional PR.

Hop311 (approved):
Nice work, one thing to mention is that `MarketInstance` currently does very little and could easily be folded into a couple helper functions in `GoodInstanceManager`. Ofc that might not remain the case forever, but it's worth bearing in mind.

PR #218, opened 2024-10-29
Condition scripts prep work

Hop311:


src/openvic-simulation/scripts/ConditionalWeight.cpp
```diff
@@ -26,23 +26,76 @@ static NodeCallback auto expect_modifier(
...
+		std::move(key_map),
 		"modifier", ZERO_OR_MORE, expect_modifier(condition_weight_items, initial_scope, this_scope, from_scope),
 		"group", ZERO_OR_MORE, [this](ast::NodeCPtr node) -> bool {
 			condition_weight_group_t items;
+
 			const bool ret = expect_dictionary_keys(
 				"modifier", ONE_OR_MORE, expect_modifier(items, initial_scope, this_scope, from_scope)
 			)(node);
+
 			if (!items.empty()) {
```
wvpm, line 92:
```cpp
if (items.empty()) {
//error
return false;
}

//emplace
return ret
```

src/openvic-simulation/scripts/ConditionalWeight.cpp
```diff
@@ -26,23 +26,76 @@ static NodeCallback auto expect_modifier(
 }
 
 node_callback_t ConditionalWeight::expect_conditional_weight(base_key_t base_key) {
```
wvpm, line 28: Why not split this up in a method per base_key_t.
I don't see much overlap.

src/openvic-simulation/scripts/ConditionalWeight.cpp
```diff
@@ -26,23 +26,76 @@ static NodeCallback auto expect_modifier(
...
+		successfully_set_up_base_keys &= add_key_map_entries(
+			key_map,
+			"days", ZERO_OR_ONE, expect_uint<uint32_t>(time_callback("days", Timespan::from_days)),
+			"months", ZERO_OR_ONE, expect_uint<uint32_t>(time_callback("months", Timespan::from_months)),
+			"years", ZERO_OR_ONE, expect_uint<uint32_t>(time_callback("years", Timespan::from_years))
+		);
+	} break;
+	default: {
+		successfully_set_up_base_keys = false;
+	}
```
wvpm, line 70: This format of switch is painfull for me to look at.
At line 70 & 71 we see a } but they are on very different levels and yet share the indentation.
At line 38, 44 & 67 the break is hidden after a } and isn't even inside the {}.
The default case doesn't even have a break.

This might just be cpp culture shock though.

wvpm (dismissed):
It's nice to see a clean up PR before an overhaul.
This will help the overhaul PR by making it more concise.

src/openvic-simulation/scripts/ConditionalWeight.cpp
Hop311: The `modifier` and `group` cases in the `expect_dictionary_key_map` call below are needed for all base types, and in general I prefer to generalise functions where possible.

src/openvic-simulation/scripts/ConditionalWeight.cpp
Hop311: - There's no need for a break on the last case/default of a switch statement, all it does it prevent the code from continuing on to the next case (without a break the code will keep executing after hitting a valid case, regardless of other case keywords in its path)
- The code style of this repo has case keywords at the same indentation level as the swtich keyword, I've seen this a lot in C++ code. Curly brackets are often used to group each case's code more clearly, but I agree having them at the same indentation level as the outer brackets makes the code harder to read. I'll move the opening brackets onto a newline after each case and indent each case's code block one level beyond the switch and case keywords.

src/openvic-simulation/scripts/ConditionalWeight.cpp
Hop311: I prefer to have the more likely outcome first, I believe it's faster that way too (some libraries have special `unlikely` markers for `if` statements where the true case is less likely than the false case)

wvpm (approved):

PR #219, opened 2024-10-30
Always apply pop history

wvpm:
Victoria 2 uses the last pop history <= startdate.
If none is available, it picks the first pop history > startdate.


Hop311 (approved):
LGTM 👍

PR #221, opened 2024-10-31
Correct rgo size calcuation

wvpm:
When determining the rgo size multiplier, only local rgo size modifiers (from terrain or static) should be taken into account.
Also rgo size is not a valid country modifier. It is used as such in vanilla's 'no land reform' but doesn't work.


Hop311:
Even though `farm_rgo_size` and `mine_rgo_size` have no effect when used in `no_land_reform`, we need to handle them somehow to prevent errors when loading vanilla (which this branch currently has)

Hop311 (approved):

PR #222, opened 2024-10-31
Load defines directly into named and typed variables rather than using an IdentifierRegistry

Hop311:


src/openvic-simulation/dataloader/NodeTools.cpp
```diff
@@ -234,19 +234,19 @@ node_callback_t NodeTools::expect_date_identifier_or_string(callback_t<Date> cal
 }
 
 node_callback_t NodeTools::expect_years(callback_t<Timespan> callback) {
-	return expect_uint<Timespan::day_t>([callback](Timespan::day_t val) -> bool {
+	return expect_int<Timespan::day_t>([callback](Timespan::day_t val) -> bool {
```
wvpm, line 237: Why?

src/openvic-simulation/misc/Define.hpp
```diff
@@ -1,87 +1,37 @@
 #pragma once
 
-#include "openvic-simulation/pop/Pop.hpp"
-#include "openvic-simulation/types/IdentifierRegistry.hpp"
-#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/misc/defines/AIDefines.hpp"
```
wvpm, line 3: Move defines directly under openvic-simulation?

wvpm (dismissed):

src/openvic-simulation/dataloader/NodeTools.cpp
Hop311: The `CAMPAIGN_EVENT_STATE_SCALE` define value is, from what I can understand, a number of days that the event time changes by for each non colonial state, and in vanilla it's `-3` (so each non-colonial state reduces the campaign event time by 3 days). This define has a comment saying it's not used, but nonetheless I decided best to allow days/months/years to be negative so they can represent changes in amount of time if necessary. It might be worth separating these functions into regular versions (zero or positive only) and "vector time" (negative allowed) versions if there are cases where we really need to disallow negative day/month/year quantities.

src/openvic-simulation/misc/Define.hpp
Hop311: Ah yeah good idea, I was thinking to do that yesterday but forgot

wvpm (approved):

PR #224, opened 2024-11-02
Cache pops by type per state and province

wvpm:
Cache pops by type so RGOs and factories can quickly iterate over owner pops when paying.

### DIFF
--- a/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
+++ b/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
@@ -41,47 +41,23 @@ ResourceGatheringOperation::ResourceGatheringOperation(decltype(employee_count_p
 	fixed_point_t::_0(), {}, pop_type_keys
 } {}
 
-void ResourceGatheringOperation::initialise_for_new_game(ProvinceInstance& location, ModifierEffectCache const& modifier_effect_cache) {
-	if (production_type_nullable == nullptr) {
-		output_quantity_yesterday = 0;
-		revenue_yesterday = 0;
-		return;
-	}
-
-	ProductionType const& production_type = *production_type_nullable;
-	const fixed_point_t size_modifier = calculate_size_modifier(location, modifier_effect_cache);
-	const Pop::pop_size_t total_worker_count_in_province = update_size_and_return_total_worker_count(location, modifier_effect_cache, size_modifier);
-	hire(location, total_worker_count_in_province);
-	Pop::pop_size_t total_owner_count_in_state_cache = 0;
-	std::vector<Pop*> owner_pops_cache {};
-	output_quantity_yesterday = produce(location, owner_pops_cache, total_owner_count_in_state_cache, modifier_effect_cache, size_modifier);
-	revenue_yesterday = output_quantity_yesterday * production_type.get_output_good().get_base_price(); //TODO sell on market
-	pay_employees(location, revenue_yesterday, total_worker_count_in_province, owner_pops_cache, total_owner_count_in_state_cache);	
-}
-
-Pop::pop_size_t ResourceGatheringOperation::update_size_and_return_total_worker_count(
+void ResourceGatheringOperation::initialise_rgo_size_multiplier(
 	ProvinceInstance& location,
-	ModifierEffectCache const& modifier_effect_cache,
-	const fixed_point_t size_modifier
+	ModifierEffectCache const& modifier_effect_cache
 ) {
 	if (production_type_nullable == nullptr) {
 		size_multiplier = fixed_point_t::_0();
 		max_employee_count_cache = fixed_point_t::_0();
-		return fixed_point_t::_0();
+		return;
 	}
 	
-	Pop::pop_size_t total_worker_count_in_province = 0; //not counting equivalents
 	ProductionType const& production_type = *production_type_nullable;
 	std::vector<Job> const& jobs = production_type.get_jobs();
-	//can't use pop_type_distribution as it is not filled correctly yet (possibly due to equivalent pop type conversion)
-	for (Pop const& pop : location.get_pops()){
-		PopType const* pop_type = pop.get_type();
-		for(Job const& job : jobs) {
-			if (job.get_pop_type() == pop_type) {
-				total_worker_count_in_province += pop.get_size();
-				break;
-			}
-		}
+	IndexedMap<PopType, Pop::pop_size_t> const& province_pop_type_distribution = location.get_pop_type_distribution();
+	
+	Pop::pop_size_t total_worker_count_in_province = 0; //not counting equivalents
+	for(Job const& job : jobs) {
+		total_worker_count_in_province += province_pop_type_distribution[*job.get_pop_type()];
 	}
 	
 	fixed_point_t base_size_modifier = fixed_point_t::_1();
@@ -98,8 +74,9 @@ Pop::pop_size_t ResourceGatheringOperation::update_size_and_return_total_worker_
 	} else {
 		size_multiplier = ((total_worker_count_in_province / (base_size_modifier * base_workforce_size)).ceil() * fixed_point_t::_1_50()).floor();
 	}
+
+	const fixed_point_t size_modifier = calculate_size_modifier(location, modifier_effect_cache);
 	max_employee_count_cache = (size_modifier * size_multiplier * base_workforce_size).floor();
-	return total_worker_count_in_province;
 }
 
 fixed_point_t ResourceGatheringOperation::calculate_size_modifier(ProvinceInstance const& location, ModifierEffectCache const& modifier_effect_cache) const {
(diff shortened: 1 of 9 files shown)


src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -123,6 +100,51 @@ fixed_point_t ResourceGatheringOperation::calculate_size_modifier(ProvinceInstan
...
+	
+	Pop::pop_size_t total_worker_count_in_province = 0; //not counting equivalents
+	for(Job const& job : jobs) {
+		total_worker_count_in_province += province_pop_type_distribution[*job.get_pop_type()];
+	}
+	
+	hire(location, total_worker_count_in_province);
+
+	Pop::pop_size_t total_owner_count_in_state_cache = 0;
+	std::vector<Pop*>* owner_pops_cache = nullptr;
```
Hop311, line 122: Unless we're planning on adding/removing pops from this vector, we can use a `const*`. We can still access mutable `Pop*`s via a `const*` vector.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -293,32 +293,27 @@ void ResourceGatheringOperation::pay_employees(
...
+	if (total_owner_count_in_state_cache > 0) {		
 		fixed_point_t owner_share = (fixed_point_t::_2() * total_owner_count_in_state_cache / total_worker_count_in_province);
 		constexpr fixed_point_t upper_limit = fixed_point_t::_0_50();
 		if (owner_share > upper_limit) {
 			owner_share = upper_limit;
 		}
 
-		for(Pop* owner_pop_nullable : owner_pops_cache) {
-			Pop& owner_pop = *owner_pop_nullable;
+		for(Pop* owner_pop_ptr : *owner_pops_cache) {
```
Hop311, line 315: Are we sure `owner_pops_cache` can't be null here? If it can then we need to check that earlier in the function, if it can't then the function should take a reference rather than a pointer.

src/openvic-simulation/map/MapInstance.cpp
```diff
@@ -128,41 +128,44 @@ bool MapInstance::apply_history_to_provinces(
...
+	const Date today,
+	DefineManager const& define_manager,
+	ModifierEffectCache const& modifier_effect_cache
+) {
+	update_gamestate(today, define_manager);
 	for (ProvinceInstance& province : province_instances.get_items()) {
-		province.initialise_for_new_game(modifier_effect_cache);
+		province.initialise_rgo(modifier_effect_cache);
 	}
+	state_manager.tick(today, modifier_effect_cache);
```
Hop311, line 170: From my understanding, this function call ultimate results in every province belonging to a state being ticked?
- Why not just call `province.tick(...)` in the loop above after calling `initialise_rgo`?
- If the ticks must happen after all RGOs are intialised, why not just use the `MapInstance`'s tick function? Yes it'll also tick water provinces, but that's likely more than offset by the memory operations of having to loop through each state set and state, in particular because that will result in provinces being ticked in an irregular order rather than going through them in the order they're stored in memory like `MapInstance`'s tick function does.
- If we really do need to do this by state, then these starting-province-tick functions need different names as we'll likely have proper state tick functions eventually for processing state-level stuff like factories and tension, which won't tick individual provinces.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -40,10 +40,15 @@ ProvinceInstance::ProvinceInstance(
...
 	pop_type_distribution { &pop_type_keys },
+	pops_cache_by_type { &pop_type_keys },
 	ideology_distribution { &ideology_keys },
 	culture_distribution {},
 	religion_distribution {},
-	max_supported_regiments { 0 } {}
+	max_supported_regiments { 0 } {
+		for (PopType const& pop_type : pop_type_keys) {
+			pops_cache_by_type[pop_type] = {};
+		}
```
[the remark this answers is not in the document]
wvpm: Great. I'm not used to that. I'll remove it.

src/openvic-simulation/map/State.cpp
```diff
@@ -41,7 +46,11 @@ void State::update_gamestate() {
 	pop_type_distribution.clear();
 	max_supported_regiments = 0;
 
-	for (ProvinceInstance const* province : provinces) {
+	for (PopType const& pop_type : *pops_cache_by_type.get_keys()) {
+		pops_cache_by_type[pop_type].clear();
+	}
+
+	for (ProvinceInstance* const province : provinces) {
```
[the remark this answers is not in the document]
wvpm: When the province is const, the pops will be const.
I need mutable pops to give them income.

src/openvic-simulation/map/State.cpp
```diff
@@ -51,7 +60,15 @@ void State::update_gamestate() {
...
-
+		IndexedMap<PopType, std::vector<Pop*>>& province_pops_cache_by_type = province->get_mutable_pops_cache_by_type();
+		for (PopType const& pop_type : *province_pops_cache_by_type.get_keys()) {
+			std::vector<Pop*>& pops_of_type = province_pops_cache_by_type[pop_type];
+			pops_cache_by_type[pop_type].insert(
+				pops_cache_by_type[pop_type].end(),
+				pops_of_type.begin(),
+				pops_of_type.end()
+			);
+		}
```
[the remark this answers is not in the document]
wvpm: I'll try this approach.

src/openvic-simulation/map/MapInstance.cpp
wvpm: I think I did that to avoid ticking unowned provinces. I change simply add the province tick in the loop above. It will exit early if there is no rgo.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: Good point. I messed up here. If there is no owner, we should exit early. Empty provinces just don't tick. I ensured this by looping through the states but the RGO tick can simply check for owner.

wvpm: Yes, because `total_owner_count_in_state_cache > 0` and that can only happen when there is an owner pop type. If there is an owner pop type, the owner_pops_cache is filled.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -123,6 +100,51 @@ fixed_point_t ResourceGatheringOperation::calculate_size_modifier(ProvinceInstan
 	return size_modifier > fixed_point_t::_0() ? size_modifier : fixed_point_t::_0();
 }
 
+void ResourceGatheringOperation::rgo_tick(ProvinceInstance& location, ModifierEffectCache const& modifier_effect_cache) {
+	if (production_type_nullable == nullptr | location.get_owner() == nullptr) {
```
Hop311, line 104:
```suggestion
	if (production_type_nullable == nullptr || location.get_owner() == nullptr) {
```

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -123,6 +100,51 @@ fixed_point_t ResourceGatheringOperation::calculate_size_modifier(ProvinceInstan
...
+	
+	hire(location, total_worker_count_in_province);
+
+	Pop::pop_size_t total_owner_count_in_state_cache = 0;
+	std::vector<Pop*> const* owner_pops_cache = nullptr;
+
+	if (production_type.get_owner().has_value()) {
+		PopType const& owner_pop_type = *production_type.get_owner()->get_pop_type();
+		total_owner_count_in_state_cache = location.get_state()->get_pop_type_distribution()[owner_pop_type];
+		owner_pops_cache = &location.get_mutable_state()->get_mutable_pops_cache_by_type()[owner_pop_type];
```
Hop311, line 127:
```suggestion
		owner_pops_cache = &location.get_state()->get_pops_cache_by_type()[owner_pop_type];
```

src/openvic-simulation/map/State.hpp
```diff
@@ -48,6 +49,7 @@ namespace OpenVic {
 		std::string get_identifier() const;
 
 		void update_gamestate();
+		IndexedMap<PopType, std::vector<Pop*>>& get_mutable_pops_cache_by_type();
```
Hop311, line 52: This isn't needed, we can access the mutable `Pop*`s via the const getter and we shouldn't be modifying the cache itself anywhere outside of the `State`.

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -159,9 +160,11 @@ namespace OpenVic {
 		bool setup(BuildingTypeManager const& building_type_manager);
 		bool apply_history_to_province(ProvinceHistoryEntry const& entry, CountryInstanceManager& country_manager);
 
-		void initialise_for_new_game(ModifierEffectCache const& modifier_effect_cache);
+		void initialise_rgo(ModifierEffectCache const& modifier_effect_cache);
 
 		void setup_pop_test_values(IssueManager const& issue_manager);
+		State* get_mutable_state();
 		plf::colony<Pop>& get_mutable_pops();
+		IndexedMap<PopType, std::vector<Pop*>>& get_mutable_pops_cache_by_type();
```
Hop311, line 168: This isn't needed, we can access the mutable `Pop*`s via the const getter and we shouldn't be modifying the cache itself anywhere outside of the `ProvinceInstance`.

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -159,9 +160,11 @@ namespace OpenVic {
 		bool setup(BuildingTypeManager const& building_type_manager);
 		bool apply_history_to_province(ProvinceHistoryEntry const& entry, CountryInstanceManager& country_manager);
 
-		void initialise_for_new_game(ModifierEffectCache const& modifier_effect_cache);
+		void initialise_rgo(ModifierEffectCache const& modifier_effect_cache);
 
 		void setup_pop_test_values(IssueManager const& issue_manager);
+		State* get_mutable_state();
```
Hop311, line 166: This isn't needed anywhere right now, although it may be useful in the future so there's an argument for keeping it.

src/openvic-simulation/map/State.hpp
wvpm: Pops need income. This is treated as modifying the pops.

PR #228, opened 2024-11-03
ModifierEffectCache & location to fields

wvpm:
Cleans up the code a lot at the cost of storing a reference to a singleton per province.

### DIFF
--- a/src/openvic-simulation/economy/production/ResourceGatheringOperation.hpp
+++ b/src/openvic-simulation/economy/production/ResourceGatheringOperation.hpp
@@ -2,14 +2,18 @@
 
 #include "openvic-simulation/economy/production/Employee.hpp"
 #include "openvic-simulation/economy/production/ProductionType.hpp"
-#include "openvic-simulation/modifier/ModifierEffectCache.hpp"
 #include "openvic-simulation/pop/Pop.hpp"
 #include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
 #include "openvic-simulation/utility/Getters.hpp"
 
 namespace OpenVic {
+	struct ProvinceInstance;
+	struct ModifierEffectCache;
+
 	struct ResourceGatheringOperation {
 	private:
+		ModifierEffectCache const& modifier_effect_cache;
+		ProvinceInstance* location_ptr;
 		ProductionType const* PROPERTY_RW(production_type_nullable);
 		fixed_point_t PROPERTY(revenue_yesterday);
 		fixed_point_t PROPERTY(output_quantity_yesterday);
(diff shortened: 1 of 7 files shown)


src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -7,6 +7,7 @@
 #include "openvic-simulation/economy/production/ResourceGatheringOperation.hpp"
 #include "openvic-simulation/military/UnitInstance.hpp"
 #include "openvic-simulation/military/UnitType.hpp"
+#include "openvic-simulation/modifier/ModifierEffectCache.hpp"
```
Hop311, line 10: IMO it's better to forward declare `struct ModifierEffectCache`, including stuff which isn't absolutely necessary just increases compilation times

src/openvic-simulation/economy/production/ResourceGatheringOperation.hpp
```diff
@@ -8,8 +8,12 @@
 #include "openvic-simulation/utility/Getters.hpp"
 
 namespace OpenVic {
+	struct ProvinceInstance;
+
 	struct ResourceGatheringOperation {
 	private:
+		ModifierEffectCache const& modifier_effect_cache;
```
Hop311, line 15: `ModifierEffectCache` could also be forward declared in this file, and then only included in `ResourceGatheringOperation.cpp`

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -33,7 +35,7 @@ ProvinceInstance::ProvinceInstance(
 	event_modifiers {},
 	slave { false },
 	crime { nullptr },
-	rgo { pop_type_keys },
+	rgo { new_modifier_effect_cache, pop_type_keys },
```
Hop311, line 38: Rather than having a `location_ptr` in the RGO struct which is set via function later, can't it be a province reference variable which is initialised in the RGO constructor using `*this` passed from here? That guarantees that each RGO will have its province set once exactly, no risk of it being null or set multiple times.

src/openvic-simulation/map/MapInstance.hpp
```diff
@@ -44,6 +44,7 @@ namespace OpenVic {
 
 		bool setup(
 			BuildingTypeManager const& building_type_manager,
+			ModifierEffectCache const& modifier_effect_cache,
```
Hop311, line 47: `ModifierEffectCache` could also be forward declared here, assuming it currently comes only from `ProvinceInstance` and that gets replaced its own forward declaration

src/openvic-simulation/map/ProvinceInstance.cpp
wvpm: No because the province pointer can be changed when its container grows and moves everything. I already tried that.

Hop311: Ah it's not when the container expands, enough space is reserved for all the province instances, instead it's because we create the `ProvinceInstance`s as r-values and move them into the container.

Hop311 (approved):

PR #231, opened 2024-11-13
Add tax rate to `CountryInstance` -  Resolves #226

### LINKED ISSUE #226: Store tax rate
### Simulation
Each country can determine its own tax rates for each strata (poor, middle and rich). The value ranges from 0% to 100% of the pop income. Tax rate is restricted by the modifiers `min_tax` and `max_tax`.

When starting a new game, tax rate is initialised at 50%. If `min_tax` or `max_tax` don't allow that, it is set to as close as possible to 50%. 

Tasks:
- [x] Create tax rates per strata properties for CountryInstance.
- [ ] Create a method to update the tax rate per strata. This method ensures the tax rate is between `min_tax` and `max_tax`.
- [x] Handle edge cases where `min_tax` > `max_tax` (test how Victoria 2 handles it).
- [x] Initialise the tax rates to 50% using the created method.

### UI
The UI part could be split off as a separate issue if desired.

The tax rate can be adjusted using the sliders in the budget menu. The sliders are bound to the tax rate. This binding goes both ways. If the slider changes, the tax rate changes. If the tax rate changes, the slider changes. Only the player country has this binding.

- [x] Create slider per strata
- [ ] Set the sliders to the tax rates of the player country
- [ ] Set the tax rate to the slider value when sliding.

Juliusfrank11:


src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -98,6 +102,10 @@ namespace OpenVic {
 
 		/* Budget */
 		fixed_point_t PROPERTY(cash_stockpile);
+		slider_value_t PROPERTY(poor_tax_rate_slider);
```
wvpm, line 105: Sliders are UI elements to adjust the tax rate.
CountryInstance is the simulation data for the country.
CountryInstance needs tax percentage (int). (or a tax rate as fixed point)

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -227,6 +235,8 @@ namespace OpenVic {
 		bool set_ruling_party(CountryParty const& new_ruling_party);
 		bool add_reform(Reform const& new_reform);
 
+		bool set_value_for_slider(SliderManager& slider_manager, int const* new_value);
```
wvpm, line 238: Why point to an int here?
The new value can be passed by value, so `const int new_value`.

src/openvic-simulation/types/Sliders.hpp
```diff
@@ -0,0 +1,44 @@
...
+        }
+        if (value < min) {
+          value = max;
+        }
+      }
+  };
+
+class SliderManager {
+  public:
+    void set_slider_value(slider_value_t& slider, int new_value) {
```
wvpm, line 31: Why not make this a method of the slider itself?

src/openvic-simulation/types/Sliders.hpp
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+namespace OpenVic {
+  // TODO: make struct for tariff slider, unlike other sliders, you can have negative values
+  struct slider_value_t {
```
wvpm, line 8: `slider_value_t` feels wrong to me. I think `RestrictedValue` might be a better name.
This type is not directly tied to a UI slider element. It merely restricts tax and tariff rates between min and max.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -56,6 +58,11 @@ CountryInstance::CountryInstance(
 
 	/* Budget */
 	cash_stockpile { 0 },
+	poor_tax_rate_slider {50},
```
Hop311, line 61:
```suggestion
	poor_tax_rate_slider { 50 },
```

Please make the formatting here consistent, each entry in the member initialiser list should have spaces between the variable name and the `{`, and between the starting value and the `{}`

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -56,6 +58,11 @@ CountryInstance::CountryInstance(
...
+	middle_tax_rate_slider { 50},
+	rich_tax_rate_slider {50},
+
+	// TODO:
```
Hop311, line 65: What is this TODO for? Either just some stuff after the `:` or remove it

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -98,6 +102,10 @@ namespace OpenVic {
 
 		/* Budget */
 		fixed_point_t PROPERTY(cash_stockpile);
+		slider_value_t PROPERTY(poor_tax_rate_slider);
+		slider_value_t PROPERTY(middle_tax_rate_slider);
```
Hop311, line 106: Poor/middle/rich strata aren't hardcoded in OpenVic, they're dynamically loaded like most other game data. This means we can't just have a regular variable for each starta's tax rate, instead we need a map from `Strata` to `slider_value_t` (or whatever name it ends up with), specifically `IndexedMap<Strata, slider_value_t>`, only requiring the vector of loaded strata as input which can be passed to it similar to how all the other `IndexedMap` key vectors are passed around

src/openvic-simulation/types/Sliders.hpp
```diff
@@ -0,0 +1,44 @@
...
+  friend struct  SliderManager;
+  private:
+      int PROPERTY(min);
```
Hop311, line 11: Best to use `int32_t` rather than `int`, just to be certain about the data type. You could even define a type alias `using int_type = int32_t` and then just use `int_type` throughout the struct which could be useful if we ever need to change the type for any reason

src/openvic-simulation/types/Sliders.hpp
```diff
@@ -0,0 +1,44 @@
...
+      int PROPERTY(max);
```
Hop311, line 12: Is it enough to have just `value`, `min` and `max`? Don't we need another pair of final min/max values to act we limits on potentially shifting regular min/max values (e.g. so that stacking modifier effects doesn't let you move a slider outside [0,100]), probably defined as `const int_type PROPERTY(final min/max variable name)` OR they could be define via template parameters similar to what you had before (it will work with primitive types like `int32_t`)

src/openvic-simulation/types/Sliders.hpp
```diff
@@ -0,0 +1,44 @@
...
+      int PROPERTY(value);
+
+  public:
+      constexpr slider_value_t(int value, int min = 0, int max = 100) {
+        // You *can* actually have min > max in Victoria 2
+        // Such a situation will result in only being able to be move between the max and min value. 
+        // This logic replicates this "feature"
```
Hop311, line 19: Shouldn't the logic below apply only in the `max < min` case, and not in the normal `min <= max` case?
Also [std::clamp](https://en.cppreference.com/w/cpp/algorithm/clamp) could be useful, particularly for the normal case.
Furthermore, there shouldn't be any unique logic in the constructor here (except maybe for constants like the final min and max variables, once they're added), instead it should set/check values using the regular setter functions so the exact same checks occur when first constructing this object and later when updating it.

src/openvic-simulation/types/Sliders.hpp
Hop311: Yeah there's no need for a `SliderManager` at all, instead the struct itself should have these setter functions:
- `void set_value(int_type new_value)` - clamping `new_value` between min and max before setting it (any logic about the `max < min` case should happen in this function too)
- `void set_bounds(int_type new_min, int_type new_max)` - clamping min and max between the final/constant min and max values, then also updating the value in case it's no longer in the allowed range (you could simply call `set_value(value)`)

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -12,8 +13,11 @@
 #include "openvic-simulation/types/Date.hpp"
 #include "openvic-simulation/types/IdentifierRegistry.hpp"
 #include "openvic-simulation/types/IndexedMap.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/types/Sliders.hpp"
 #include "openvic-simulation/utility/Getters.hpp"
 
+
```
Hop311, line 20: Could you remove the extra newline here? In general we only ever use at most a single empty line between blocks of code

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -10,6 +10,8 @@
 #include "openvic-simulation/politics/Ideology.hpp"
 #include "openvic-simulation/research/Invention.hpp"
 #include "openvic-simulation/research/Technology.hpp"
+#include "openvic-simulation/types/Sliders.hpp"
```
Hop311, line 13: Following the code style of this repo, includes are sorted into groups (usually `<STL headers>`, maybe `<openvic-dataloader headers>`, and then `"openvic-simulation headers"`) and then alphabetically ordered within those groups. In this case, the `Sliders.hpp` include should go after the `fixed_point/FixedPoint.hpp` include

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -280,6 +287,10 @@ bool CountryInstance::add_reform(Reform const& new_reform) {
 	}
 }
 
+bool CountryInstance::set_value_for_slider(SliderManager& slider, int const* new_value) {
```
Hop311, line 290: As mentioned previously there's no need for a `SliderManager`, and this function should probably take a `Strata const&` and an `slider_value_t::int_type` as arguments - then you can use the strata to look up the right slider in the previously-mentioned `IndexedMap<Strata, slider_value_t>` and set it's value to the other argument.
We only need an external setter function for the sliders' values, all the min/max setting will be done internally in the country gamestate update function based off of modifier effect totals

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -974,8 +985,8 @@ void CountryInstance::update_modifier_sum(Date today, StaticModifierCache const&
 	modifier_sum.add_modifier(static_modifier_cache.get_infamy(), country_source, infamy);
 	modifier_sum.add_modifier(static_modifier_cache.get_literacy(), country_source, national_literacy);
 	modifier_sum.add_modifier(static_modifier_cache.get_plurality(), country_source, plurality);
-	// TODO - difficulty modifiers, war, peace, debt_default_to, bad_debter, generalised_debt_default,
-	//        total_occupation, total_blockaded, in_bankrupcy
+	// TODO - difficulty modifiers, war, peace, debt_default_to, bad_debtor, generalised_debt_default,
```
Hop311, line 988: Good catch on these typos, some of them are copying typos in the static modifier names vic2 uses but the variables in `StaticModifierCache` that they're mapped to still use correctly spelt names, so it's how we should be refering to them 👍

src/openvic-simulation/country/CountryInstance.hpp
Hop311: Nah I think we should include min/max checking in the gamestate itself to be extra safe + provide a single source of truth for slider value clamping. If anything we need another 2 variables in the struct (probably const) representing the final min and max values that the modifier-effect based min and max values are clamped between

Hop311 (commented):
Great work, let's see if we can finish this off tomorrow and get it merged 🙂

src/openvic-simulation/types/Sliders.hpp
Juliusfrank11: This was the original plan, but the tariff slider makes one edge case because unlike every other slider, it has the bounds [-100,100] and I couldn't figure out the syntax in `CountryInstance` to make it so I can specify a different "`strict_min`". I can implement this and consult on the server for the syntax

src/openvic-simulation/country/CountryInstance.cpp
Juliusfrank11: I have a spellchecker for code since I have dyslexia, I just cleaned these up to get rid of "unknown word" errors. But I'm glad they were actual typos, I made sure to download the UK english dictionary so I don't edit any UK spellings

src/openvic-simulation/types/Sliders.hpp
wvpm: Types are named PascalCase without pre- or suffix either way.

src/openvic-simulation/types/Sliders.hpp
```diff
@@ -0,0 +1,67 @@
...
+          if (this != &other) {
+              min = other.min;
+              max = other.max;
+              value = other.value;
+
+          }
+          return *this;
+      }
+    };
+}
```
Hop311, line 67:
```suggestion
#include <algorithm>
#include <cstdint>

#include "openvic-simulation/utility/Getters.hpp"

namespace OpenVic {
	using slider_value_int_type = int32_t;

	template<slider_value_int_type STRICT_MIN, slider_value_int_type STRICT_MAX>
	struct slider_value_t {
		using int_type = slider_value_int_type;

		static constexpr int_type strict_min = STRICT_MIN;
		static constexpr int_type strict_max = STRICT_MAX;

		static_assert(strict_min < strict_max);

	private:
		int_type PROPERTY(min);
		int_type PROPERTY(max);
		int_type PROPERTY(value);

	public:
		constexpr slider_value_t(
			int_type new_value = (strict_max - strict_min) / 2, int_type new_min = strict_min, int_type new_max = strict_max
		) {
			set_bounds(new_max, new_min);
			set_value(new_value);
		}

		constexpr void set_value(int_type new_value) {
			if (min <= max) {
				value = std::clamp(new_value, min, max);
			} else {
				// You *can* actually have min > max in Victoria 2
				// Such a situation will result in only being able to be move between the max and min value.
				// This logic replicates this "feature"
				if (new_value > max) {
					value = min;
				} else {
					value = max;
				}
			}
		}

		constexpr void set_bounds(int_type new_max, int_type new_min) {
			min = std::clamp(new_min, strict_min, strict_max);
			max = std::clamp(new_max, strict_min, strict_max);
			set_value(value);
		}

		constexpr slider_value_t& operator=(slider_value_t const& other) {
			if (this != &other) {
				min = other.min;
				max = other.max;
				value = other.value;
			}
			return *this;
		}
	};
}
```
There's quite a bit here but I think it will improve the struct a lot:
- Make strict min/max template arguments, allowing them to be customised for different slider values and also preventing issues with `operator=` between slider values with different strict min/max.
- You could add default values to the template arguments (0 and 100), but IMO it's more clear to explicitly define them where they're used (we won't have to do that every time as we can define other type aliases, as I'll show in `CountryInstance.hpp`).
- You can't use a type alias before it's defined, so we'll need to define `slider_value_int_type` before the template. By setting `int_type` inside the template to equal it, we make it accessible from the template's type while still only defining it in one place.
- Add variables for strict min/max so they can be accessed via instances of the template.
- The `static_assert(strict_min < strict_max);` will be checked during compilation, this is better than regular `assert`s which are only checked in debug builds (and will crash the program if they fail).
- In the constructor, we can set the default args based on the strict min/max template arguments, e.g. so that the tariff slider value has a default value of 0. We should call `set_bounds` before `set_value` as the bounds will start off default initialised to 0 (or a random value with MSVC which doesn't default initialise primitives unless you explicitly tell it to) and so setting the value first will bound it based on incorrect min/max. We could explicitly default initialise value/min/max, but just calling `set_bounds` then `set_value` is guarunteed to set everything up correctly.
- `set_value` and `set_bounds` need to be `constexpr` otherwise the constructor will fail to compile, and `operator=` can also be `constexpr`.
- `set_value` can use `std::clamp` for the `min <= max` case (this requires including `<algorithm>`) and `set_bounds` can also use it for both `min` and `max` as we ensured `strict_min < strict_max` with the earlier `static_assert`.

If you copy over the above code (keeping the `pragma once` above and making sure everything is using tabs rather than spaces) this file should be perfect.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -98,6 +101,8 @@ namespace OpenVic {
 
 		/* Budget */
 		fixed_point_t PROPERTY(cash_stockpile);
+		IndexedMap<Strata, slider_value_t> PROPERTY(tax_rate_by_strata);
```
Hop311, line 104:
```suggestion
		IndexedMap<Strata, tax_slider_value_t> PROPERTY(tax_rate_by_strata);
```
If you add the type alias `using tax_slider_value_t = slider_value_t<0, 100>;` above (e.g. under `using unit_variant_t = uint8_t;`) then we can use it throughout this class rather than needing to explicitly give the template parameters each time (even if the template has default parameters we'd still have to write `slider_value_t<>` each time).

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -280,6 +285,11 @@ bool CountryInstance::add_reform(Reform const& new_reform) {
 	}
 }
 
+bool CountryInstance::set_value_for_slider(slider_value_t& slider, const int32_t new_value) {
+	slider.set_value(new_value);
+	return true;
+}
```
Hop311, line 291:
```suggestion
void CountryInstance::set_strata_tax_rate(Strata const& strata, const tax_slider_value_t::int_type new_value) {
	tax_rate_by_strata[strata].set_value(new_value);
}
```
- This shouldn't take a `slider_value_t&`, no one outside this struct should be able to get a non-const reference to the tax rate sliders.
- Instead it should take a `Strata const&` which will be used to look up the correct slider internally.
- It doesn't need to return a `bool` if it's always going to return true.
- `new_value`'s type should match the argument for `set_value`, so `tax_slider_value_t::int_type`.
- Make sure to change the function's header declaration to:
```cpp
		void set_strata_tax_rate(Strata const& strata, const tax_slider_value_t::int_type new_value);
```

src/openvic-simulation/types/Sliders.hpp
wvpm: This is a more complex object. It has dynamic state validation.

Juliusfrank11: I see, I will change it to `SliderValue`

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -280,6 +285,10 @@ bool CountryInstance::add_reform(Reform const& new_reform) {
 	}
 }
 
+void CountryInstance::set_strata_tax_rate(Strata const& strata, const StandardSliderValue::int_type new_value) {
```
wvpm, line 288: `new_value` is very vague.
`tax_percentage` would be better.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1,4 +1,5 @@
 #include "CountryInstance.hpp"
+#include <cstdint>
```
wvpm, line 2: Not needed I think.

src/openvic-simulation/country/CountryInstance.cpp
Hop311: I think it's clear enough, after all it's an integer type so it definitely can't be a decimal proportion

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -98,6 +104,8 @@ namespace OpenVic {
 
 		/* Budget */
 		fixed_point_t PROPERTY(cash_stockpile);
+		IndexedMap<Strata, StandardSliderValue> PROPERTY(tax_rate_by_strata);
```
wvpm, line 107: How/where are the min and max values adjusted based on modifiers (policy) or is that out of scope?
(Yes I know I wrote the issue and should've made the scope clearer....)

src/openvic-simulation/types/SliderValue.hpp
```diff
@@ -0,0 +1,56 @@
...
+			set_bounds(new_max, new_min);
+			set_value(new_value);
+		}
+
+		constexpr void set_value(int_type new_value) {
+			if (min <= max) {
+				value = std::clamp(new_value, min, max);
+			} else {
+				// You *can* actually have min > max in Victoria 2
+				// Such a situation will result in only being able to be move between the max and min value.
```
wvpm, line 40: A well placed code comment :thumbsup:

src/openvic-simulation/types/SliderValue.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <algorithm>
+#include <cstdint>
+
+#include "openvic-simulation/utility/Getters.hpp"
+
+namespace OpenVic {
+	using slider_value_int_type = int32_t;
+
+	template<slider_value_int_type STRICT_MIN, slider_value_int_type STRICT_MAX>
```
wvpm, line 11: Nice solution with a template for min and max + static assert.

src/openvic-simulation/country/CountryInstance.cpp
Hop311: Although we get that include from other headers too, we do use types defined in `<cstdint>` in this file so it's not useless/wasteful. If we go for a full "include what you use" approach like Spartan advocates, then for every file we'll have every single header who's contents we need included explicitly (this doesn't result in any meaningful compile or run time slowdown, and it avoid issues when removing includes in headers that other files depend on but don't include directly themselves).

Hop311 (approved):

src/openvic-simulation/country/CountryInstance.hpp
wvpm: Out of scope then, great follow up issue.

Issue #225, opened 2024-11-03
Store pop income


wvpm:
Pops gain income from various sources. The income per source is used in tooltips. Tooltips only show todays income.
Sources include:

- paychecks from RGO or factory work
- paychecks from RGO or factory ownership
- income from selling artisanal production
- unemployment subsidies
- pensions
- government salary (administration, education and/or military)
- one time income from event or decision
- loan interest (not enabled in Victoria 2, but could be enabled in OpenVic as a game option or mod configuration)

Relevant file: https://github.com/OpenVicProject/OpenVic-Simulation/blob/master/src/openvic-simulation/pop/Pop.hpp & the .cpp

Tasks:

- [x] Create a method that accepts RGO worker income (`Pop::add_rgo_worker_income`)
- [x] Create a method that accepts RGO owner income (`Pop::add_rgo_owner_income`)
- [ ] Create a method that accepts artisanal income
- [ ] Create a method that accepts factory worker income
- [ ] Create a method that accepts factory owner income
- [ ] Create a method that accepts unemployment subsidies
- [ ] Create a method that accepts pensions
- [ ] Create a method that accepts government salary from administration
- [ ] Create a method that accepts government salary from education
- [ ] Create a method that accepts government salary from military
- [ ] Create a method that accepts income from event/decision
- [ ] Create a method that accepts loan interest payments
- [ ] For each above method, store the income and its source for 1 day tick. (So it starts counting from 0 every day.) Note methods can be called multiple times in 1 day, so sum the income per source.
- [ ] Store the income from all sources in the `income` property.

Below is an image of how Victoria 2 uses the tooltips. Note that the Victoria 2 groups certain types together that OpenVic might keep separate, hence the separate methods.
![{12A2ED07-9033-4E9E-865F-8463FE505194}](https://github.com/user-attachments/assets/bf0a8958-9ac5-4676-913a-1f5775cf8abf)

PR #233, opened 2024-11-15
Added stores for pop Income solves #225

Fxrmillan:
I added the stores directly to the pop type. I thought of creating an object for them but worried it would hurt performance. I also considered making them a single array but worried that might hurt readability. For the clear_income_per_source() function, I wasn't sure how to get it to trigger every tick, and I'm not sure creating new fixed_point objects would be the most efficient approach. One idea I had was to zero out the sources by subtracting them from themselves, but I was worried that would cause problems since the economy is parallelized.

### DIFF
--- a/src/openvic-simulation/pop/Pop.cpp
+++ b/src/openvic-simulation/pop/Pop.cpp
@@ -1,5 +1,6 @@
+#define KEEP_DO_FOR_ALL_TYPES_OF_INCOME
 #include "Pop.hpp"
-
+#undef KEEP_DO_FOR_ALL_TYPES_OF_INCOME
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/country/CountryInstance.hpp"
 #include "openvic-simulation/defines/Define.hpp"
@@ -44,6 +45,11 @@ Pop::Pop(PopBase const& pop_base, decltype(ideologies)::keys_t const& ideology_k
 	life_needs_fulfilled { 0 },
 	everyday_needs_fulfilled { 0 },
 	luxury_needs_fulfilled { 0 },
+	#define INITALIZE_POP_INCOME_STORES(name)\
+		name { 0 },
+
+	DO_FOR_ALL_TYPES_OF_POP_INCOME(INITALIZE_POP_INCOME_STORES)
+	#undef INITALIZE_POP_INCOME_STORES
 	max_supported_regiments { 0 } {}
 
 void Pop::setup_pop_test_values(IssueManager const& issue_manager) {
@@ -159,9 +165,23 @@ void Pop::update_gamestate(
 	}
 }
 
-//TODO store income
-void Pop::add_rgo_owner_income(const fixed_point_t income) {}
-void Pop::add_rgo_worker_income(const fixed_point_t income) {}
+#define DEFINE_ADD_INCOME_FUNCTIONS(name)\
+	void Pop::add_##name(const fixed_point_t pop_income){\
+		name += pop_income;\
+		income += pop_income;\
+	}
+
+DO_FOR_ALL_TYPES_OF_POP_INCOME(DEFINE_ADD_INCOME_FUNCTIONS)
+#undef DEFINE_ADD_INCOME_FUNCTIONS
+
+#define SET_ALL_INCOME_TO_ZERO(name)\
+	name = fixed_point_t::_0();
+
+void Pop::clear_all_income(){
+	DO_FOR_ALL_TYPES_OF_POP_INCOME(SET_ALL_INCOME_TO_ZERO)
+	#undef DO_FOR_ALL_TYPES_OF_POP_INCOME
+	#undef SET_ALL_INCOME_TO_ZERO
+}
 
 Strata::Strata(std::string_view new_identifier) : HasIdentifier { new_identifier } {}
 
--- a/src/openvic-simulation/pop/Pop.hpp
+++ b/src/openvic-simulation/pop/Pop.hpp
@@ -49,6 +49,26 @@ namespace OpenVic {
 		);
 	};
 
+	#define DO_FOR_ALL_TYPES_OF_POP_INCOME(F)\
+		F(rgo_owner_income)\
+		F(rgo_worker_income)\
+		F(artisanal_income)\
+		F(factory_worker_income)\
+		F(factory_owner_income)\
+		F(unemployment_subsidies)\
+		F(pensions)\
+		F(government_salary_administration)\
+		F(government_salary_education)\
+		F(government_salary_military)\
+		F(event_and_decision_income)\
+		F(loan_interest_payments)
+	
+	#define DECLARE_POP_INCOME_STORES(income_type)\
+		fixed_point_t PROPERTY(income_type); 
+	
+	#define DECLARE_POP_INCOME_STORE_FUNCTIONS(name)\
+		void add_##name(const fixed_point_t pop_income);
+
 	/* REQUIREMENTS:
 	 * POP-18, POP-19, POP-20, POP-21, POP-34, POP-35, POP-36, POP-37
 	 */
@@ -84,6 +104,9 @@ namespace OpenVic {
 		fixed_point_t PROPERTY(everyday_needs_fulfilled);
 		fixed_point_t PROPERTY(luxury_needs_fulfilled);
 
+		DO_FOR_ALL_TYPES_OF_POP_INCOME(DECLARE_POP_INCOME_STORES);
+		#undef DECLARE_POP_INCOME_STORES
+		
 		size_t PROPERTY(max_supported_regiments);
 
 		Pop(PopBase const& pop_base, decltype(ideologies)::keys_t const& ideology_keys);
@@ -104,8 +127,10 @@ namespace OpenVic {
 			const fixed_point_t pop_size_per_regiment_multiplier
 		);
 
-		void add_rgo_owner_income(const fixed_point_t income);
-		void add_rgo_worker_income(const fixed_point_t income);
+		DO_FOR_ALL_TYPES_OF_POP_INCOME(DECLARE_POP_INCOME_STORE_FUNCTIONS)
+		#undef DECLARE_POP_INCOME_STORE_FUNCTIONS
+		void clear_all_income();
+
 	};
 
 	struct Strata : HasIdentifier {
@@ -346,3 +371,6 @@ namespace OpenVic {
 		bool parse_scripts(DefinitionManager const& definition_manager);
 	};
 }
+#ifndef KEEP_DO_FOR_ALL_TYPES_OF_INCOME 
+	#undef DO_FOR_ALL_TYPES_OF_POP_INCOME
+#endif


wvpm:
Don't worry about "creating new fixed_point objects", think of them as integers.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -44,7 +44,17 @@ Pop::Pop(PopBase const& pop_base, decltype(ideologies)::keys_t const& ideology_k
...
 	luxury_needs_fulfilled { 0 },
-	max_supported_regiments { 0 } {}
+	max_supported_regiments { 0 },
+	artisanal_income{ 0 },
+	factory_worker_income{ 0 },
+	factory_owner_income{ 0 },
+	unemployment_subsidies_income{ 0 },
+	pensions_income{ 0 },
+	government_salary_administration_income{ 0 },
+    government_salary_education_income{ 0 },
```
wvpm, line 54: Please add a tab here and a space before the { for the new properties.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
 //TODO store income
 void Pop::add_rgo_owner_income(const fixed_point_t income) {}
 void Pop::add_rgo_worker_income(const fixed_point_t income) {}
+//adds income to the temporary stores and then to general income store
```
wvpm, line 175: No need for this comment. The code is clear enough.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
 //TODO store income
```
wvpm, line 172: Edit the 2 methods below to store the income just like the new methods do.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
...
+}
+void Pop::add_factory_worker_income(const fixed_point_t income){
+	this->factory_worker_income += income;
+	this->income += income;
+}
+void Pop::add_factory_owner_income(const fixed_point_t income){
+	this->factory_owner_income += income;
+	this->income += income;
+}
+void Pop::add_unemployment_subsidies_income(const fixed_point_t income){
```
wvpm, line 188: I'd drop the _income for unemployment_subsidies, pensions and government_salary.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
...
+void Pop::add_event_and_decision_income(const fixed_point_t income){
+	this->event_and_decision_income += income;
+	this->income += income;
+}
+void Pop::add_loan_interest_payments(const fixed_point_t income){
+	this->loan_interest_payments += income;
+	this->income += income;
+}
+void Pop::clear_income_per_source(){ //resets all of the seprate stores
+	this->artisanal_income = fixed_point_t(0);
```
wvpm, line 217: No need for `this->` anywhere.

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -83,6 +83,22 @@ namespace OpenVic {
...
+		fixed_point_t PROPERTY(factory_worker_income);
+		fixed_point_t PROPERTY(factory_owner_income);
+		fixed_point_t PROPERTY(unemployment_subsidies_income);
+		fixed_point_t PROPERTY(pensions_income);
+		fixed_point_t PROPERTY(government_salary_administration_income);
+		fixed_point_t PROPERTY(government_salary_education_income);
+		fixed_point_t PROPERTY(government_salary_military_income);
+		fixed_point_t PROPERTY(event_and_decision_income);
+		fixed_point_t PROPERTY(loan_interest_payments);
+
```
wvpm, line 96: No need for new lines here.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
...
+}
+void Pop::add_event_and_decision_income(const fixed_point_t income){
+	this->event_and_decision_income += income;
+	this->income += income;
+}
+void Pop::add_loan_interest_payments(const fixed_point_t income){
+	this->loan_interest_payments += income;
+	this->income += income;
+}
+void Pop::clear_income_per_source(){ //resets all of the seprate stores
```
wvpm, line 216: `clear_income_sources` or `clear_all_income` would be a better name than "per source".
"per source" sounds like it expects a source as parameter.

Also the code is clear enough, so no need for comment.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
...
+}
+void Pop::add_government_salary_administration_income(const fixed_point_t income){
+	this->government_salary_administration_income += income;
+	this->income += income;
+}
+void Pop::add_government_salary_education_income(const fixed_point_t income){
+	this->government_salary_education_income += income;
+	this->income += income;
+}
+void Pop::add_government_salary_military_income(const fixed_point_t income){
```
wvpm, line 204: These are a lot of very similar methods. If you want, you can refactor them using a macro.
A macro is basically build in code generation.

wvpm (changes requested):
add_rgo_owner_income & add_rgo_worker_income are still empty.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
 //TODO store income
 void Pop::add_rgo_owner_income(const fixed_point_t income) {}
 void Pop::add_rgo_worker_income(const fixed_point_t income) {}
+//adds income to the temporary stores and then to general income store
+void Pop::add_artisanal_income(const fixed_point_t income){
+	artisanal_income += income;
+	this->income += income;
```
Hop311, line 178: In all of these income adding functions, change the argument's name to something other than `income` so there's no risk of clash with the `Pop`'s `income` member variable, and then remove all the `this->`s which will no longer be necessary

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -162,6 +172,59 @@ void Pop::update_gamestate(
...
+}
+void Pop::add_loan_interest_payments(const fixed_point_t income){
+	this->loan_interest_payments += income;
+	this->income += income;
+}
+void Pop::clear_income_per_source(){ //resets all of the seprate stores
+	this->artisanal_income = fixed_point_t(0);
+	this->factory_worker_income = fixed_point_t(0);
+	this->factory_owner_income = fixed_point_t(0);
```
Hop311, line 219: Replace all uses of `fixed_point_t(0)` with `fixed_point_t::_0()`, using a fixed constant rather than going through the normal constructor

src/openvic-simulation/pop/Pop.hpp
Hop311: Following the style of the repo, there should never be more than a single blank line regardless of the context

src/openvic-simulation/pop/Pop.cpp
Hop311: Very good idea, could also be used in the variable and function declarations to streamline the whole process. We can discuss this tomorrow if you want any guidance on it 🙂

Hop311 (commented):
Great work, I left a couple comments on top of General's review, but we can get this all worked out easily tomorrow and get this merged within the next 18 hours 👍

src/openvic-simulation/pop/Pop.cpp
wvpm: `amount` for example

Fxrmillan:
I was struggling with some of the namespace access for the macros so please feel free to correct me if I put them in an Unoptimal spot

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -49,6 +49,26 @@ namespace OpenVic {
...
+		F(government_salary_education)\
+		F(government_salary_military)\
+		F(event_and_decision_income)\
+		F(loan_interest_payments)
+	
+	#define DECLARE_POP_INCOME_STORES(income_type)\
+		fixed_point_t PROPERTY(income_type); 
+	
+	#define DECLARE_POP_INCOME_STORE_FUNCTIONS(name)\
+		void add_##name(const fixed_point_t income);
```
wvpm, line 70:
```suggestion
		void add_##name(const fixed_point_t amount);
```

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -159,9 +164,22 @@ void Pop::update_gamestate(
...
+		income += popIncome;\
+	}
+
+DO_FOR_ALL_TYPES_OF_POP_INCOME(DEFINE_ADD_INCOME_FUNCTIONS)
+#undef DEFINE_ADD_INCOME_FUNCTIONS
+
+#define SET_ALL_INCOME_TO_ZERO(name)\
+	name = fixed_point_t::_0();
+
+void Pop::clear_all_income(){ //resets all of the seprate stores
```
wvpm, line 179: No need for a comment. The code is clear.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -159,9 +164,22 @@ void Pop::update_gamestate(
 	}
 }
 
-//TODO store income
-void Pop::add_rgo_owner_income(const fixed_point_t income) {}
-void Pop::add_rgo_worker_income(const fixed_point_t income) {}
+#define DEFINE_ADD_INCOME_FUNCTIONS(name)\
+	void Pop::add_##name(const fixed_point_t popIncome){\
```
wvpm, line 168: It's best practice to match parameter names in the .hpp
For the compiler it's fine, for a human reader it can be confusing.

wvpm (dismissed):
Nice work

wvpm (dismissed):

Spartan322 (changes requested):
This needs to be squashed

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -346,3 +371,6 @@ namespace OpenVic {
 		bool parse_scripts(DefinitionManager const& definition_manager);
 	};
 }
+#ifndef DO_FOR_ALL_TYPES_OF_POP_INCOME
```
Hop311, line 374: The define for keeping this macro needs to have a different name from the macro, otherwise it gets overwritten by the macro (producing a compiler warning but still working) and then doesn't get undefined since `#ifndef DO_FOR_ALL_TYPES_OF_INCOME` always fails.
Change it to `KEEP_DO_FOR_ALL_TYPES_OF_INCOME` here and at the top of the cpp file

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -159,9 +165,22 @@ void Pop::update_gamestate(
...
+
+DO_FOR_ALL_TYPES_OF_POP_INCOME(DEFINE_ADD_INCOME_FUNCTIONS)
+#undef DEFINE_ADD_INCOME_FUNCTIONS
+
+#define SET_ALL_INCOME_TO_ZERO(name)\
+	name = fixed_point_t::_0();
+
+void Pop::clear_all_income(){
+	DO_FOR_ALL_TYPES_OF_POP_INCOME(SET_ALL_INCOME_TO_ZERO)
+	#undef DO_FOR_ALL_TYPES_OF_POP_INCOME
```
Hop311, line 182: Add this as well:
```cpp
#undef SET_ALL_INCOME_TO_ZERO
```

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -84,6 +104,9 @@ namespace OpenVic {
 		fixed_point_t PROPERTY(everyday_needs_fulfilled);
 		fixed_point_t PROPERTY(luxury_needs_fulfilled);
 
+		DO_FOR_ALL_TYPES_OF_POP_INCOME(DECLARE_POP_INCOME_STORES);
+		#undef DEFINE_INCOME_TYPES
```
Hop311, line 108: Should be:
```suggestion
		#undef DECLARE_POP_INCOME_STORES
```

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -104,8 +127,10 @@ namespace OpenVic {
 			const fixed_point_t pop_size_per_regiment_multiplier
 		);
 
-		void add_rgo_owner_income(const fixed_point_t income);
-		void add_rgo_worker_income(const fixed_point_t income);
+		DO_FOR_ALL_TYPES_OF_POP_INCOME(DECLARE_POP_INCOME_STORE_FUNCTIONS)
+		#undef DECLARE_POP_INCOME_ADD_FUNCTIONS
```
Hop311, line 131: Should be: 
```suggestion
		#undef DECLARE_POP_INCOME_STORE_FUNCTIONS
```

Hop311 (commented):
Some final comments + squashing the PR into a single commit, then we'll be ready to merge

PR #234, opened 2024-11-17
Expand Mapmode struct + add default mapmodes

Hop311:


BrickPi (approved):
Important changes, LGTM. We eventually need to sort out if/how we can expose discrete stuff like the religion mapmode to modders in a hardcoded v2 (even if for that limited case its just add an extra mapmode and let modders include it in the GUI should they choose to).

PR #235, opened 2024-11-18
Split pop.hpp

wvpm:
Split pop.hpp in 3:
1. Pop.hpp (for the actual pops)
2. PopSize.hpp for the pop size type only
3. PopType.hpp for strata, poptype & the manager.

This removes dependencies on Pop in production type and breaks circular dependencies.
See also https://discord.com/channels/1063392556160909312/1064542494114725909/1308169974182842440


Hop311 (approved):
LGTM, needing to include `Pop.hpp` for `pop_size_t` has been annoying for a long time, there are a lot of files where we could probably just include `PopSize.hpp` and forward declare anything else like `Pop`/`PopBase`/`PopType`, but that can be left for a dedicated include cleanup PR.
My only other comment is that I would personally put `PopSize.hpp` in the `pop` folder as it's specifically to do with pops rather than be a general use type, but that's just a matter of personal preference.

PR #243, opened 2024-12-03
Fix IndexedMap for bool Value case

Hop311:


src/openvic-simulation/types/IndexedMap.hpp
```diff
@@ -302,17 +311,27 @@ namespace OpenVic {
...
 			const size_t index = get_index_from_item(key);
 			if (index < get_values().size()) {
-				return &get_values()[index];
+				if constexpr (std::same_as<value_const_ptr_type, value_const_iterator>) {
+					return get_values().begin() + index;
+				} else {
+					return &get_values()[index];
+				}
+			} else {
+				return {};
```
wvpm, line 333: Why return a new object instead of a nullptr here?

Spartan322 (approved):

PR #246, opened 2024-12-14
Fix base_iterator_t friend of itself warning

Spartan322:


Hop311:
`IndexedMap` can be tested by running the headless SIM built from [the test-indexed-map branch](https://github.com/OpenVicProject/OpenVic-Simulation/tree/test-indexed-map)

Hop311 (approved):

PR #249, opened 2024-12-15
Add flag struct

SegmentFaultEnjoyer:
Adds two files called Flags.hpp and Flags.cpp, and use them for country, province and global flags


Spartan322:
Flags.hpp needs to be renamed to GlobalFlags.hpp, also I don't see why flags need to be a macro. They should be a class.

BrickPi (approved):

PR #256, opened 2024-12-20
Ideology, issue, vote, culture and religion distributions + leader name check

Hop311:


Spartan322:
02f2b0b's commit message is too long, also why is 56704b5 separated from it?

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -68,9 +68,12 @@ namespace OpenVic {
 
 		fixed_point_t PROPERTY(literacy);
 
-		IndexedMap<Ideology, fixed_point_t> PROPERTY(ideologies);
-		fixed_point_map_t<Issue const*> PROPERTY(issues);
-		IndexedMap<CountryParty, fixed_point_t> PROPERTY(votes);
+		// All of these should have a total size equal to the pop size, allowing the distributions from different pops to be
+		// added together with automatic weighting based on their relative sizes. Similarly, the province, state and country
+		// equivalents of these distributions will have a total size equal to their total population size.
+		IndexedMap<Ideology, fixed_point_t> PROPERTY(ideology_distribution);
```
BrickPi, line 74: Why aren't these `fixed_point_map_t`?

Hop311: IndexedMaps are better than fixed point maps (hash maps) by almost every metric, faster reading and writing, take up less space to store the same amount of data, etc. The drawback is that they always have an entry for every single key, which leads to a lot of wasted space for maps with a lot of unused keys (e.g. culture distributions usually use a small fraction of all the defined cultures). For stuff like pop types and ideologies we expect most maps to have an entry for almost all keys, so it's definitely worth using IndexedMaps.

BrickPi (approved):
Can't wait for more pie charts! Great work.

PR #265, opened 2025-01-17
Add country military stats, load assigned leaders, fix 1861 core errors

Hop311:


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1063,7 +1077,42 @@ void CountryInstance::_update_military(
 		mobilisation_potential_regiment_count = mobilisation_max_regiment_count;
 	}
 
-	// TODO - update max_ship_supply, leadership_points, war_exhaustion
+	war_exhaustion_max = get_modifier_effect_value_nullcheck(modifier_effect_cache.get_max_war_exhaustion());
+	war_exhaustion = std::clamp(war_exhaustion, fixed_point_t::_0(), war_exhaustion_max);
```
wvpm, line 1081: This is wrong.
war_exhaustion_max is not a 'hard' limit for war exhaustion.
It merely prevents increasing it above that level.

Imagine you have war_exhaustion_max 50 and war_exhaustion 49, you then gain 2 exhaustion. This would increase it to 50 instead of 51 due to war_exhaustion_max.

If you have war_exhaustion 50 and war_exhaustion_max 20, war_exhaustion would still be 50.
It can decrease (down to the hard lower limit of 0).

Also there is a hard limit of 100.

src/openvic-simulation/defines/Define.cpp
```diff
@@ -22,4 +22,17 @@ bool DefineManager::load_defines_file(ast::NodeCPtr root) {
 			pops_defines.get_name(), ONE_EXACTLY, pops_defines.expect_defines()
 		)
 	)(root);
+
+	// TODO - move to MilitaryDefines?
+
+	if (military_defines.get_leader_recruit_cost() <= fixed_point_t::_0()) {
```
wvpm, line 28: This doesn't seem like functional error to me.
It is unexpected configuration but not invalid.

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -157,8 +157,12 @@ namespace OpenVic {
 
 		bool set_owner(CountryInstance* new_owner);
 		bool set_controller(CountryInstance* new_controller);
-		bool add_core(CountryInstance& new_core);
-		bool remove_core(CountryInstance& core_to_remove);
+		// The warn argument controls whether a log message is emitted when a core already does/doesn't exist, e.g. we may
+		// want to know if there are redundant province history instructions setting a core multiple times, but we may not
+		// want to be swamped with log messages by effect scripts which use add_core/remove_core very liberally. Regardless
+		// of warn's value, both functions will still return true if the core in question already does/doesn't exist.
+		bool add_core(CountryInstance& new_core, bool warn);
```
wvpm, line 164: Nice change.
I expected add_x not to error when x is already present.

Also cores are unique, so adding a core twice cannot result in a double core or other strange outcome.
Adding a core to a province that already has one, should simply do nothing.

wvpm (changes requested):
war_exhaustion_max

src/openvic-simulation/defines/Define.cpp
Hop311: It's not a valid configuration in the base game:
- `LEADER_RECRUIT_COST = 0` causes Vic2 to crash when you start a campaign (presumably due to divide by 0 errors and/or AIs trying to hire infinitely many leaders).
- `LEADER_RECRUIT_COST < 0` results in you starting with a negative amount of leadership and being able to hire a single leader after which your leaderships updates to 0 and you're never able to get it negative again, even with console commands, and so never able to hire any more leaders.

Of course we don't want to replicate the crash, and I don't think we should replicate the buggy negative cost behaviour either. I considered treating `LEADER_RECRUIT_COST = 0` as free leaders, but as this is an extension beyond Vic2's behaviour I don't think it's currently worth the effort of adapting the UI and AI/auto-create leaders setting to work around it. The simplest solution to keep OpenVic from crashing when given these configurations which are invalid in Vic2 is to default to a valid configuration, e.g. `LEADER_RECRUIT_COST = 1`. Furthermore, I decided to do this check during defines loading so it happens once with a single clear error message rather than having to make unnecessary checks every time we use the value (i.e. once per country per gamestate update).

src/openvic-simulation/country/CountryInstance.cpp
wvpm: Nope, you were right and pointed this out in the discord. See https://discord.com/channels/1063392556160909312/1077648343863406613/1330326727104135189 and beyond for reference.

wvpm (approved):

src/openvic-simulation/country/CountryInstance.cpp
Hop311: - The upper limit is hard when applied, it's just that it's only applied when the `war_exhaustion` value changes, so you can have something like `50/20` after max war exhaustion goes down but it'll get clamped to the `[0, 20]` range immediately after any changes are made to the current war exhaustion value (e.g. on each month tick while at peace war exhaustion changes by `-1`, forcing clamping). I've otped to handle this by not changing `war_exhaustion` during gamestate updates, only recalculating `max_war_exhaustion`, and then insisting that `war_exhaustion` is only ever changed via the `change_war_exhaustion(delta)` function which adds `delta` to `war_exhaustion` and clamps the result in the `[0, max_war_exhaustion]` range.
- Also 100 isn't a hard limit on max war exhaustion, it's just the highest value used in vanilla Vic2, you can push it higher by editing the `war` or `peace` static modifiers or adding your own modifier with a `max_war_exhaustion` effect.
- Max war exhaustion also doesn't have a hard lower limit (i.e. it can go negative), but this results in buggy current war exhaustion value clamping behaviour as well as seemingly not providing any war exhaustion modifier contributions (the UI is very buggy so it's hard to tell what is and isn't real in these cases, but I couldn't see negative war exhaustion resulting in positive RGO throughput modifier effect contributions, for example). I've opted to implement a hard lower limit of `0` to avoid this buggy behaviour, if someone does end up relying on it (and they can show us how it works) then we can adapt the system in the future to accomodate it.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1063,7 +1081,46 @@ void CountryInstance::_update_military(
...
+	naval_unit_start_experience +=
+		get_modifier_effect_value_nullcheck(modifier_effect_cache.get_naval_unit_start_experience());
+
+	recruit_time = fixed_point_t::_1() +
+		get_modifier_effect_value_nullcheck(modifier_effect_cache.get_unit_recruitment_time());
+	combat_width = (
+		fixed_point_t::parse(military_defines.get_base_combat_width()) +
+		get_modifier_effect_value_nullcheck(modifier_effect_cache.get_combat_width_additive())
+	) * (
+		fixed_point_t::_1() + get_modifier_effect_value_nullcheck(modifier_effect_cache.get_combat_width_percentage_change())
```
wvpm, line 1112: combat_width_percentage_change is only from terrain according to the google sheet.
It makes no sense to apply it to a country instance.

If you intended to allow combat_width_percentage_change to be used as a country modifier effect, then this is fine.

wvpm (dismissed):

wvpm (approved):

PR #271, opened 2025-01-22
Accurately mimic Victoria 2's farm & mine mechanics + game rules

wvpm:
Context: https://discord.com/channels/1063392556160909312/1064542494114725909/1331560403112693793
Weird Paradox behaviour

### DIFF
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -862,7 +862,10 @@ bool Dataloader::_load_sound_effect_defines(DefinitionManager& definition_manage
 
 }
 
-bool Dataloader::load_defines(DefinitionManager& definition_manager) {
+bool Dataloader::load_defines(
+	GameRulesManager const& game_rules_manager,
+	DefinitionManager& definition_manager
+) {
 	if (roots.empty()) {
 		Logger::error("Cannot load defines - Dataloader has no roots!");
 		return false;
@@ -942,7 +945,9 @@ bool Dataloader::load_defines(DefinitionManager& definition_manager) {
 		Logger::error("Failed to load pop types!");
 		ret = false;
 	}
-	if (!definition_manager.get_economy_manager().load_production_types_file(definition_manager.get_pop_manager(),
+	if (!definition_manager.get_economy_manager().load_production_types_file(
+		game_rules_manager,
+		definition_manager.get_pop_manager(),
 		parse_defines_cached(lookup_file(production_types_file))
 	)) {
 		Logger::error("Failed to load production types!");
--- a/src/openvic-simulation/dataloader/Dataloader.hpp
+++ b/src/openvic-simulation/dataloader/Dataloader.hpp
@@ -9,6 +9,7 @@ namespace OpenVic {
 	namespace fs = std::filesystem;
 
 	struct DefinitionManager;
+	struct GameRulesManager;
 	class UIManager;
 
 	template<typename _UniqueFileKey>
@@ -109,7 +110,7 @@ namespace OpenVic {
 
 		/* Load and parse all of the text defines data, including parsing cached condition and effect scripts after all the
 		 * static data is loaded. Paths to the base and mod defines must have been supplied with set_roots.*/
-		bool load_defines(DefinitionManager& definition_manager);
+		bool load_defines(GameRulesManager const& game_rules_manager, DefinitionManager& definition_manager);
 
 	private:
 		/* Parse the cached Nodes of every condition and effect script in the defines.
--- a/src/openvic-simulation/economy/production/ProductionType.cpp
+++ b/src/openvic-simulation/economy/production/ProductionType.cpp
@@ -16,6 +16,7 @@ Job::Job(
 	amount { new_amount } {}
 
 ProductionType::ProductionType(
+	GameRulesManager const& new_game_rules_manager,
 	const std::string_view new_identifier,
 	const std::optional<Job> new_owner,
 	std::vector<Job>&& new_jobs,
@@ -30,6 +31,7 @@ ProductionType::ProductionType(
 	const bool new_is_farm,
 	const bool new_is_mine
 ) : HasIdentifier { new_identifier },
+	game_rules_manager { new_game_rules_manager },
 	owner { new_owner },
 	jobs { std::move(new_jobs) },
 	template_type { new_template_type },
@@ -40,8 +42,8 @@ ProductionType::ProductionType(
 	bonuses { std::move(new_bonuses) },
 	maintenance_requirements { std::move(new_maintenance_requirements) },
 	coastal { new_is_coastal },
-	farm { new_is_farm },
-	mine { new_is_mine } {}
+	_is_farm { new_is_farm },
+	_is_mine { new_is_mine } {}
 
 bool ProductionType::parse_scripts(DefinitionManager const& definition_manager) {
 	bool ret = true;
@@ -94,6 +96,7 @@ node_callback_t ProductionTypeManager::_expect_job_list(
 }
 
 bool ProductionTypeManager::add_production_type(
+	GameRulesManager const& game_rules_manager,
 	const std::string_view identifier,
 	std::optional<Job> owner,
 	std::vector<Job>&& jobs,
@@ -169,6 +172,7 @@ bool ProductionTypeManager::add_production_type(
 	}
 
 	const bool ret = production_types.add_item({
+		game_rules_manager,
 		identifier, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), *output_good,
 		base_output_quantity, std::move(bonuses), std::move(maintenance_requirements), is_coastal, is_farm, is_mine
 	});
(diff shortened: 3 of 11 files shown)


src/openvic-simulation/GameRulesManager.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "openvic-simulation/utility/Getters.hpp"
+namespace OpenVic {
+	struct GameRulesManager {
+	private:
+		bool PROPERTY_RW(use_legacy_farm_mine_logic);
+		bool PROPERTY_RW(use_legacy_ui);
```
Hop311, line 8:
```suggestion
		bool PROPERTY_RW(use_legacy_farm_mine_logic, true);
		bool PROPERTY_RW(use_legacy_ui, true);
```
- You need to explicitly initialise the bools, otherwise they'll have random values with MSVC. If you rebase onto master you can use the new `PROPERTY` initialisation value feature Spartan added, as shown above (adding the starting value as another argument in the macro).
- Using the `PROPERTY` initialisation means we won't need a constructor, so you can remove `GameRulesManager();` from this file and delete `GameRulesManager.cpp` altogether. Also maybe move this to the `misc` folder? It fits with general infrastructure stuff like `SimulationClock` in there.
- I'm not sure what `use_legacy_ui` is meant to be for, I expect we'd use rules for specific features rather than general "legacy UI" vs "new UI"
- IMO the rules should enable new/divergent features, rather than enabling legacy stuff. I expect there'll be rules with more options than just true/false, which means there won't be a single non-legacy baseline. Better to have "faithful" as the baseline/default and have rules to enable specific divergent behaviour.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -10,6 +10,7 @@
 #include "openvic-simulation/DefinitionManager.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
 #include "openvic-simulation/utility/StringUtils.hpp"
+#include "GameRulesManager.hpp"
```
Hop311, line 13:
```suggestion
#include "openvic-simulation/GameRulesManager.hpp"
```
This include isn't really needed anyway, I suspect clangd added it automatically?

src/openvic-simulation/dataloader/Dataloader.hpp
```diff
@@ -3,6 +3,7 @@
 #include <openvic-dataloader/csv/Parser.hpp>
 #include <openvic-dataloader/v2script/Parser.hpp>
 
+#include "openvic-simulation/GameRulesManager.hpp"
```
Hop311, line 6: Rather than including it here, you could forward declare it at the top of the namespace in this file?

src/openvic-simulation/economy/production/ProductionType.cpp
```diff
@@ -40,8 +42,29 @@ ProductionType::ProductionType(
...
+bool ProductionType::get_is_mine_for_tech() const {
+	return _is_mine;
+}
+bool ProductionType::get_is_mine_for_non_tech() const {
+	if (!game_rules_manager.get_use_legacy_farm_mine_logic()) {
+		return _is_mine;
+	}
+
+	return !_is_farm;
+}
```
Hop311, line 67: Move these to the header file and make them `constexpr`. Also IMO it reads better to avoid the `!`, so something like:
```cpp
bool ProductionType::get_is_mine_for_non_tech() const {
	if (game_rules_manager.get_use_legacy_farm_mine_logic()) {
		return !_is_farm;
	} else {
		return _is_mine;
	}
}
```

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -105,12 +105,20 @@ fixed_point_t ResourceGatheringOperation::calculate_size_modifier() const {
...
+	//TODO WVPM split rgo size modifiers based on tech/non-tech
 	fixed_point_t size_modifier = fixed_point_t::_1();
-	if (production_type.is_farm()) {
+	if (production_type.get_is_farm_for_tech()) {
+		size_modifier += location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_size_global());
+	}
+	if (production_type.get_is_farm_for_non_tech()) {
 		size_modifier += location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_size_global())
 			+ location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_size_local());
 	}
```
Hop311, line 116:
```suggestion
	if (production_type.get_is_farm_for_tech()) {
		size_modifier += location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_size_global());
	}
	if (production_type.get_is_farm_for_non_tech()) {
		size_modifier += location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_size_local());
	}
```
- For the non-tech case you only want to be adding the local modifier, otherwise you'll end up double counting it if both tech and non-tech farm is true.
- Same for the mine modifier effects below.
- Double check exactly which modifier effects are used for `base_size_modifier` in `initialise_rgo_size_multiplier` and for `size_modifier` here, in all 4 main cases: nothing, just `farm = yes`, just `mine = yes`, and both `farm = yes` and `mine = yes`.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -257,13 +265,19 @@ fixed_point_t ResourceGatheringOperation::produce(const pop_size_t total_owner_c
...
 
-	if (production_type.is_farm()) {
+	//TODO WVPM split rgo eff modifiers based on tech/non-tech
+	if (production_type.get_is_farm_for_tech()) {
 		throughput_multiplier += location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_throughput_global());
+	}
+	if (production_type.get_is_farm_for_tech()) {
 		output_multilpier += location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_output_global())
 			+ location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_output_local());
 	}
```
Hop311, line 275: - Both if statements have `get_is_farm_for_tech`, one of them should be `get_is_farm_for_non_tech`.
- Once the effects are changed to be per-source:
  - the `get_is_farm_for_tech` section will lookup the throughput & output effect, store it in a temporary variable, and then add it to both `throughput_multiplier` and `output_multiplier`.
  - the `get_is_farm_for_non_tech` section will lookup both the local and global output effects and add them to `output_multilpier`.
- This is all repeated for the mine cases below

Hop311 (commented):
[I wrote a PR](https://github.com/OpenVicProject/OpenVic-Simulation/pull/269) changing how `farm_rgo_eff` is converted into modifier effects earlier today, if you approve that then you could rebase this branch onto master after it's merged and all that'll be left to do is splitting the related RGO code into tech and non_tech sections

src/openvic-simulation/economy/production/ProductionType.cpp
```diff
@@ -169,14 +172,15 @@ bool ProductionTypeManager::add_production_type(
...
+		game_rules_manager,
 		identifier, owner, std::move(jobs), template_type, base_workforce_size, std::move(input_goods), *output_good,
 		base_output_quantity, std::move(bonuses), std::move(maintenance_requirements), is_coastal, is_farm, is_mine
 	});
 
 	if (ret && (template_type == RGO)) {
 		ProductionType const& production_type = production_types.get_items().back();
 		ProductionType const*& current_rgo_pt = good_to_rgo_production_type[*output_good];
-		if (current_rgo_pt == nullptr || (is_farm && !current_rgo_pt->is_farm())) {
+		if (current_rgo_pt == nullptr || (is_farm && !current_rgo_pt->_is_farm)) {
```
Hop311, line 183: Not urgent to change, but are we sure this doesn't also have weird behaviour for non-standard farm/mine settings? Currently we're bascially following `get_is_farm_for_non_tech`

src/openvic-simulation/economy/production/ProductionType.hpp
```diff
@@ -76,6 +77,26 @@ namespace OpenVic {
 		bool parse_scripts(DefinitionManager const& definition_manager);
 
 	public:
+		constexpr bool get_is_farm_for_tech() const {
+			if (game_rules_manager.get_use_simple_farm_mine_logic()) {
+				return _is_farm;
+			}
+
+			return !_is_mine && _is_farm;
```
Hop311, line 85:
```suggestion
			return !_is_mine && _is_farm;
```
Stray tab character at the end of this line

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -79,10 +79,11 @@ void ResourceGatheringOperation::initialise_rgo_size_multiplier() {
...
 
 	fixed_point_t base_size_modifier = fixed_point_t::_1();
-	if (production_type.is_farm()) {
+	if (production_type.get_is_farm_for_non_tech()) {
 		base_size_modifier += location.get_modifier_effect_value_nullcheck(modifier_effect_cache.get_farm_rgo_size_local());
 	}
-	if (production_type.is_mine()) {
+
```
Hop311, line 85:
```suggestion

```
Stray tab characters in this file on lines: 85, 116, 124, 268.

Hop311 (approved):

PR #275, opened 2025-01-27
Dynamic pricing

wvpm:
Implementation of the python code. See also https://discord.com/channels/1063392556160909312/1105422811809452072/1301900686904000527

### DIFF
--- a/src/openvic-simulation/economy/GoodInstance.cpp
+++ b/src/openvic-simulation/economy/GoodInstance.cpp
@@ -1,4 +1,5 @@
 #include "GoodInstance.hpp"
+#include "openvic-simulation/utility/CompilerFeatureTesting.hpp"
 
 using namespace OpenVic;
 
@@ -12,38 +13,41 @@ GoodInstance::GoodInstance(GoodDefinition const& new_good_definition, GameRulesM
 	game_rules_manager { new_game_rules_manager },
 	price { new_good_definition.get_base_price() },
 	is_available { new_good_definition.get_is_available_from_start() },
-	price_history { MONTHS_OF_PRICE_HISTORY, new_good_definition.get_base_price() } {
+	price_history { MONTHS_OF_PRICE_HISTORY, new_good_definition.get_base_price() }
+	{
+		on_use_exponential_price_changes_changed();
+		update_next_price_limits();
+	}
 
-	update_next_price_limits();
+void GoodInstance::on_use_exponential_price_changes_changed() {
+	if (game_rules_manager.get_use_exponential_price_changes()) {
+		absolute_maximum_price = fixed_point_t::usable_max();
+		absolute_minimum_price = fixed_point_t::epsilon() << exponential_price_change_shift;
+	} else {
+		absolute_maximum_price = std::min(
+			good_definition.get_base_price() * 5,
+			fixed_point_t::usable_max()
+		);
+		absolute_minimum_price = std::max(
+			good_definition.get_base_price() * 22 / 100,
+			fixed_point_t::epsilon()
+		);
+	}
 }
 
 void GoodInstance::update_next_price_limits() {
-	if(game_rules_manager.get_use_exponential_price_changes()) {
-		const fixed_point_t max_change = price >> 6;
-		max_next_price = std::min(
-			fixed_point_t::usable_max(),
-			price + max_change
-		);
-		min_next_price = std::max(
-			fixed_point_t::epsilon(),
-			price - max_change
-		);
-	} else {
-		max_next_price = std::min(
-			std::min(
-				good_definition.get_base_price() * 5,
-				fixed_point_t::usable_max()
-			),
-			price + fixed_point_t::_1() / fixed_point_t::_100()
-		);
-		min_next_price = std::max(
-			std::max(
-				good_definition.get_base_price() * 22 / fixed_point_t::_100(),
-				fixed_point_t::epsilon()
-			),
-			price - fixed_point_t::_1() / fixed_point_t::_100()
-		);
-	}
+	const fixed_point_t max_price_change = game_rules_manager.get_use_exponential_price_changes()
+		? price >> exponential_price_change_shift
+		: fixed_point_t::_0_01();
+
+	max_next_price = std::min(
+		absolute_maximum_price,
+		price + max_price_change
+	);
+	min_next_price = std::max(
+		absolute_minimum_price,
+		price - max_price_change
+	);
 }
 
 void GoodInstance::add_buy_up_to_order(GoodBuyUpToOrder&& buy_up_to_order) {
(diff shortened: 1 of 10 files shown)


src/openvic-simulation/economy/GoodInstance.cpp
```diff
@@ -18,37 +18,42 @@ GoodInstance::GoodInstance(GoodDefinition const& new_good_definition, GameRulesM
...
+		absolute_maximum_price,
+		price + max_price_change
+	);
+	min_next_price = std::max(
+		absolute_minimum_price,
+		price - max_price_change
+	);
+}
+
 void GoodInstance::add_buy_up_to_order(GoodBuyUpToOrder&& buy_up_to_order) {
```
schombert, line 57: Just FYI, prices in Victoria 2 aren't stable when supply meets demand, as this class would seem to imply. Although, since this class represents a non-trivial departure from Victoria 2 mechanics (for example, there is only one global price but many different supply pools to buy from -- local, sphere market, and global, so this class can neither represent a good in a single supply pool, since each pool would then have its own price, nor can it represent the total state of the good since it has only a single supply), maybe that difference was intended.

Anyways, the rule for how supply and demand affect prices in Victoria 2 is this:
Prices move up or down by 0.01 a day to a limit of 5x or one fifth the base price in order to track the square root of (demand / supply) times the base price of the commodity. Thus, when supply exactly meets demand the price trends towards the base price of the good; it is not stable at that ratio.

You may be thinking to yourself: "that doesn't make a lick of sense". And I have to agree with you there, but that is simply how the Victoria 2 economy works, and if you want to deliver the vanilla experience, that is what you have to emulate.

wvpm: True, the code doesn't match Victoria 2 yet. That depends on https://github.com/OpenVicProject/OpenVic/issues/288
As for a minimum of 1/5th. I've read that but couldn't replicate it. 0.22x base price was the lowest I could achieve by cheating a stockpile and dumping it on the market.

Anyway, this PR is not about matching Victoria 2 mechanics, it's about the OpenVic dynamic price mechanic.

schombert: Ah, so you are going to have some other class that implements the tracking towards the square root of the ratio? I did not realize that.

wvpm: There were bugs in the old code so I rewrote it and added TODO's for the square root ratio.

src/openvic-simulation/economy/GoodInstance.cpp
```diff
@@ -57,53 +61,225 @@ void GoodInstance::add_market_sell_order(GoodMarketSellOrder&& market_sell_order
...
+
+			quantity_traded_yesterday = fixed_point_t::_0();
+			//figure out how much every buyer bought
+			for (int i = 0; i < buy_up_to_orders.size(); i++) {
+				GoodBuyUpToOrder const& buy_up_to_order = buy_up_to_orders[i];
+				const fixed_point_t quantity_bought
+					= quantity_bought_per_order[i]
+					= std::min(
+					buy_up_to_order.get_max_quantity(),
+					buy_up_to_order.get_money_to_spend() / new_price
```
Hop311, line 253: Are we absolutely sure `new_price` cannot be 0 here?

src/openvic-simulation/economy/GoodInstance.cpp
```diff
@@ -57,53 +61,225 @@ void GoodInstance::add_market_sell_order(GoodMarketSellOrder&& market_sell_order
...
+				quantity_traded_yesterday += quantity_bought;
+				buy_up_to_order.get_after_trade()({
+					quantity_bought,
+					buy_up_to_order.get_money_to_spend() - quantity_bought * new_price
+				});
+			}
+		}
+
+		for (GoodMarketSellOrder const& market_sell_order : market_sell_orders) {
+			const fixed_point_t quantity_sold = market_sell_order.get_quantity() * quantity_traded_yesterday / supply_sum;
```
Hop311, line 265: Are we absolutely sure `supply_sum` cannot be 0 here?

Hop311 (commented):
Also we still have massive log spam from the macros in Pop.cpp, could you suppress those, whether by preventing the bad values from reaching there, removing the log messages altogether (still returning if necessary), or providing a way to optionally turn them off. Currently I have to manually edit them out whenever I want to test something with time passing

src/openvic-simulation/economy/GoodInstance.cpp
```diff
@@ -57,53 +61,225 @@ void GoodInstance::add_market_sell_order(GoodMarketSellOrder&& market_sell_order
...
+			if (game_rules_manager.get_use_optimal_pricing()) {
+				const fixed_point_t affordable_price = buy_up_to_order.get_affordable_price();
+				if (affordable_price > min_next_price) {
+					//no point selling lower as it would not attract more buyers
+					min_next_price = affordable_price;
+				}
+			}
+
+			demand_sum += buy_up_to_order.get_max_quantity();
+			const fixed_point_t purchasing_power = purchasing_power_per_order[i] = buy_up_to_order.get_money_to_spend() / max_next_price;
```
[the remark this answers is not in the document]
wvpm: Yes.
```cpp
max_next_price = std::min(
	absolute_maximum_price,
	price + max_price_change
);
```

```cpp
if (game_rules_manager.get_use_exponential_price_changes()) {
	absolute_maximum_price = fixed_point_t::usable_max();
	absolute_minimum_price = fixed_point_t::epsilon() << exponential_price_change_shift;
} else {
	absolute_maximum_price = std::min(
		good_definition.get_base_price() * 5,
		fixed_point_t::usable_max()
	);
	absolute_minimum_price = std::max(
		good_definition.get_base_price() * 22 / 100,
		fixed_point_t::epsilon()
	);
}
```

It could only be 0 if `base price == 0 && !game_rules_manager.get_use_exponential_price_changes()`.
Base price of 0 is invalid and handled during the dataloading.

src/openvic-simulation/economy/GoodInstance.cpp
wvpm: Yes. All orders must have a positive quanity.
`if (market_sell_orders.empty()) {` handles the case where there are no sell orders.

src/openvic-simulation/economy/GoodInstance.cpp
wvpm: Yes, `new_price >= min_next_price`
```cpp
	min_next_price = std::max(
		absolute_minimum_price,
		price - max_price_change
	);
```

```cpp
		absolute_minimum_price = std::max(
			good_definition.get_base_price() * 22 / 100,
			fixed_point_t::epsilon()
		);
```

Hop311 (approved):
Price changes now successfully track in the trade menu, although they'll be crashing to the minimum price until we get more actors involved for some more interesting behaviour

PR #276, opened 2025-01-27
Add PROPERTY_PTR with non-const pointer getter + cleanup constructors

Hop311:


src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -192,46 +192,46 @@ namespace OpenVic {
...
+		bool PROPERTY_CUSTOM_PREFIX(mobilised, is, false);
+		bool PROPERTY_CUSTOM_PREFIX(disarmed, is, false);
+		bool PROPERTY_RW(auto_create_leaders, true);
+		bool PROPERTY_RW(auto_assign_leaders, true);
 		fixed_point_t PROPERTY(organisation_regain);
 		fixed_point_t PROPERTY(land_organisation);
 		fixed_point_t PROPERTY(naval_organisation);
 		fixed_point_t PROPERTY(land_unit_start_experience);
 		fixed_point_t PROPERTY(naval_unit_start_experience);
 		fixed_point_t PROPERTY(recruit_time);
```
wvpm, line 226: Shouldn't this be initialised to 1  (as previously)?

wvpm (approved):

src/openvic-simulation/country/CountryInstance.hpp
Hop311: I noticed this too but decided against it as we shouldn't be using those values before they're first set in update_gamestate

PR #281, opened 2025-01-28
Add country research and leadership point gain + tech research

Hop311:


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -704,6 +704,43 @@ bool CountryInstance::is_primary_or_accepted_culture(Culture const& culture) con
...
+}
+
+fixed_point_t CountryInstance::get_research_progress() const {
+	return current_research_cost > fixed_point_t::_0()
+		? invested_research_points / current_research_cost
+		: fixed_point_t::_0();
+}
+
+bool CountryInstance::can_research_tech(Technology const& technology, Date today) const {
+	return technology.get_year() <= today.get_year() && is_civilised() && !is_technology_unlocked(technology) &&
```
wvpm, line 723: Split up large boolean chains with named variables.
`(technology.get_index_in_area() == 0 || is_technology_unlocked(*technology.get_area().get_technologies([technology.get_index_in_area() - 1]))`
Could be named `is_next_step_in_area`.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -704,6 +704,43 @@ bool CountryInstance::is_primary_or_accepted_culture(Culture const& culture) con
...
+		&technology != current_research && (technology.get_index_in_area() == 0 || is_technology_unlocked(
+			*technology.get_area().get_technologies()[technology.get_index_in_area() - 1]
+		));
+}
+
+void CountryInstance::start_research(Technology const& technology, InstanceManager const& instance_manager) {
+	if (!can_research_tech(technology, instance_manager.get_today())) {
```
wvpm, line 730: You can wrap such assert ifs in `OV_unlikely(condition)`. Spartan pointed out this has performance benefit.

src/openvic-simulation/country/CountryInstance.cpp
Hop311: I might add a temporary variable for `technology.get_index_in_area()` as it's used twice, but creating variables for things which are only used once has a performance impact that is undesirable particularly in functions like this one which will be called very often.

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -272,7 +272,7 @@ namespace OpenVic {
 			return stream;
 		}
 
-		std::string to_string(size_t decimal_places = 0) const {
+		std::string to_string(size_t decimal_places = -1) const {
```
wvpm, line 275: Why -1?
I don't even understand what displaying -1 decimals means.
Would that be value - value % 10?

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -704,6 +704,49 @@ bool CountryInstance::is_primary_or_accepted_culture(Culture const& culture) con
...
+
+fixed_point_t CountryInstance::get_research_progress() const {
+	return current_research_cost > fixed_point_t::_0()
+		? invested_research_points / current_research_cost
+		: fixed_point_t::_0();
+}
+
+bool CountryInstance::can_research_tech(Technology const& technology, Date today) const {
+	if (
+		technology.get_year() > today.get_year() || !is_civilised() || is_technology_unlocked(technology) ||
```
wvpm, line 724: format

wvpm (dismissed):

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
Hop311: - `-1` means use as many decimals as it takes to display the entire value, so anything from none (e.g. `1`) to whatever the smallest represeantable non-zero value is (e.g. `1.1111111...`).
- `0` means no decimal places, so just a whole number with no point at all
- `n > 0` means exactly that many decimal places, including trailing zeroes if necessary

BrickPi (approved):

PR #288, opened 2025-02-02
Add trade data, good enabling, price history

Hop311:


src/openvic-simulation/types/ValueHistory.hpp
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		T get_total() const {
+			return std::accumulate(begin(), end(), T {});
+		}
+
+		void push_back(T const& value) {
+			if (!empty()) {
+				// Move all values back by one (with the first value being discarded) and place the new value at the end
+				for (iterator it = begin(); it != end(); ++it) {
```
BrickPi, line 52: Could we use `std::rotate` here?

src/openvic-simulation/types/ValueHistory.hpp
```diff
@@ -0,0 +1,102 @@
...
+				back() = value;
+			}
+		}
+
+		void set_size(size_t new_size, T const& fill_value = {}) {
+			const difference_type size_diff = new_size - size();
+
+			if (size_diff < 0) {
+				// Move the last new_size elements to the front, discarding the values before them
+				for (iterator it = begin(); it != begin() + new_size; ++it) {
```
BrickPi, line 64: `std::rotate` here too?

src/openvic-simulation/types/ValueHistory.hpp
Hop311: That's a nice STL algorithm, if I'd known about it before I probably would've used it yeah, but since I've already worked out the minimal logic it's better to stick with it rather than using something which does more than we need (we don't care about copying values that get pushed out back to the other end of the vector)

src/openvic-simulation/InstanceManager.cpp
```diff
@@ -69,7 +70,7 @@ void InstanceManager::tick() {
 	map_instance.map_tick(today);
 	country_instance_manager.tick(*this);
 	unit_instance_manager.tick();
-	market_instance.execute_orders();
+	market_instance.execute_orders(today.is_month_start());
```
wvpm, line 73: Don't mix this in with executing orders please.
It is not related to that at all.

Instead just do the check after the market.
```cpp
if (today.is_month_start()) {
   market_instance.record_prices();
}
```

wvpm (changes requested):
Split price history recording to separate method.

src/openvic-simulation/types/ValueHistory.hpp
```diff
@@ -0,0 +1,102 @@
...
+				}
+				for (iterator it = begin(); it != begin() + size_diff; ++it) {
+					*it = fill_value;
+				}
+			}
+		}
+	};
+
+	template<typename T>
+	std::ostream& operator<<(std::ostream& stream, ValueHistory<T> const& history) {
```
wvpm, line 85: Is this debug printing?

Hop311: Yes, this is for logging, not for the actual game UI

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -186,7 +188,29 @@ namespace OpenVic {
...
 
 		/* Trade */
-		// TODO - total amount of each good exported and imported
+	public:
+		struct good_data_t {
+			fixed_point_t stockpile_amount;
+			fixed_point_t stockpile_change_yesterday; // positive if we bought, negative if we sold
+
+			bool is_automated = true;
+			bool is_selling = false; // buying if false
```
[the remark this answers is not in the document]
Hop311: Sure, or just have named constants of some kind, I'm gonna leave it for now and we can decide how best to represent it when we actually implement this feature

src/openvic-simulation/types/ValueHistory.hpp
wvpm: I agree with BrickPi std::rotate is perfect for this.
Still you shouldn't rotate every time you add a value.
You simply rotate just before reading all values.

src/openvic-simulation/types/ValueHistory.hpp
```diff
@@ -0,0 +1,102 @@
...
+			if (!empty()) {
+				// Move all values back by one (with the first value being discarded) and place the new value at the end
+				for (iterator it = begin(); it != end(); ++it) {
+					*it = *(it + 1);
+				}
+				back() = value;
+			}
+		}
+
+		void set_size(size_t new_size, T const& fill_value = {}) {
```
wvpm, line 59: Making size mutable complicates things significantly and without real benefit.
We should never have to resize it. The capacity is known before hand.

For the ledger, you could calculate the time between startdate & enddate. For other cases we have hardcoded length (36 monthly entries).

src/openvic-simulation/economy/GoodInstance.cpp
```diff
@@ -109,9 +106,18 @@ void GoodInstance::execute_orders() {
 	if (new_price != price) {
 		update_next_price_limits();
 	}
+
+	price = new_price;
 }
 
-bool GoodInstanceManager::setup_goods(GoodDefinitionManager const& good_definition_manager, GameRulesManager const& game_rules_manager) {
+void GoodInstance::record_price_history() {
```
wvpm, line 113: I love this simplicity.

wvpm (approved):

PR #291, opened 2025-02-03
Mod Compatibility Fixes

BrickPi:
- Implements a "ModManager" which loads .mod files
- Implements `replace_path` when searching for files
- Myriad changes to better reflect how V2 handles systems
- HPM now successfully loads and runs under OpenVic

### DIFF
--- a/src/headless/main.cpp
+++ b/src/headless/main.cpp
@@ -1,5 +1,6 @@
 #include <openvic-simulation/country/CountryInstance.hpp>
 #include <openvic-simulation/dataloader/Dataloader.hpp>
+#include <openvic-simulation/dataloader/ModManager.hpp>
 #include <openvic-simulation/economy/GoodDefinition.hpp>
 #include <openvic-simulation/economy/production/ProductionType.hpp>
 #include <openvic-simulation/economy/production/ResourceGatheringOperation.hpp>
@@ -17,7 +18,7 @@ static void print_help(std::ostream& stream, char const* program_name) {
 		<< "    -t : Run tests after loading defines.\n"
 		<< "    -b : Use the following path as the base directory (instead of searching for one).\n"
 		<< "    -s : Use the following path as a hint to search for a base directory.\n"
-		<< "Any following paths are read as mod directories, with priority starting at one above the base directory.\n"
+		<< "Any following paths are read as mods (/path/to/my/MODNAME.mod), with priority starting at one above the base directory.\n"
 		<< "(Paths with spaces need to be enclosed in \"quotes\").\n";
 }
 
@@ -55,15 +56,28 @@ static void print_rgo(ProvinceInstance const& province) {
 	}
 }
 
-static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests) {
+static bool run_headless(const fs::path root, std::vector<std::string> mods, bool run_tests) {
 	bool ret = true;
+	Dataloader::path_vector_t roots = { root };
+	Dataloader::path_vector_t replace_paths = {};
 
 	GameManager game_manager { []() {
 		Logger::info("State updated");
 	}, nullptr };
 
+	Logger::info("===== Loading mod descriptors... =====");
+	ret &= game_manager.load_mod_descriptors(std::move(mods));
+	for (auto const& mod : game_manager.get_mod_manager().get_mods()) {
+		roots.emplace_back(root / mod.get_dataloader_root_path());
+		for (std::string_view path : mod.get_replace_paths()) {
+			if (std::find(replace_paths.begin(), replace_paths.end(), path) == replace_paths.end()) {
+				replace_paths.emplace_back(path);
+			} 
+		}
+	}
+
 	Logger::info("===== Loading definitions... =====");
-	ret &= game_manager.set_roots(roots);
+	ret &= game_manager.set_roots(roots, replace_paths);
 	ret &= game_manager.load_definitions(
 		[](std::string_view key, Dataloader::locale_t locale, std::string_view localisation) -> bool {
 			return true;
@@ -141,6 +155,7 @@ int main(int argc, char const* argv[]) {
 
 	char const* program_name = StringUtils::get_filename(argc > 0 ? argv[0] : nullptr, "<program>");
 	fs::path root;
+	std::vector<std::string> mods;
 	bool run_tests = false;
 	int argn = 0;
 
@@ -196,15 +211,13 @@ int main(int argc, char const* argv[]) {
 			return -1;
 		}
 	}
-	Dataloader::path_vector_t roots { root };
 	while (argn < argc) {
-		static const fs::path mod_directory = "mod";
-		roots.emplace_back(root / mod_directory / argv[argn++]);
+		mods.emplace_back(argv[argn++]);
 	}
 
 	std::cout << "!!! HEADLESS SIMULATION START !!!" << std::endl;
 
-	const bool ret = run_headless(roots, run_tests);
+	const bool ret = run_headless(root, std::move(mods), run_tests);
 
 	std::cout << "!!! HEADLESS SIMULATION END !!!" << std::endl;
 
--- a/src/openvic-simulation/GameManager.cpp
+++ b/src/openvic-simulation/GameManager.cpp
@@ -9,10 +9,23 @@ GameManager::GameManager(
 		new_gamestate_updated_callback ? std::move(new_gamestate_updated_callback) : []() {}
 	}, clock_state_changed_callback {
 		new_clock_state_changed_callback ? std::move(new_clock_state_changed_callback) : []() {}
-	}, definitions_loaded { false } {}
+	}, definitions_loaded { false }, mod_descriptors_loaded { false } {}
 
-bool GameManager::set_roots(Dataloader::path_vector_t const& roots) {
-	if (!dataloader.set_roots(roots)) {
+bool GameManager::load_mod_descriptors(std::vector<std::string> descriptors) {
+	if (mod_descriptors_loaded) {
+		Logger::error("Cannot load mod descriptors - already loaded!");
+		return false;
+	}
+
+	if (!dataloader.load_mod_descriptors(std::move(descriptors), mod_manager)) {
+		Logger::error("Failed to load mod descriptors!");
+		return false;
+	}
+	return true;
+}
+
+bool GameManager::set_roots(Dataloader::path_vector_t const& roots, Dataloader::path_vector_t const& replace_paths) {
+	if (!dataloader.set_roots(roots, replace_paths)) {
 		Logger::error("Failed to set dataloader roots!");
 		return false;
 	}
--- a/src/openvic-simulation/country/CountryDefinition.cpp
+++ b/src/openvic-simulation/country/CountryDefinition.cpp
@@ -11,6 +11,7 @@
 #include "openvic-simulation/pop/Culture.hpp"
 #include "openvic-simulation/types/Colour.hpp"
 #include "openvic-simulation/types/IdentifierRegistry.hpp"
+#include "utility/Logger.hpp"
 #include <openvic-dataloader/v2script/AbstractSyntaxTree.hpp>
 
 using namespace OpenVic;
@@ -129,13 +130,19 @@ bool CountryDefinitionManager::load_countries(
 }
 
 bool CountryDefinitionManager::load_country_colours(ast::NodeCPtr root) {
-	return country_definitions.expect_item_dictionary([](CountryDefinition& country, ast::NodeCPtr colour_node) -> bool {
-		return expect_dictionary_keys(
-			"color1", ONE_EXACTLY, expect_colour(assign_variable_callback(country.primary_unit_colour)),
-			"color2", ONE_EXACTLY, expect_colour(assign_variable_callback(country.secondary_unit_colour)),
-			"color3", ONE_EXACTLY, expect_colour(assign_variable_callback(country.tertiary_unit_colour))
-		)(colour_node);
-	})(root);
+	return country_definitions.expect_item_dictionary_and_default(
+		[](std::string_view key, ast::NodeCPtr value) -> bool {
+			Logger::warning("country_colors.txt references country tag ", key, " which is not defined!");
+			return true;
+		},
+		[](CountryDefinition& country, ast::NodeCPtr colour_node) -> bool {
+			return expect_dictionary_keys(
+				"color1", ONE_EXACTLY, expect_colour(assign_variable_callback(country.primary_unit_colour)),
+				"color2", ONE_EXACTLY, expect_colour(assign_variable_callback(country.secondary_unit_colour)),
+				"color3", ONE_EXACTLY, expect_colour(assign_variable_callback(country.tertiary_unit_colour))
+			)(colour_node);
+		}
+	)(root);
 }
 
 node_callback_t CountryDefinitionManager::load_country_party(
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -30,12 +30,17 @@ static fs::path ensure_forward_slash_path(std::string_view path) {
 #endif
 }
 
-bool Dataloader::set_roots(path_vector_t const& new_roots) {
+bool Dataloader::set_roots(path_vector_t const& new_roots, path_vector_t const& new_replace_paths) {
 	if (!roots.empty()) {
 		Logger::warning("Overriding existing dataloader roots!");
 		roots.clear();
+		replace_paths.clear();
 	}
 	bool ret = true;
+	for (const fs::path& replace_path : new_replace_paths) {
+		Logger::info("Adding replace path: ", replace_path);
+		replace_paths.push_back(replace_path);
+	}
 	for (std::reverse_iterator<path_vector_t::const_iterator> it = new_roots.crbegin(); it != new_roots.crend(); ++it) {
 		if (std::find(roots.begin(), roots.end(), *it) == roots.end()) {
 			if (fs::is_directory(*it)) {
@@ -60,23 +65,72 @@ bool Dataloader::set_roots(path_vector_t const& new_roots) {
 fs::path Dataloader::lookup_file(std::string_view path, bool print_error) const {
 	const fs::path filepath { ensure_forward_slash_path(path) };
 
+#if defined(FILESYSTEM_CASE_INSENSITIVE)
+	/* Case-insensitive filesystem */
+	for (fs::path const& root : roots) {
+		const fs::path composed = root / filepath;
+		if (fs::is_regular_file(composed)) {
+			if (root == roots.back()) {
+				bool ignore = false;
+				for (auto const& replace_path : replace_paths) {
+					if (filepath.string().starts_with(replace_path.string())) {
+						ignore = true;
+						break;
+					}
+				}
+				if (!ignore) {
+					return composed;
+				}
+			} else {
+				return composed;
+			}
+		}
+	}
+#else
 	/* Case-sensitive filesystem */
 	const std::string_view filename = StringUtils::get_filename(path);
 	for (fs::path const& root : roots) {
 		const fs::path composed = root / filepath;
 		if (fs::is_regular_file(composed)) {
-			return composed;
+			if (root == roots.back()) {
+				bool ignore = false;
+				for (auto const& replace_path : replace_paths) {
+					if (filepath.string().starts_with(replace_path.string())) {
+						ignore = true;
+						break;
+					}
+				}
+				if (!ignore) {
+					return composed;
+				}
+			} else {
+				return composed;
+			}
 		}
 		std::error_code ec;
 		for (fs::directory_entry const& entry : fs::directory_iterator { composed.parent_path(), ec }) {
 			if (entry.is_regular_file()) {
 				const fs::path file = entry;
 				if (StringUtils::strings_equal_case_insensitive(file.filename().string(), filename)) {
-					return file;
+					if (root == roots.back()) {
+						bool ignore = false;
+						for (auto const& replace_path : replace_paths) {
+							if (filepath.string().starts_with(replace_path.string())) {
+								ignore = true;
+								break;
+							}
+						}
+						if (!ignore) {
+							return file;
+						}
+					} else {
+						return file;
+					}
 				}
 			}
 		}
 	}
+#endif
 
 	if (print_error) {
 		Logger::error("Lookup for \"", path, "\" failed!");
@@ -109,6 +163,18 @@ Dataloader::path_vector_t Dataloader::_lookup_files_in_dir(
 	for (fs::path const& root : roots) {
 		const size_t root_len = root.string().size();
 		std::error_code ec;
+		if (root == roots.back()) {
+			bool ignore = false;
+			for (auto const& replace_path : replace_paths) {
+				if (dirpath.string().starts_with(replace_path.string())) {
+					ignore = true;
+					break;
+				}
+			}
+			if (ignore) {
+				continue;
+			}
+		}
 		for (fs::directory_entry const& entry : _DirIterator { root / dirpath, ec }) {
 			if (entry.is_regular_file()) {
 				fs::path file = entry;
@@ -255,6 +321,21 @@ v2script::Parser& Dataloader::parse_defines_cached(fs::path const& path) {
 
 void Dataloader::free_cache() {
 	cached_parsers.clear();
+}
+
+bool Dataloader::load_mod_descriptors(std::vector<std::string> descriptors, ModManager& mod_manager) {
+	bool ret = true;
+
+	for (std::string_view descriptor_path : descriptors) {
+		if (!mod_manager.load_mod_file(parse_defines(ensure_forward_slash_path(descriptor_path)).get_file_node())) {
+			Logger::error("Failed to load ", descriptor_path);
+			ret = false;
+		}
+	}
+
+	mod_manager.lock_mods();
+
+	return ret;
 }
 
 bool Dataloader::_load_interface_files(UIManager& ui_manager) const {
--- a/src/openvic-simulation/dataloader/Dataloader.cpp
+++ b/src/openvic-simulation/dataloader/Dataloader.cpp
@@ -816,6 +816,11 @@ bool Dataloader::_load_map_dir(DefinitionManager& definition_manager) const {
 		ret = false;
 	}
 
+	if (!map_definition.set_water_province_list(water_province_identifiers)) {
+		Logger::error("Failed to set water provinces!");
+		ret = false;
+	}
+
 	{
 		std::vector<colour_t> colours;
 		if (!MapDefinition::load_region_colours(parse_defines(lookup_file(region_colours)).get_file_node(), colours)) {
@@ -831,11 +836,6 @@ bool Dataloader::_load_map_dir(DefinitionManager& definition_manager) const {
 			Logger::error("Failed to load region file!");
 			ret = false;
 		}
-	}
-
-	if (!map_definition.set_water_province_list(water_province_identifiers)) {
-		Logger::error("Failed to set water provinces!");
-		ret = false;
 	}
 
 	if (!map_definition.get_terrain_type_manager().load_terrain_types(
--- a/src/openvic-simulation/dataloader/Dataloader.hpp
+++ b/src/openvic-simulation/dataloader/Dataloader.hpp
@@ -1,9 +1,11 @@
 #pragma once
 
+#include <string_view>
 #include <openvic-dataloader/csv/Parser.hpp>
 #include <openvic-dataloader/v2script/Parser.hpp>
 
 #include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "openvic-simulation/dataloader/ModManager.hpp"
 
 namespace OpenVic {
 	namespace fs = std::filesystem;
@@ -23,6 +25,7 @@ namespace OpenVic {
 
 	private:
 		path_vector_t PROPERTY(roots);
+		path_vector_t PROPERTY(replace_paths);
 		std::vector<ovdl::v2script::Parser> cached_parsers;
 
 		bool _load_interface_files(UIManager& ui_manager) const;
@@ -87,8 +90,12 @@ namespace OpenVic {
 		///
 		static fs::path search_for_game_path(fs::path hint_path = {});
 
-		/* In reverse-load order, so base defines first and final loaded mod last */
-		bool set_roots(path_vector_t const& new_roots);
+		/// @brief Sets the directories the dataloader should pull content from.
+		///
+		/// @param new_roots Dataloader roots in reverse-load order, so base defines first and final loaded mod last
+		/// @param new_replace_paths All base define paths that should be ignored entirely in favour of mods.
+		/// @return True if successful, false if failed.
+		bool set_roots(path_vector_t const& new_roots, path_vector_t const& new_replace_paths);
 
 		/* REQUIREMENTS:
 		 * DAT-24
@@ -107,6 +114,9 @@ namespace OpenVic {
 		bool apply_to_files(path_vector_t const& files, NodeTools::callback_t<fs::path const&> callback) const;
 
 		string_set_t lookup_dirs_in_dir(std::string_view path) const;
+
+		/* Load all mod descriptors passed by the user. Importantly, loads dependencies and replace_paths for us to check. */
+		bool load_mod_descriptors(std::vector<std::string> descriptors, ModManager& mod_manager);
 
 		/* Load and parse all of the text defines data, including parsing cached condition and effect scripts after all the
 		 * static data is loaded. Paths to the base and mod defines must have been supplied with set_roots.*/
--- a/src/openvic-simulation/dataloader/ModManager.cpp
+++ b/src/openvic-simulation/dataloader/ModManager.cpp
@@ -1,0 +1,49 @@
+#include "ModManager.hpp"
+
+#include <string>
+#include <string_view>
+
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "openvic-simulation/types/HasIdentifier.hpp"
+#include "types/IdentifierRegistry.hpp"
+
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
+
+Mod::Mod(std::string_view new_identifier, std::string_view new_path, std::optional<std::string_view> new_user_dir, std::vector<std::string> new_replace_paths, std::vector<std::string> new_dependencies)
+    : HasIdentifier { new_identifier }, dataloader_root_path { new_path }, user_dir { new_user_dir }, replace_paths { new_replace_paths }, dependencies { new_dependencies } {}
+
+ModManager::ModManager() {}
+
+bool ModManager::load_mod_file(ast::NodeCPtr root) {
+    std::string_view identifier;
+    std::string_view path;
+    std::optional<std::string_view> user_dir;
+    std::vector<std::string> replace_paths;
+    std::vector<std::string> dependencies;
+
+    bool ret = NodeTools::expect_dictionary_keys(
+        "name", ONE_EXACTLY, expect_string(assign_variable_callback(identifier)),
+        "path", ONE_EXACTLY, expect_string(assign_variable_callback(path)),
+        "user_dir", ZERO_OR_ONE, expect_string(assign_variable_callback_opt(user_dir)),
+        "replace_path", ZERO_OR_MORE, expect_string(vector_callback_string(replace_paths)),
+        "dependencies", ZERO_OR_ONE, expect_list_reserve_length(dependencies, expect_string(vector_callback_string(dependencies)))
+    )(root);
+
+    std::vector<std::string_view> previous_mods = mods.get_item_identifiers();
+    for (std::string_view dependency : dependencies) {
+		if (std::find(previous_mods.begin(), previous_mods.end(), dependency) == previous_mods.end()) {
+            ret = false;
+			Logger::error("Mod ", identifier, " has unmet dependency ", dependency);
+		}
+	}
+
+    if (ret) {
+        ret &= mods.add_item(
+            { identifier, path, user_dir, std::move(replace_paths), std::move(dependencies) },
+            duplicate_fail_callback
+        );
+    }
+
+    return ret;
+}
--- a/src/openvic-simulation/dataloader/ModManager.cpp
+++ b/src/openvic-simulation/dataloader/ModManager.cpp
@@ -40,8 +40,7 @@ bool ModManager::load_mod_file(ast::NodeCPtr root) {
 
     if (ret) {
         ret &= mods.add_item(
-            { identifier, path, user_dir, std::move(replace_paths), std::move(dependencies) },
-            duplicate_fail_callback
+            { identifier, path, user_dir, std::move(replace_paths), std::move(dependencies) }
         );
     }
 
--- a/src/openvic-simulation/dataloader/NodeTools.hpp
+++ b/src/openvic-simulation/dataloader/NodeTools.hpp
@@ -4,6 +4,7 @@
 #include <functional>
 #include <optional>
 #include <type_traits>
+#include <vector>
 
 #include <openvic-dataloader/detail/SymbolIntern.hpp>
 #include <openvic-dataloader/v2script/AbstractSyntaxTree.hpp>
@@ -569,6 +570,8 @@ namespace OpenVic {
 				return true;
 			};
 		}
+
+		callback_t<std::string_view> vector_callback_string(std::vector<std::string>& vec);
 
 		template<typename T, typename U, typename... SetArgs>
 		Callback<T> auto set_callback(tsl::ordered_set<U, SetArgs...>& set, bool warn = false) {
--- a/src/openvic-simulation/economy/production/ProductionType.cpp
+++ b/src/openvic-simulation/economy/production/ProductionType.cpp
@@ -220,8 +220,7 @@ bool ProductionTypeManager::load_production_types_file(
 
 			std::string_view template_id = "";
 			bool found_template = false;
-			const bool ret =
-				expect_key(template_symbol, expect_identifier(assign_variable_callback(template_id)), &found_template)(value);
+			const bool ret = expect_key(template_symbol, expect_identifier(assign_variable_callback(template_id)), &found_template)(value);
 			if (found_template) {
 				if (ret) {
 					templates.emplace(template_id);
@@ -229,6 +228,12 @@ bool ProductionTypeManager::load_production_types_file(
 				} else {
 					Logger::error("Failed get template identifier for ", key);
 					return false;
+				}
+			} else {
+				// this registers templates too, works in some mods (PDM/DoD) that have templates that are unused but aren't factories themselves
+				if (key.ends_with(template_symbol.view())) {
+					templates.emplace(key);
+					template_target_map.emplace(key, key);
 				}
 			}
 			return true;
--- a/src/openvic-simulation/history/CountryHistory.cpp
+++ b/src/openvic-simulation/history/CountryHistory.cpp
@@ -2,6 +2,7 @@
 
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/DefinitionManager.hpp"
+#include "dataloader/NodeTools.hpp"
 
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
@@ -135,7 +136,16 @@ bool CountryHistoryMap::_load_history_entry(
 		),
 		"civilized", ZERO_OR_ONE, expect_bool(assign_variable_callback(entry.civilised)),
 		"prestige", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(entry.prestige)),
-		"ruling_party", ZERO_OR_ONE, country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party)),
+		"ruling_party", ZERO_OR_ONE, [this, &entry](ast::NodeCPtr value) -> bool { // theres this awesome thing where vic2 doesn't really care if the party you put here doesn't exist, and DoD uses that for whatever reason. To maintain compatibility, I've assigned the first defined party if the party assigned by ruling_party isn't real and warned about it in the log - brickpi
+			country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party), true)(value);
+			if (!entry.ruling_party.has_value()) {
+				entry.ruling_party = &country.get_front_party();
+				std::string_view def {};
+				expect_identifier(assign_variable_callback(def))(value);
+				Logger::warning("In ", country.get_identifier(), " history: ruling_party ", def, " does NOT exist! Using ", country.get_front_party().get_identifier(), " instead.");
+			}
+			return true;
+		},
 		"last_election", ZERO_OR_ONE, expect_date(assign_variable_callback(entry.last_election)),
 		"upper_house", ZERO_OR_ONE, politics_manager.get_ideology_manager().expect_ideology_dictionary(
 			[&entry](Ideology const& ideology, ast::NodeCPtr value) -> bool {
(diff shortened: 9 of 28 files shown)


src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -60,23 +65,72 @@ bool Dataloader::set_roots(path_vector_t const& new_roots) {
...
+				}
+				if (!ignore) {
+					return composed;
+				}
+			} else {
+				return composed;
+			}
+		}
+	}
+#else
```
Spartan322, line 89: https://github.com/OpenVicProject/OpenVic-Simulation/pull/287#issue-2825496891

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -60,23 +65,72 @@ bool Dataloader::set_roots(path_vector_t const& new_roots) {
...
+								ignore = true;
+								break;
+							}
+						}
+						if (!ignore) {
+							return file;
+						}
+					} else {
+						return file;
+					}
```
Spartan322, line 128: This should be extracted to a function.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -60,23 +65,72 @@ bool Dataloader::set_roots(path_vector_t const& new_roots) {
...
+						ignore = true;
+						break;
+					}
+				}
+				if (!ignore) {
+					return composed;
+				}
+			} else {
+				return composed;
+			}
```
Spartan322, line 108: This should be extracted to a function.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -109,6 +163,18 @@ Dataloader::path_vector_t Dataloader::_lookup_files_in_dir(
...
+			for (auto const& replace_path : replace_paths) {
+				if (dirpath.string().starts_with(replace_path.string())) {
+					ignore = true;
+					break;
+				}
+			}
+			if (ignore) {
+				continue;
+			}
+		}
```
Spartan322, line 177: Pretty sure this also should be extracted to a function.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -735,6 +816,11 @@ bool Dataloader::_load_map_dir(DefinitionManager& definition_manager) const {
 		ret = false;
 	}
 
+	if (!map_definition.set_water_province_list(water_province_identifiers)) {
+		Logger::error("Failed to set water provinces!");
+		ret = false;
+	}
+
```
Spartan322, line 823: Is this a required change for this PR?

src/openvic-simulation/dataloader/Dataloader.hpp
```diff
@@ -87,8 +90,12 @@ namespace OpenVic {
 		///
 		static fs::path search_for_game_path(fs::path hint_path = {});
 
-		/* In reverse-load order, so base defines first and final loaded mod last */
-		bool set_roots(path_vector_t const& new_roots);
+		/// @brief Sets the directories the dataloader should pull content from.
+		///
+		/// @param new_roots Dataloader roots in reverse-load order, so base defines first and final loaded mod last
+		/// @param new_replace_paths All base define paths that should be ignored entirely in favour of mods.
+		/// @return True if successful, false if failed.
```
Spartan322, line 97: Doxygen style comments are a bit inconsistent with other forms of documentation, we probably should decide how we'll handle documentation consistently.

src/openvic-simulation/dataloader/Dataloader.hpp
```diff
@@ -87,8 +90,12 @@ namespace OpenVic {
...
+		bool set_roots(path_vector_t const& new_roots, path_vector_t const& new_replace_paths);
```
Spartan322, line 98: Probably should become
```suggestion
		bool set_root(fs::path const& new_roots, path_vector_t const& new_replace_paths);
```
or something like that.

src/openvic-simulation/dataloader/ModManager.cpp
```diff
@@ -0,0 +1,48 @@
...
+        "path", ONE_EXACTLY, expect_string(assign_variable_callback(path)),
+        "user_dir", ZERO_OR_ONE, expect_string(assign_variable_callback_opt(user_dir)),
+        "replace_path", ZERO_OR_MORE, expect_string(vector_callback_string(replace_paths)),
+        "dependencies", ZERO_OR_ONE, expect_list_reserve_length(dependencies, expect_string(vector_callback_string(dependencies)))
+    )(root);
+
+    std::vector<std::string_view> previous_mods = mods.get_item_identifiers();
+    for (std::string_view dependency : dependencies) {
+		if (std::find(previous_mods.begin(), previous_mods.end(), dependency) == previous_mods.end()) {
+            ret = false;
```
Spartan322, line 36:
```suggestion
			ret = false;
```

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/DefinitionManager.hpp"
+#include "dataloader/NodeTools.hpp"
```
Spartan322, line 5:
```suggestion
#include "openvic-simulation/dataloader/NodeTools.hpp"
```

src/openvic-simulation/dataloader/ModManager.cpp
```diff
@@ -0,0 +1,48 @@
+#include "ModManager.hpp"
+
+#include <string>
+#include <string_view>
+
+#include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "openvic-simulation/types/HasIdentifier.hpp"
+#include "types/IdentifierRegistry.hpp"
```
Spartan322, line 8:
```suggestion
#include "openvic-simulation/types/IdentifierRegistry.hpp"
```

src/openvic-simulation/country/CountryDefinition.cpp
```diff
@@ -11,6 +11,7 @@
 #include "openvic-simulation/pop/Culture.hpp"
 #include "openvic-simulation/types/Colour.hpp"
 #include "openvic-simulation/types/IdentifierRegistry.hpp"
+#include "utility/Logger.hpp"
```
Spartan322, line 14:
```suggestion
#include "openvic-simulation/utility/Logger.hpp"
```

src/openvic-simulation/dataloader/ModManager.cpp
```diff
@@ -0,0 +1,48 @@
...
+        "dependencies", ZERO_OR_ONE, expect_list_reserve_length(dependencies, expect_string(vector_callback_string(dependencies)))
+    )(root);
+
+    std::vector<std::string_view> previous_mods = mods.get_item_identifiers();
+    for (std::string_view dependency : dependencies) {
+		if (std::find(previous_mods.begin(), previous_mods.end(), dependency) == previous_mods.end()) {
+            ret = false;
+			Logger::error("Mod ", identifier, " has unmet dependency ", dependency);
+		}
+	}
```
Spartan322, line 39: Indentation is off.

src/openvic-simulation/military/Deployment.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Deployment.hpp"
 
 #include "openvic-simulation/DefinitionManager.hpp" /* gosh don't we all just love circular inclusion :DDD */
+#include "military/LeaderTrait.hpp"
```
Spartan322, line 4:
```suggestion
#include "openvic-simulation/military/LeaderTrait.hpp"
```

src/openvic-simulation/modifier/ModifierEffectCache.hpp
```diff
@@ -50,6 +50,7 @@ namespace OpenVic {
 		ModifierEffect const* PROPERTY(artisan_input);
 		ModifierEffect const* PROPERTY(artisan_output);
 		ModifierEffect const* PROPERTY(artisan_throughput);
+		ModifierEffect const* PROPERTY(assimilation_rate_g);
```
wvpm, line 53: assimilation_rate_global
Don't shorten it.

src/openvic-simulation/modifier/ModifierEffectCache.hpp
```diff
@@ -321,6 +324,7 @@ namespace OpenVic {
 			ModifierEffect const* PROPERTY(life_needs);
 			ModifierEffect const* PROPERTY(everyday_needs);
 			ModifierEffect const* PROPERTY(luxury_needs);
+			ModifierEffect const* PROPERTY(savings); // used by HPM, PDM, TGC, etc, and has localisation in vanilla but never used in vanilla. not entirely sure what it does either. -brickpi
```
wvpm, line 327: I only know `poor_savings_modifier` could you document this in the google sheet?
https://docs.google.com/spreadsheets/d/18KtZoV58btUmfdylSnU2VjLzw6edIDnJzKehuGO8FhI/edit?gid=0#gid=0

src/openvic-simulation/modifier/ModifierEffectCache.hpp
```diff
@@ -82,7 +83,8 @@ namespace OpenVic {
 		ModifierEffect const* PROPERTY(influence_modifier);
 		ModifierEffect const* PROPERTY(issue_change_speed);
 		ModifierEffect const* PROPERTY(land_attack_modifier);
-		ModifierEffect const* PROPERTY(land_attrition);
+		ModifierEffect const* PROPERTY(land_attrition_country);
+		ModifierEffect const* PROPERTY(land_attrition_tech);
```
wvpm, line 87: Avoid _tech & _country distrinctions.
Use functional distinctions.
What is the difference in effect between these 2?

If there they share the same effect, use a single modifier and register it as `shared_tech_country`.

wvpm (changes requested):
See modifier comments.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -151,15 +151,16 @@ bool ModifierManager::setup_modifier_effects() {
...
-	ret &= register_technology_modifier_effect(
+	ret &= register_base_country_modifier_effect(
 		modifier_effect_cache.artisan_input, "artisan_input", false, PROPORTION_DECIMAL, {}, has_no_effect
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_base_country_modifier_effect(
 		modifier_effect_cache.artisan_output, "artisan_output", true, PROPORTION_DECIMAL, {}, has_no_effect
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_base_country_modifier_effect(
```
wvpm, line 160: Use shared_tech_country

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -151,15 +151,16 @@ bool ModifierManager::setup_modifier_effects() {
 		modifier_effect_cache.administrative_efficiency_modifier, "administrative_efficiency_modifier", true,
 		PROPORTION_DECIMAL, ModifierEffect::make_default_modifier_effect_localisation_key("administrative_efficiency")
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_base_country_modifier_effect(
 		modifier_effect_cache.artisan_input, "artisan_input", false, PROPORTION_DECIMAL, {}, has_no_effect
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_base_country_modifier_effect(
```
wvpm, line 157: Use shared_tech_country

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -151,15 +151,16 @@ bool ModifierManager::setup_modifier_effects() {
 		modifier_effect_cache.administrative_efficiency_modifier, "administrative_efficiency_modifier", true,
 		PROPORTION_DECIMAL, ModifierEffect::make_default_modifier_effect_localisation_key("administrative_efficiency")
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_base_country_modifier_effect(
```
wvpm, line 154: Use shared_tech_country

src/openvic-simulation/modifier/ModifierEffectCache.hpp
wvpm, line 53: Could you also add `assimilation_rate` as `assimilation_rate_local`?
That's the province modifier.

wvpm, line 53: Both `global_assimilation_rate` & `assimilation_rate` already exist.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -151,15 +151,16 @@ bool ModifierManager::setup_modifier_effects() {
...
 		modifier_effect_cache.artisan_output, "artisan_output", true, PROPORTION_DECIMAL, {}, has_no_effect
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_base_country_modifier_effect(
 		modifier_effect_cache.artisan_throughput, "artisan_throughput", true, PROPORTION_DECIMAL, {}, has_no_effect
 	);
+	ret &= register_base_country_modifier_effect(modifier_effect_cache.assimilation_rate_g, "assimilation_rate", true, PROPORTION_DECIMAL, {}, has_no_effect),
```
wvpm, line 163: `assimilation_rate` has no effect when used on a country.
It is a province modifier.

src/openvic-simulation/modifier/ModifierEffectCache.hpp
wvpm: I tested it myself, land_attrition is not a valid country modifier.

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -263,8 +264,11 @@ bool ModifierManager::setup_modifier_effects() {
 		modifier_effect_cache.land_attack_modifier, "land_attack_modifier", true, PROPORTION_DECIMAL,
 		ModifierEffect::make_default_modifier_effect_localisation_key("land_attack")
 	);
+	ret &= register_base_country_modifier_effect( // both of these are used and work
```
wvpm, line 267: This does not work as a country modifier. You can test it by adding it to event modifiers.

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -4,6 +4,7 @@
 #include "openvic-simulation/economy/GoodDefinition.hpp"
 #include "openvic-simulation/map/ProvinceDefinition.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
+#include "dataloader/NodeTools.hpp"
```
Spartan322, line 7:
```suggestion
#include "openvic-simulation/dataloader/NodeTools.hpp"
```

src/openvic-simulation/interface/UI.cpp
```diff
@@ -1,6 +1,7 @@
 #include "UI.hpp"
 
 #include "openvic-simulation/types/Colour.hpp"
+#include "dataloader/NodeTools.hpp"
```
Spartan322, line 4:
```suggestion
#include "openvic-simulation/dataloader/NodeTools.hpp"
```

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -504,6 +491,120 @@ static constexpr colour_t colour_at(uint8_t const* colour_data, int32_t idx) {
...
+// Constants in the River BMP Palette
+static constexpr uint8_t START_COLOUR = 0;
+static constexpr uint8_t MERGE_COLOUR = 1;
+static constexpr uint8_t RIVER_SIZE_MIN = 2;
+static constexpr uint8_t RIVER_SIZE_MAX = 11;
+
+void MapDefinition::_trace_river(BMP& rivers_bmp, ivec2_t start, river_t& river) {
+	struct TraceSegment {
+		ivec2_t point;
+		uint8_t direction; /* 0 - start, 1 - up, 2 - down, 3 - left, 4 - right */
```
Spartan322, line 503: This should be an enum.

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -504,6 +491,120 @@ static constexpr colour_t colour_at(uint8_t const* colour_data, int32_t idx) {
...
+		points.push_back(segment.point);
+
+		uint8_t size = river_data[segment.point.x + segment.point.y * rivers_bmp.get_width()] - 1; // determine river size by colour
+		bool river_complete = false;
+		size_t recursion_limit = 0;
+
+		while(true) {
+			++recursion_limit;
+			if (recursion_limit == 4096) {
+				Logger::error("River segment starting @ (", points.front().x, ", ", points.front().y, ") exceeded length limit of 4096 pixels. Check for misplaced pixel or circular river.");
```
Spartan322, line 525: Magic numbers are bad, magic numbers should always be named constants.

src/openvic-simulation/military/Deployment.cpp
```diff
@@ -1,6 +1,7 @@
 #include "Deployment.hpp"
```
Spartan322, line 1:
```suggestion
#include "openvic-simulation/military/Deployment.hpp"
```

src/openvic-simulation/scripts/Condition.cpp
```diff
@@ -79,6 +79,7 @@ bool ConditionManager::setup_conditions(DefinitionManager const& definition_mana
 	ret &= add_condition("any_greater_power", GROUP, COUNTRY, COUNTRY);
 	ret &= add_condition("any_neighbor_country", GROUP, COUNTRY, COUNTRY);
 	ret &= add_condition("any_owned_province", GROUP, COUNTRY, PROVINCE);
+	ret &= add_condition("any_owned", GROUP, COUNTRY, PROVINCE); // alias? not in vanilla, but used by HPM poptypes, so used by a ton of derivative mods too
```
wvpm, line 82: This is an effect scope. See https://vic2.paradoxwikis.com/List_of_scopes#any_owned
It is easily mistaken for a condition scope.

src/openvic-simulation/pop/PopType.cpp
```diff
@@ -557,6 +561,7 @@ bool PopManager::generate_modifiers(ModifierManager& modifier_manager) const {
 		ModifierEffectCache::strata_effects_t& this_strata_effects = strata_effects[strata];
 
 		strata_modifier(this_strata_effects.income_modifier, "_income_modifier", true); // Has no effect in game
+		strata_modifier(this_strata_effects.savings, "_savings_modifier", true); // No or unclear in game effect
```
wvpm, line 564: This only exist for poor, not for rich or middle. See the modifiers documentation sheet.
https://docs.google.com/spreadsheets/d/18KtZoV58btUmfdylSnU2VjLzw6edIDnJzKehuGO8FhI/edit?gid=0#gid=0

src/openvic-simulation/military/Deployment.cpp
```diff
@@ -87,27 +88,36 @@ bool DeploymentManager::load_oob_file(
...
-				.expect_leader_trait_identifier_or_string(assign_variable_callback_pointer(leader_personality)),
+				.expect_leader_trait_identifier_or_string(assign_variable_callback_pointer(leader_personality), false, true),
 			"background", ONE_EXACTLY, definition_manager.get_military_manager().get_leader_trait_manager()
-				.expect_leader_trait_identifier_or_string(assign_variable_callback_pointer(leader_background)),
+				.expect_leader_trait_identifier_or_string(assign_variable_callback_pointer(leader_background), false, true),
 			"prestige", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(leader_prestige)),
 			"picture", ZERO_OR_ONE, expect_identifier_or_string(assign_variable_callback(picture))
 		)(node);
 
+		// Default cases for leader personality and background match vic2 behaviour of ignoring invalid traits. Vic2 has a weird noTrait state which is undesirable, so instead we give them the first applicable trait we can find.
```
wvpm, line 98: We should match Victoria 2 with the noTrait instead of deviating.
It's up to the modder to fix his mod and add valid traits.

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -135,7 +136,16 @@ bool CountryHistoryMap::_load_history_entry(
 		),
 		"civilized", ZERO_OR_ONE, expect_bool(assign_variable_callback(entry.civilised)),
 		"prestige", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(entry.prestige)),
-		"ruling_party", ZERO_OR_ONE, country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party)),
+		"ruling_party", ZERO_OR_ONE, [this, &entry](ast::NodeCPtr value) -> bool { // theres this awesome thing where vic2 doesn't really care if the party you put here doesn't exist, and DoD uses that for whatever reason. To maintain compatibility, I've assigned the first defined party if the party assigned by ruling_party isn't real and warned about it in the log - brickpi
```
wvpm, line 139: Does Victoria 2 also select the first defined party?
Or does Victoria 2 do the noParty thing?
Also I expect the party has to at minimum be active at the starting date.

src/openvic-simulation/economy/production/ProductionType.cpp
```diff
@@ -230,6 +229,12 @@ bool ProductionTypeManager::load_production_types_file(
 					Logger::error("Failed get template identifier for ", key);
 					return false;
 				}
+			} else {
```
wvpm, line 232: Example/proof?

src/headless/main.cpp
```diff
@@ -55,15 +56,28 @@ static void print_rgo(ProvinceInstance const& province) {
 	}
 }
 
-static bool run_headless(Dataloader::path_vector_t const& roots, bool run_tests) {
+static bool run_headless(const fs::path root, std::vector<std::string> mods, bool run_tests) {
```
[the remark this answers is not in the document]
BrickPi: Each mod passed is also a root. The base game is passed separately, because we have to validate mods and check that the directory passed from the .mod file is exists etc.

src/openvic-simulation/dataloader/Dataloader.cpp
BrickPi: Yeah, was previously causing issues with HPM.

src/openvic-simulation/modifier/ModifierEffectCache.hpp
BrickPi: Modifiers in the wrong place / that don't exist in V2 now emit a warning, these extraneous ones have been removed.

src/openvic-simulation/military/Deployment.cpp
BrickPi: Fixed, now leaves them as nullptrs for UI to display noTrait.

src/openvic-simulation/history/CountryHistory.cpp
BrickPi: No longer picks another party, simply ignores the offending line of the history entry.

src/openvic-simulation/economy/production/ProductionType.cpp
BrickPi: Tested with PDM

src/openvic-simulation/politics/Government.cpp
```diff
@@ -58,7 +58,7 @@ bool GovernmentTypeManager::load_government_types_file(IdeologyManager const& id
 			size_t total_expected_ideologies = 0;
 			bool ret = expect_dictionary_keys_and_default(
 				increment_callback(total_expected_ideologies),
-				"election", ONE_EXACTLY, expect_bool(assign_variable_callback(elections)),
+				"election", ZERO_OR_ONE, expect_bool(assign_variable_callback(elections)),
```
wvpm, line 61: Correct. It defaults to false. (`elections = false` was already in the old code.)

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -151,13 +151,13 @@ bool ModifierManager::setup_modifier_effects() {
 		modifier_effect_cache.administrative_efficiency_modifier, "administrative_efficiency_modifier", true,
 		PROPORTION_DECIMAL, ModifierEffect::make_default_modifier_effect_localisation_key("administrative_efficiency")
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_shared_tech_country_modifier_effect(
 		modifier_effect_cache.artisan_input, "artisan_input", false, PROPORTION_DECIMAL, {}, has_no_effect
```
wvpm, line 155: Correct. Note they have no effect though...

src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -263,7 +263,7 @@ bool ModifierManager::setup_modifier_effects() {
 		modifier_effect_cache.land_attack_modifier, "land_attack_modifier", true, PROPORTION_DECIMAL,
 		ModifierEffect::make_default_modifier_effect_localisation_key("land_attack")
 	);
-	ret &= register_technology_modifier_effect(
+	ret &= register_shared_tech_country_modifier_effect(
```
wvpm, line 266: This is not valid as country modifier. It only works in tech.

src/openvic-simulation/interface/GUI.cpp
```diff
@@ -114,7 +116,9 @@ bool Icon::_fill_key_map(NodeTools::case_insensitive_key_map_t& key_map, UIManag
 
 		"frame", ZERO_OR_ONE, expect_uint(assign_variable_callback(frame)),
 		"scale", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(scale)),
-		"rotation", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(rotation))
+		"rotation", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(rotation)),
+		"maxWidth", ZERO_OR_ONE, success_callback, /* some mods use this, there's no effect afaik */
```
wvpm, line 120: Since it's not used, modders should receive a warning that their code does nothing.
Otherwise we pretend to understand it and then discard these values.

src/openvic-simulation/economy/production/ProductionType.cpp
wvpm: Can you provide a minimum working example?
Victoria 2 does nothing special with the name of a production type afaik.

Also, what is the point of having an unused template?
Is it not merely an unused production type?

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -135,7 +136,15 @@ bool CountryHistoryMap::_load_history_entry(
 		),
 		"civilized", ZERO_OR_ONE, expect_bool(assign_variable_callback(entry.civilised)),
 		"prestige", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(entry.prestige)),
-		"ruling_party", ZERO_OR_ONE, country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party)),
+		"ruling_party", ZERO_OR_ONE, [this, &entry](ast::NodeCPtr value) -> bool {
+			country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party), true)(value);
+			if (!entry.ruling_party.has_value()) {
+				std::string_view def {};
+				expect_identifier(assign_variable_callback(def))(value);
+				Logger::warning("In ", country.get_identifier(), " history: ruling_party ", def, " does NOT exist!");
```
wvpm, line 144: This results in an invalid state that OpenVic can't handle yet.
It's not a minor warning we can just ignore.
We'll either have to add support for a noParty or treat this as a game breaking error.

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -159,7 +160,11 @@ bool ProvinceHistoryMap::_load_history_entry(
 				}
 			}
 			return ret;
-		}
+		},
+		"revolt", ZERO_OR_ONE, expect_dictionary_keys(
+			"type", ONE_EXACTLY, definition_manager.get_politics_manager().get_rebel_manager().expect_rebel_type_identifier(assign_variable_callback_pointer_opt(entry.revolt)),
+			"controller", ONE_EXACTLY, success_callback
```
wvpm, line 166: Anything that is ignored should result in a warning.
We don't want to give modders the false impression that we do something with this.

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -159,7 +160,11 @@ bool ProvinceHistoryMap::_load_history_entry(
 				}
 			}
 			return ret;
-		}
+		},
+		"revolt", ZERO_OR_ONE, expect_dictionary_keys(
```
wvpm, line 164: Could you provide a minimum working example?
I could not replicate it with
```
revolt = {
	type = liberal_rebels
}
```

src/openvic-simulation/pop/PopType.cpp
```diff
@@ -546,6 +546,10 @@ bool PopManager::load_pop_bases_into_vector(
 	if (non_integer_size != nullptr && !size.is_integer()) {
 		*non_integer_size = true;
 	}
+	if (culture == nullptr || religion == nullptr) {
+		Logger::warning("No/invalid culture or relgion defined for pop of size ", size, ", ignored!");
```
wvpm, line 550: This warning should contain more information.
At least provide the province containing the pop. Preferably also provide the culture and/or religion text values.

src/openvic-simulation/history/ProvinceHistory.cpp
BrickPi: Whoops - another case of me trusting mods too much! Got rid of the variable and now simply warn that this does nothing.

src/openvic-simulation/pop/PopType.cpp
BrickPi: The way PopHistory callbacks are structured theres no good way to get the province id, and they'll be able to see culture / religion text values in the warning generated by expect_dictionary_keys above.

src/openvic-simulation/GameManager.cpp
```diff
@@ -6,10 +6,23 @@ GameManager::GameManager(
 	InstanceManager::gamestate_updated_func_t new_gamestate_updated_callback
 ) : gamestate_updated_callback {
 		new_gamestate_updated_callback ? std::move(new_gamestate_updated_callback) : []() {}
-	}, definitions_loaded { false } {}
+	}, definitions_loaded { false }, mod_descriptors_loaded { false } {}
 
-bool GameManager::set_roots(Dataloader::path_vector_t const& roots) {
-	if (!dataloader.set_roots(roots)) {
+bool GameManager::load_mod_descriptors(std::vector<std::string> descriptors) {
```
Spartan322, line 11:
```suggestion
bool GameManager::load_mod_descriptors(std::vector<std::string> const& descriptors) {
```

src/openvic-simulation/dataloader/Dataloader.hpp
```diff
@@ -115,6 +124,9 @@ namespace OpenVic {
 
 		string_set_t lookup_dirs_in_dir(std::string_view path) const;
 
+		/* Load all mod descriptors passed by the user. Importantly, loads dependencies and replace_paths for us to check. */
+		bool load_mod_descriptors(std::vector<std::string> descriptors, ModManager& mod_manager);
```
Spartan322, line 128:
```suggestion
		bool load_mod_descriptors(std::vector<std::string> const& descriptors, ModManager& mod_manager);
```

src/openvic-simulation/GameManager.cpp
```diff
@@ -6,10 +6,23 @@ GameManager::GameManager(
...
+	if (mod_descriptors_loaded) {
+		Logger::error("Cannot load mod descriptors - already loaded!");
+		return false;
+	}
+
+	if (!dataloader.load_mod_descriptors(std::move(descriptors), mod_manager)) {
```
Spartan322, line 17:
```suggestion
	if (!dataloader.load_mod_descriptors(descriptors, mod_manager)) {
```

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -35,12 +35,17 @@ static fs::path ensure_forward_slash_path(std::string_view path) {
...
 
-bool Dataloader::set_roots(path_vector_t const& new_roots) {
+bool Dataloader::set_roots(path_vector_t const& new_roots, path_vector_t const& new_replace_paths) {
 	if (!roots.empty()) {
 		Logger::warning("Overriding existing dataloader roots!");
 		roots.clear();
+		replace_paths.clear();
 	}
 	bool ret = true;
+	for (const fs::path& replace_path : new_replace_paths) {
```
Spartan322, line 45:
```suggestion
	for (fs::path const& replace_path : new_replace_paths) {
```

src/openvic-simulation/dataloader/NodeTools.hpp
```diff
@@ -120,7 +121,10 @@ using namespace std::string_view_literals;
 		constexpr bool key_value_success_callback(std::string_view, ast::NodeCPtr) {
 			return true;
 		}
-
+		inline bool key_value_warn_callback(std::string_view key, ast::NodeCPtr) {
```
wvpm, line 124: `map_key_value_ignore_invalid_callback` was added in a recent PR.
Please use that instead as it lists the valid keys.

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -135,7 +136,37 @@ bool CountryHistoryMap::_load_history_entry(
...
+			country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party), true)(value);
+			if (!entry.ruling_party.has_value()) {
+				std::string_view def {};
+				bool valid = true;
+				
+				expect_identifier(assign_variable_callback(def))(value);
+
+				// if the party specified is invalid, replace with the first valid party
+				for (const auto& party : country.get_parties()) {
+					if (party.get_start_date() < entry.get_date() && entry.get_date() < party.get_end_date()) {
```
wvpm, line 149: Shouldn't it be `party.get_start_date() <= entry.get_date()`?
So it's valid on the start date.

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -135,7 +136,37 @@ bool CountryHistoryMap::_load_history_entry(
...
+					}
+				}
+
+				// if there's somehow no valid party, we use the first defined party even though it isn't valid for the entry
+				if (!entry.ruling_party.has_value()) {
+					valid = false;
+					entry.ruling_party.emplace(&country.get_front_party());
+				}
+				
+				Logger::warning(
```
wvpm, line 161: Could be error if the party isn't valid.
You can use `Logger::warn_or_error(valid, ...)`

src/openvic-simulation/history/CountryHistory.cpp
```diff
@@ -135,7 +136,37 @@ bool CountryHistoryMap::_load_history_entry(
 		),
 		"civilized", ZERO_OR_ONE, expect_bool(assign_variable_callback(entry.civilised)),
 		"prestige", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(entry.prestige)),
-		"ruling_party", ZERO_OR_ONE, country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party)),
+		"ruling_party", ZERO_OR_ONE, [this, &entry](ast::NodeCPtr value) -> bool {
+			country.expect_party_identifier(assign_variable_callback_pointer_opt(entry.ruling_party), true)(value);
+			if (!entry.ruling_party.has_value()) {
+				std::string_view def {};
+				bool valid = true;
```
wvpm, line 143: Please prefix boolean values with `is_` or `has_` (or similar).
This makes it easier to read and instantly clear that it is a boolean.

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -159,7 +160,8 @@ bool ProvinceHistoryMap::_load_history_entry(
 				}
 			}
 			return ret;
-		}
+		},
+		"revolt", ZERO_OR_ONE, [&entry](ast::NodeCPtr node) -> bool { Logger::warning("In province history for ", entry.get_province(), " \"revolt\" has no effect!"); return true; }
```
wvpm, line 164: Do we need a custom callback for this?
Wouldn't the default `map_key_value_invalid_callback` suffice?

src/openvic-simulation/interface/GUI.cpp
```diff
@@ -114,7 +115,10 @@ bool Icon::_fill_key_map(NodeTools::case_insensitive_key_map_t& key_map, UIManag
 
 		"frame", ZERO_OR_ONE, expect_uint(assign_variable_callback(frame)),
 		"scale", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(scale)),
-		"rotation", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(rotation))
+		"rotation", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(rotation)),
+
+		"maxWidth", ZERO_OR_ONE, [](ast::NodeCPtr node) -> bool { Logger::warning("maxWidth has no effect, so this line will do nothing."); return true; }, /* some mods use these to no effect */
```
wvpm, line 120: Do we need a custom callback for this?
Wouldn't the default `map_key_value_invalid_callback` suffice?

src/openvic-simulation/map/TerrainType.cpp
```diff
@@ -172,7 +172,11 @@ bool TerrainTypeManager::_load_terrain_type_mapping(std::string_view mapping_key
 	TerrainTypeMapping::index_t priority = 0;
 	bool has_texture = true;
 
-	bool ret = expect_dictionary_keys(
+	bool ret = expect_dictionary_keys_and_default(
```
wvpm, line 175: Do we need a custom callback for this?
Wouldn't the default `map_key_value_invalid_callback` suffice?

src/openvic-simulation/military/Deployment.cpp
```diff
@@ -87,9 +88,9 @@ bool DeploymentManager::load_oob_file(
 			"date", ONE_EXACTLY, expect_date_identifier_or_string(assign_variable_callback(leader_date)),
 			"type", ONE_EXACTLY, UnitTypeManager::expect_branch_identifier(assign_variable_callback(leader_branch)),
 			"personality", ONE_EXACTLY, definition_manager.get_military_manager().get_leader_trait_manager()
-				.expect_leader_trait_identifier_or_string(assign_variable_callback_pointer(leader_personality)),
+				.expect_leader_trait_identifier_or_string(assign_variable_callback_pointer(leader_personality), false, true),
```
wvpm, line 91: Please use named constexpr booleans so readers know what false and true mean here.
Example `constexpr bool do_warn = true;`

wvpm (approved):
Approved with minor comments.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -100,6 +130,17 @@ fs::path Dataloader::lookup_image_file(std::string_view path) const {
 	return lookup_file(path);
 }
 
+bool Dataloader::should_ignore_path(fs::path const& path, path_vector_t const& replace_paths) const {
+	bool ignore = false;
+	for (auto const& replace_path : replace_paths) {
```
Spartan322, line 135: Please don't rely type deduction unless you have no choice.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -65,19 +70,44 @@ bool Dataloader::set_roots(path_vector_t const& new_roots) {
...
 		}
 		std::error_code ec;
 		for (fs::directory_entry const& entry : fs::directory_iterator { composed.parent_path(), ec }) {
 			if (entry.is_regular_file()) {
 				const fs::path file = entry;
 				if (StringUtils::strings_equal_case_insensitive(file.filename().string(), filename)) {
-					return file;
+					if (root == roots.back()) {
+						bool ignore = false;
+						for (auto const& replace_path : replace_paths) {
```
Spartan322, line 99: Please don't rely type deduction unless you have no choice.

src/openvic-simulation/dataloader/Dataloader.cpp
```diff
@@ -65,19 +70,44 @@ bool Dataloader::set_roots(path_vector_t const& new_roots) {
...
 
-	/* Case-sensitive filesystem */
 	const std::string_view filename = StringUtils::get_filename(path);
 	for (fs::path const& root : roots) {
 		const fs::path composed = root / filepath;
 		if (fs::is_regular_file(composed)) {
-			return composed;
+			if (root == roots.back()) {
+				bool ignore = false;
+				for (auto const& replace_path : replace_paths) {
```
Spartan322, line 79: Please don't rely type deduction unless you have no choice.

src/openvic-simulation/dataloader/ModManager.cpp
```diff
@@ -0,0 +1,48 @@
...
+		"name", ONE_EXACTLY, expect_string(assign_variable_callback(identifier)),
+		"path", ONE_EXACTLY, expect_string(assign_variable_callback(path)),
+		"user_dir", ZERO_OR_ONE, expect_string(assign_variable_callback_opt(user_dir)),
+		"replace_path", ZERO_OR_MORE, expect_string(vector_callback_string(replace_paths)),
+		"dependencies", ZERO_OR_ONE, expect_list_reserve_length(dependencies, expect_string(vector_callback_string(dependencies)))
+	)(root);
+
+	std::vector<std::string_view> previous_mods = mods.get_item_identifiers();
+	for (std::string_view dependency : dependencies) {
+		if (std::find(previous_mods.begin(), previous_mods.end(), dependency) == previous_mods.end()) {
```
Spartan322, line 35: Wouldn't this correctly require even just a simple dependency graph to do correctly? Dependency checks require all the mod descriptiors to foremost be loaded

src/openvic-simulation/dataloader/NodeTools.hpp
BrickPi: The issue with `map_key_value_ignore_invalid_callback` for this usecase is it returns false, when the vic2 behaviour is to continue and ignore, hence a warning callback fits better here.

src/openvic-simulation/dataloader/ModManager.cpp
BrickPi: We could do it that way, for now I haven't since the user decides load order manually by passing arguments in order. Probably something to look at in the future if we add a dedicated GUI for mod loading.

src/headless/main.cpp
```diff
@@ -109,15 +112,28 @@ static std::chrono::nanoseconds run_pathing_test(AStarPathing& pathing) {
...
 	bool ret = true;
+	Dataloader::path_vector_t roots = { root };
+	Dataloader::path_vector_t replace_paths = {};
 
 	GameManager game_manager { []() {
 		Logger::info("State updated");
 	} };
 
+	Logger::info("===== Loading mod descriptors... =====");
+	ret &= game_manager.load_mod_descriptors(std::move(mods));
```
Spartan322, line 125: std::move(mods) does literally nothing

```suggestion
	ret &= game_manager.load_mod_descriptors(mods);
```

BrickPi: Fixed

src/headless/main.cpp
```diff
@@ -218,6 +234,7 @@ int main(int argc, char const* argv[]) {
 
 	char const* program_name = StringUtils::get_filename(argc > 0 ? argv[0] : nullptr, "<program>");
 	fs::path root;
+	std::vector<std::string> mods;
```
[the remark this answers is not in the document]
Spartan322: Technically we could have std::vector<std::string_view> here, but is vector of string_views is not convertible to vector of strings so everything need to support vector of string_views instead.

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -159,7 +160,8 @@ bool ProvinceHistoryMap::_load_history_entry(
 				}
 			}
 			return ret;
-		}
+		},
+		"revolt", ZERO_OR_ONE, [&entry](ast::NodeCPtr node) -> bool { Logger::warning("In province history for ", entry.get_province(), " key \"revolt\" has no effect!"); return true; }
```
wvpm, line 164: No need for a custom message. Use either `map_key_value_ignore_invalid_callback` or `map_key_value_invalid_callback` to handle all such cases.

src/openvic-simulation/dataloader/NodeTools.hpp
wvpm: `map_key_value_ignore_invalid_callback` is only added in #363  see https://github.com/OpenVicProject/OpenVic-Simulation/pull/363/files#diff-863e89118363d8d3e04ddc1233a97cc3c3fef0851315e8d61847db35e681c07c
Once that is merged, you can use it here. If there is an obstacle to merging, I'll take the nodetools part out and make a separate PR for that.

src/openvic-simulation/interface/GUI.cpp
wvpm: A `success_callback` lies to modders by acting as if these values are parsed. Instead they are invalid and should result in a warning.

My clangd is broken so I can't untangle the nodetools overloads mess.
Still there is a method that uses `map_key_value_invalid_callback` as default callback so invalid keys are reported as warnings and valid keys are listed. Use that instead of adding entries for made up keys.

wvpm (changes requested):
Avoid custom callbacks for invalid keys. Use the default handling instead.
Either `map_key_value_invalid_callback` or `map_key_value_ignore_invalid_callback`

src/openvic-simulation/dataloader/NodeTools.hpp
BrickPi: I'm keeping this callback in here as `map_key_value_ignore_invalid_callback` doesn't work when compile-time unknown keys are valid, such as production methods and buildings. In that case, where some keys are specifically invalid, rather than only some keys being specifically valid, which is what `...ignore_invalid_callback` works for, `key_value_warn_callback` makes more sense.

src/openvic-simulation/economy/production/ProductionType.cpp
BrickPi: I'm not sure if the name itself is what V2 uses here to distinguish, but these unused templates definitely behave differently to normal production types, hence our previous code didn't work when loading them as it treated them like normal production types and errored out.

src/openvic-simulation/scripts/Condition.cpp
BrickPi: Fixed, removed that condition, now warns it's invalid.

Spartan322 (dismissed):

src/openvic-simulation/history/ProvinceHistory.cpp
```diff
@@ -159,7 +160,8 @@ bool ProvinceHistoryMap::_load_history_entry(
 				}
 			}
 			return ret;
-		}
+		},
+		"revolt", ZERO_OR_ONE, [](ast::NodeCPtr node) -> bool { return key_value_warn_callback("revolt", node); }
```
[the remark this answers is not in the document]
BrickPi: Good catch, removed.

src/openvic-simulation/interface/GUI.cpp
```diff
@@ -114,7 +115,10 @@ bool Icon::_fill_key_map(NodeTools::case_insensitive_key_map_t& key_map, UIManag
 
 		"frame", ZERO_OR_ONE, expect_uint(assign_variable_callback(frame)),
 		"scale", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(scale)),
-		"rotation", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(rotation))
+		"rotation", ZERO_OR_ONE, expect_fixed_point(assign_variable_callback(rotation)),
+
+		"maxWidth", ZERO_OR_ONE, [](ast::NodeCPtr node) -> bool { return key_value_warn_callback("maxWidth", node); }, /* some mods use these to no effect */
```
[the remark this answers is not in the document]
wvpm: I searched for the usage of the parent function and it leads to `openvic-simulation\src\openvic-simulation\interface\LoadBase.hpp`.
There the key_map is passed to `ret &= NodeTools::expect_dictionary_key_map(std::move(key_map))(node);`.

This is the implementation:
```cpp
		template<StringMapCase Case>
		NodeCallback auto expect_dictionary_key_map(template_key_map_t<Case>&& key_map) {
			return expect_dictionary_key_map_and_length_and_default(
				MOV(key_map), default_length_callback, map_key_value_invalid_callback<template_key_map_t<Case>>
			);
		}
```

So it falls back to `map_key_value_invalid_callback`, which treats keys not present in the map as invalid.
If you want to use warnings, replace the code in `LoadBase.hpp` with `NodeTools::expect_dictionary_key_map_and_default(std::move(key_map), NodeTools::map_key_value_ignore_invalid_callback)(node);` which ignores invalid values and returns true instead of false.

My main point is that `maxWidth` and `maxHeight` should not be treated special just because some mod mistakenly added it. As Victoria 2 does not recognise it, we should treat it as any other invalid key.

wvpm (dismissed):

PR #292, opened 2025-02-03
Add simulation dev console

Spartan322:
Add `InstanceManager::set_today_and_update`
Add `SimulationClock::force_advance_game`
Add contrast and invert colour functions

### DIFF
--- a/src/openvic-simulation/InstanceManager.cpp
+++ b/src/openvic-simulation/InstanceManager.cpp
@@ -1,6 +1,7 @@
 #include "InstanceManager.hpp"
 
 #include "openvic-simulation/DefinitionManager.hpp"
+#include "openvic-simulation/console/ConsoleInstance.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
 
 using namespace OpenVic;
@@ -27,6 +28,7 @@ InstanceManager::InstanceManager(
 		std::bind(&InstanceManager::tick, this), std::bind(&InstanceManager::update_gamestate, this),
 		clock_state_changed_callback ? std::move(clock_state_changed_callback) : []() {}
 	},
+	console_instance { *this },
 	gamestate_updated { gamestate_updated_callback ? std::move(gamestate_updated_callback) : []() {} } {}
 
 void InstanceManager::set_gamestate_needs_update() {
@@ -200,6 +202,18 @@ bool InstanceManager::update_clock() {
 	}
 
 	simulation_clock.conditionally_advance_game();
+	return true;
+}
+
+bool InstanceManager::set_today_and_update(Date new_today) {
+	if (!is_game_session_started()) {
+		Logger::error("Cannot update clock - game session not started!");
+		return false;
+	}
+
+	today = new_today;
+	gamestate_needs_update = true;
+	update_gamestate();
 	return true;
 }
 
--- a/src/openvic-simulation/console/ConsoleInstance.cpp
+++ b/src/openvic-simulation/console/ConsoleInstance.cpp
@@ -1,0 +1,381 @@
+#include "openvic-simulation/console/ConsoleInstance.hpp"
+
+#include <cctype>
+#include <charconv>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <string_view>
+
+#include "openvic-simulation/DefinitionManager.hpp"
+#include "openvic-simulation/InstanceManager.hpp"
+#include "openvic-simulation/country/CountryInstance.hpp"
+#include "openvic-simulation/map/ProvinceInstance.hpp"
+#include "openvic-simulation/military/UnitType.hpp"
+#include "openvic-simulation/military/Wargoal.hpp"
+#include "openvic-simulation/misc/Event.hpp"
+#include "openvic-simulation/research/Invention.hpp"
+#include "openvic-simulation/research/Technology.hpp"
+#include "openvic-simulation/types/Colour.hpp"
+#include "openvic-simulation/types/Date.hpp"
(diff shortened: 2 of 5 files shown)


wvpm (dismissed):
Only update date, don't tick when using console commands

src/openvic-simulation/types/Colour.hpp
```diff
@@ -325,6 +326,28 @@ namespace OpenVic {
...
+		constexpr basic_colour_t contrast() const {
+			constexpr double LUMINANCE_FLIP = 0.342;
+			constexpr double POWER = 2.4;
+			constexpr double RED_MULTIPLIER = 0.2126729;
+			constexpr double GREEN_MULTIPLIER = 0.7151522;
+			constexpr double BLUE_MULTIPLIER = 0.0721750;
+
+			double luminance = std::pow(red / max_value, POWER) * RED_MULTIPLIER +
+				std::pow(green / max_value, POWER) * GREEN_MULTIPLIER + std::pow(blue, POWER) * BLUE_MULTIPLIER;
+			return luminance >= LUMINANCE_FLIP ? basic_colour_t::from_floats(0, 0 ,0) : basic_colour_t::from_floats(1, 1, 1);
```
Hop311, line 348:
```suggestion
			return luminance >= LUMINANCE_FLIP ? basic_colour_t::from_floats(0, 0, 0) : basic_colour_t::from_floats(1, 1, 1);
```
Formatting

src/openvic-simulation/InstanceManager.cpp
```diff
@@ -203,6 +205,17 @@ bool InstanceManager::update_clock() {
...
 }
 
+bool InstanceManager::set_today_and_update(Date new_today) {
+	if (!is_game_session_started()) {
+		Logger::error("Cannot update clock - game session not started!");
+		return false;
+	}
+
+	today = new_today;
+	simulation_clock.conditionally_advance_game();
```
Hop311, line 215: This won't necessarily change anything, you'll want to `set_gamestate_needs_update();` before calling it to make sure stuff actually updates (this won't guarantee a tick, but I don't think we want that as it would change the day). TBH, you could just call `update_gamestate()` directly and ignore the simulation clock altogether, it's entirely based on machine clock time so it doesn't care about changes to the SIM date.

src/openvic-simulation/console/ConsoleCommand.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+namespace OpenVic {
+	struct ConsoleInstance;
+
+	struct ConsoleCommand {
+		friend struct ConsoleInstance;
+
+		struct Argument {
+			ConsoleInstance& console;
+			std::span<std::string_view> arguments;
```
Hop311, line 15: This file could do with a `#include <string_view>`

src/openvic-simulation/console/ConsoleInstance.hpp
```diff
@@ -0,0 +1,111 @@
...
+		bool setup_commands();
+
+	public:
+		InstanceManager& get_instance_manager() {
+			return instance_manager;
+		}
+
+		InstanceManager const& get_instance_manager() const {
+			return instance_manager;
+		}
```
Hop311, line 109: These could be generated instead by wrapping `PROPERTY_REF` around the variable declaration.

src/openvic-simulation/console/ConsoleInstance.cpp
```diff
@@ -0,0 +1,362 @@
...
+			break;
+		}
+
+		// Skip the closing quote
+		if (quote && *it == quote) {
+			it++;
+		}
+	}
+
+	decltype(commands)::iterator cmd_it = commands.find(arguments[0]);
```
Hop311, line 128: Are we sure `arguments` won't be empty, or are we fine with accessing the uninitialised reserved space if it is empty?

src/openvic-simulation/console/ConsoleInstance.cpp
```diff
@@ -0,0 +1,362 @@
...
+	bool success = true;
+	Date date = Date::from_string(value_string, &success, true);
+	if (!success) {
+		write_error("Invalid date");
+		return std::nullopt;
+	}
+	return date;
+}
+
+Invention const* ConsoleInstance::validate_invention(std::string_view value_string) {
```
Hop311, line 292: Why not check for empty input when validating inventions, CBs and tech? Integers (signed and unsigned) and booleans also don't have an empty check, but all other implemented validation functions do. TBF it's not necessary to check for any of the IdentifierRegistry-based validators, as we always disallow empty identifiers

Hop311 (dismissed):
Great work, very elegant 💯

src/openvic-simulation/console/ConsoleInstance.hpp
Spartan322: That causes a warning because InstanceManager is itself a reference, not a member value.

src/openvic-simulation/console/ConsoleInstance.cpp
Spartan322: The integers and booleans automatically reject empty values as invalid.

src/openvic-simulation/console/ConsoleInstance.hpp
Hop311: Ahhh ok, maybe we should add an equivalent to `PROPERTY_PTR` which, when applied to a non-const pointer variable, creates non-const pointer and const pointer getters

Hop311 (dismissed):

wvpm (approved):

PR #295, opened 2025-02-04
Support gold mines & report production

wvpm:
Gold mines refers to any production type that outputs a good with `money = yes`. These are directly converted to cash and not traded.

Also both artisans & rgo now report their daily production to the province controller (if any).
The country can sum the total production and derive the gold income (in budget) from this.

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.cpp
+++ b/src/openvic-simulation/country/CountryInstance.cpp
@@ -1382,6 +1382,19 @@ void CountryInstance::tick(InstanceManager& instance_manager) {
 	if (leadership_point_stockpile > max_leadership_point_stockpile) {
 		leadership_point_stockpile = max_leadership_point_stockpile;
 	}
+
+	//TODO add gold income from national money goods production
+	//income_from_gold = country_defines.gold_to_cash_rate * sum (total_good_production * good.get_base_price()) for each good with is_money=true
+}
+
+void CountryInstance::report_artisan_output(ProductionType const& production_type, const fixed_point_t quantity) {
+	//TODO record artisan output
+}
+void CountryInstance::report_factory_output(ProductionType const& production_type, const fixed_point_t quantity) {
+	//TODO record factory output
+}
+void CountryInstance::report_rgo_output(GoodDefinition const& good, const fixed_point_t quantity) {
+	//TODO record rgo output
 }
 
 void CountryInstanceManager::update_rankings(Date today, DefineManager const& define_manager) {
--- a/src/openvic-simulation/utility/StringUtils.hpp
+++ b/src/openvic-simulation/utility/StringUtils.hpp
@@ -176,6 +176,10 @@ namespace OpenVic::StringUtils {
 		return result;
 	}
 
+	inline constexpr std::string_view bool_to_yes_no(bool b) {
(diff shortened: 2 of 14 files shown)


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1382,6 +1382,19 @@ void CountryInstance::tick(InstanceManager& instance_manager) {
 	if (leadership_point_stockpile > max_leadership_point_stockpile) {
 		leadership_point_stockpile = max_leadership_point_stockpile;
 	}
+
+	//TODO add gold income from national money goods production
+	//income_from_gold = country_defines.gold_to_cash_rate * sum (total_good_production * good.get_base_price()) for each good with is_money=true
+}
+
+void CountryInstance::report_artisan_output(GoodDefinition const& good, const fixed_point_t quantity) {
```
wvpm, line 1390: Artisan production is reported per production type and not per good.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1382,6 +1382,19 @@ void CountryInstance::tick(InstanceManager& instance_manager) {
...
+	//TODO record artisan output
+}
+void CountryInstance::report_factory_output(GoodDefinition const& good, const fixed_point_t quantity) {
```
wvpm, line 1393: Factory production is reported per production type and not per good.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1382,6 +1382,19 @@ void CountryInstance::tick(InstanceManager& instance_manager) {
...
+	//TODO record factory output
+}
+void CountryInstance::report_rgo_output(GoodDefinition const& good, const fixed_point_t quantity) {
```
wvpm, line 1396: RGO production is reported per good as there is only 1 production type per good used.

schombert: this is true in vanilla. I'm not sure that it is necessarily true for all mods.

wvpm: It is true for all cases. You define a good as rgo in the province history. This good is mapped to a production type. You cannot specify which production type for the good you want. The game always maps a good to a certain production type. See https://github.com/OpenVicProject/OpenVic-Simulation/blob/master/src/openvic-simulation/economy/production/ProductionType.cpp#L180

Hop311 (approved):
Great work, I'll get around to dropping the tradeable check from the trade menu UI code

src/openvic-simulation/utility/StringUtils.hpp
```diff
@@ -176,6 +176,10 @@ namespace OpenVic::StringUtils {
 		return result;
 	}
 
+	inline constexpr std::string bool_to_yes_no(bool b) {
```
Hop311, line 179: Best to return either `std::string_view` here, `std::string` requires dynamically allocating memory and copying the raw `"yes"` or `"no"` data into it every time this function is called. Also I think this won't compile on Mac, it doesn't have good `constexpr` support for `std::string` (but it does for `std::string_view`)

Hop311 (approved):

PR #297, opened 2025-02-08
Pop needs & spending

wvpm:
- [x] Pops buy needs
- [x] Pops update their needs fullfilled percentages
- [x] Pops keep a record of fulfilled goods per need category
- [x] Move initial values to PROPERTY macro.
- [x] Count unlocked inventions on country instance.

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.cpp
+++ b/src/openvic-simulation/country/CountryInstance.cpp
@@ -671,7 +671,15 @@ bool CountryInstance::modify_invention_unlock(
 		return false;
 	}
 
+	const bool invention_was_unlocked = unlock_level > 0;
 	unlock_level += unlock_level_change;
+	if (invention_was_unlocked != (unlock_level > 0)) {
+		if (invention_was_unlocked) {
+			inventions_count--;
+		} else {
+			inventions_count++;
+		}
+	}
 
 	bool ret = true;
 
@@ -1389,14 +1397,24 @@ void CountryInstance::tick(InstanceManager& instance_manager) {
 	//income_from_gold = country_defines.gold_to_cash_rate * sum (total_good_production * good.get_base_price()) for each good with is_money=true
 }
 
+
+void CountryInstance::report_pop_need_consumption(PopType const& pop_type, GoodDefinition const& good, const fixed_point_t quantity) {
+	//TODO
+}
+void CountryInstance::report_artisan_input_consumption(ProductionType const& production_type, GoodDefinition const& good, const fixed_point_t quantity) {
+	//TODO
+}
 void CountryInstance::report_artisan_output(ProductionType const& production_type, const fixed_point_t quantity) {
-	//TODO record artisan output
+	//TODO
+}
+void CountryInstance::report_factory_input_consumption(ProductionType const& production_type, GoodDefinition const& good, const fixed_point_t quantity) {
+	//TODO
 }
 void CountryInstance::report_factory_output(ProductionType const& production_type, const fixed_point_t quantity) {
-	//TODO record factory output
+	//TODO
 }
 void CountryInstance::report_rgo_output(GoodDefinition const& good, const fixed_point_t quantity) {
-	//TODO record rgo output
+	//TODO
 }
 
 CountryInstance::good_data_t& CountryInstance::get_good_data(GoodInstance const& good_instance) {
--- a/src/openvic-simulation/economy/production/ArtisanalProducer.cpp
+++ b/src/openvic-simulation/economy/production/ArtisanalProducer.cpp
@@ -1,10 +1,7 @@
 #include "ArtisanalProducer.hpp"
 
-#include "openvic-simulation/economy/GoodDefinition.hpp"
+#include "openvic-simulation/economy/GoodInstance.hpp"
 #include "openvic-simulation/economy/production/ProductionType.hpp"
-#include "openvic-simulation/economy/trading/BuyResult.hpp"
-#include "openvic-simulation/economy/trading/MarketInstance.hpp"
-#include "openvic-simulation/economy/trading/SellResult.hpp"
 #include "openvic-simulation/map/ProvinceInstance.hpp"
 #include "openvic-simulation/modifier/ModifierEffectCache.hpp"
 #include "openvic-simulation/pop/Pop.hpp"
@@ -12,19 +9,24 @@
 using namespace OpenVic;
 
 ArtisanalProducer::ArtisanalProducer(
-	MarketInstance& new_market_instance,
+	GoodInstanceManager const& new_good_instance_manager,
 	ModifierEffectCache const& new_modifier_effect_cache,
 	GoodDefinition::good_definition_map_t&& new_stockpile,
 	ProductionType const& new_production_type,
 	fixed_point_t new_current_production
-) : market_instance { new_market_instance },
+) : good_instance_manager { new_good_instance_manager },
 	modifier_effect_cache { new_modifier_effect_cache },
 	stockpile { std::move(new_stockpile) },
 	production_type { new_production_type },
 	current_production { new_current_production }
-	{}
+	{
+		max_quantity_to_buy_per_good.reserve(new_production_type.get_input_goods().size());
+	}
 
 void ArtisanalProducer::artisan_tick(Pop& pop) {
+	max_quantity_to_buy_per_good.clear();
+
+	//TODO get from pool instead of create & destroy each time
 	GoodDefinition::good_definition_map_t goods_to_buy_and_max_price { };
 	GoodDefinition::good_definition_map_t demand { };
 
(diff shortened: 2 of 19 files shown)


src/openvic-simulation/pop/Pop.cpp
```diff
@@ -273,9 +277,17 @@ void Pop::pop_tick() {
...
 	if (artisanal_producer_nullable != nullptr) {
 		artisanal_producer_nullable->artisan_tick(*this);
 	}
+
+	const fixed_point_t total_cash_to_spend = cash - cash_allocated_for_artisanal_spending;
+	//see vic2 consumer python
+}
+
+void Pop::allocate_for_artisanal_spending(const fixed_point_t amount) {
+	cash_allocated_for_artisanal_spending = amount;
```
Hop311, line 290: Should this definitely be `=` rather than `+=`?

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -227,6 +207,30 @@ std::stringstream Pop::get_pop_context_text() const {
 	return pop_context;
 }
 
+void Pop::fill_needs_fulfilled_goods() {
+	if(type == nullptr) {
```
Hop311, line 211: Space formatting here and in the `[X,Y]` code below, but regardless this `type == nullptr` check shouldn't be necessary, we always initialise `type` with a non-null `PopType const&` and the only way it can change is being swapped with an equivalent non-null `PopType`

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -50,7 +50,7 @@ namespace OpenVic {
 		F(artisan_inputs_expense)
 
 	#define DECLARE_POP_MONEY_STORES(money_type)\
-		fixed_point_t PROPERTY(money_type);
+		fixed_point_t PROPERTY(money_type, fixed_point_t::_0());
```
Hop311, line 53: Fixed points technically don't need explicit instantiation values, they'll happily default initialise to 0 via their default constructor unless explicitly told to do something else.

src/openvic-simulation/pop/PopValuesFromProvince.cpp
```diff
@@ -0,0 +1,56 @@
...
+	ModifierEffectCache::strata_effects_t const& strata_effects = modifier_effect_cache.get_strata_effects()[strata];
+	PopsDefines const& defines = parent.get_defines();
+	fixed_point_t shared_base_needs_scalar = defines.get_base_goods_demand()
+		* province.get_modifier_effect_value(*modifier_effect_cache.get_goods_demand());
+
+	fixed_point_t invention_needs_scalar = fixed_point_t::_1();
+	CountryInstance const* const owner_nullable = province.get_owner();
+	if(owner_nullable != nullptr) {
+		CountryInstance const& owner = *owner_nullable;
+		shared_base_needs_scalar *= fixed_point_t::_1() + owner.get_plurality();
```
Hop311, line 23:
```suggestion
		shared_base_needs_scalar *= fixed_point_t::_1() + owner.get_plurality() / 100;
```
`owner.get_plurality` is stored on a 0-100 range rather than 0.00-1.00, so you'll need to normalise it before adding it to `1`

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -669,6 +669,13 @@ bool CountryInstance::modify_invention_unlock(
 		return false;
 	}
 
+	if (unlock_level == 0) {
+		if (unlock_level_change > 0) {
+			inventions_count++;
+		}
+	} else if (unlock_level == -unlock_level_change) {
```
Hop311, line 676: We can't assume `unlock_level` is only `0`, `1` or `-1` here, technically we could have higher values e.g. from an invention being enabled with a value > 1 or multiple times in country history

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -273,9 +278,118 @@ void Pop::pop_tick() {
...
+			price_inverse_sum += good_instance.get_price_inverse();
+		}
+		
+		for (auto [good_definition,quantity] : type_never_null.get_life_needs()) {
+			GoodInstance const& good_instance = good_instance_manager.get_good_instance_from_definition(*good_definition);
+			fixed_point_t cash_to_spend_on_good = cash_left_to_spend * good_instance.get_price_inverse() / price_inverse_sum;
+			if (cash_to_spend_on_good > quantity * good_instance.get_max_next_price()) {
+				//spend max, remove from price_inverse_sum and recalculate the rest.
+			}
+		}
```
Hop311, line 344: Spacing: `[good_definition, quantity]`

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -6,14 +6,14 @@
...
 #include "openvic-simulation/economy/production/ResourceGatheringOperation.hpp"
-#include "openvic-simulation/economy/trading/MarketInstance.hpp"
 #include "openvic-simulation/military/UnitInstance.hpp"
 #include "openvic-simulation/military/UnitType.hpp"
 #include "openvic-simulation/modifier/ModifierSum.hpp"
 #include "openvic-simulation/pop/Pop.hpp"
 #include "openvic-simulation/types/FlagStrings.hpp"
 #include "openvic-simulation/types/HasIdentifier.hpp"
 #include "openvic-simulation/types/OrderedContainers.hpp"
+#include "pop/PopValuesFromProvince.hpp"
```
Hop311, line 16:
```suggestion
#include "openvic-simulation/pop/PopValuesFromProvince.hpp"
```

src/openvic-simulation/pop/PopValuesFromProvince.cpp
```diff
@@ -0,0 +1,49 @@
...
+	ModifierEffectCache::strata_effects_t const& strata_effects = modifier_effect_cache.get_strata_effects()[strata];
+	PopsDefines const& defines = parent.get_defines();
+	fixed_point_t shared_base_needs_scalar = defines.get_base_goods_demand()
+		* province.get_modifier_effect_value(*modifier_effect_cache.get_goods_demand());
+
+	fixed_point_t invention_needs_scalar = fixed_point_t::_1();
+	CountryInstance const* const owner_nullable = province.get_owner();
+	if(owner_nullable != nullptr) {
+		CountryInstance const& owner = *owner_nullable;
+		shared_base_needs_scalar *= fixed_point_t::_1() + owner.get_plurality() / fixed_point_t::_100();
```
Hop311, line 22: It's better to use normal `int32_t`s for simple integer multiplication or division with fixed points, it can be evaluated with no shifting required unlike when using other fixed points, even constants. Addition, subtraction and ordering comparison operators on the other hand are faster with two fixed points than mixed types, as integers need to be shifted up to the same level as the fixed point's internal value

src/openvic-simulation/country/CountryInstance.cpp
wvpm: I don't assume that.
I check if == 0 (it's locked) && change > 0 (it will be unlocked) then ++.
Else (so it was unlocked as negative values aren't allowed) if level = -change (so it will become 0) --.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -669,6 +669,13 @@ bool CountryInstance::modify_invention_unlock(
 		return false;
 	}
 
+	if (unlock_level <= 0) {
+		if (unlock_level_change > unlock_level) {
```
Hop311, line 673: Here `unlock_level <= 0`, so it'll be less than any positive change. Instead we need to compare `unlock_level_change` to `-unlock_level `

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -669,14 +669,15 @@ bool CountryInstance::modify_invention_unlock(
 		return false;
 	}
 
-	if (unlock_level <= 0) {
-		if (unlock_level_change > unlock_level) {
+	const bool invention_was_unlocked = is_invention_unlocked(invention);
+	unlock_level += unlock_level_change;
+	if (invention_was_unlocked != is_invention_unlocked(invention)) {
```
Hop311, line 674: `is_invention_unlocked` does require an IndexedMap lookup, which is admitedly is just an addition,. but we can skip it by using `unlock_level > 0` before and after adding `unlock_level_change` to it instead

Hop311 (approved):

Hop311 (approved):
Just needs to be squashed and this branch will be ready to merge 👍

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -350,8 +349,8 @@ void ResourceGatheringOperation::pay_employees(
 
 	total_owner_income_cache = 0;
 	total_employee_income_cache = 0;
-	if (revenue <= 0 || total_worker_count_in_province <= 0) {
-		if (revenue < 0) { Logger::error("Negative revenue for province ", location.get_identifier()); }
+	if (revenue <= fixed_point_t::_0() || total_worker_count_in_province <= 0) {
+		if (revenue < fixed_point_t::_0()) { Logger::error("Negative revenue for province ", location.get_identifier()); }
 		if (total_worker_count_in_province < 0) { Logger::error("Negative total worker count for province ", location.get_identifier()); }
```
Hop311, line 354:
```suggestion
		if (revenue < fixed_point_t::_0()) {
			Logger::error("Negative revenue for province ", location.get_identifier());
		}
		if (total_worker_count_in_province < 0) {
			Logger::error("Negative total worker count for province ", location.get_identifier());
		}
```
Better to always break for `{}` blocks, makes the code easier to read

src/openvic-simulation/economy/production/ArtisanalProducer.cpp
```diff
@@ -1,150 +1,157 @@
...
+	{
+		max_quantity_to_buy_per_good.reserve(new_production_type.get_input_goods().size());
+	}
 
 void ArtisanalProducer::artisan_tick(Pop& pop) {
+	max_quantity_to_buy_per_good.clear();
+
+	//TODO get from pool instead of create & destroy each time
 	GoodDefinition::good_definition_map_t goods_to_buy_and_max_price { };
 	GoodDefinition::good_definition_map_t demand { };
```
Hop311, line 31: These could be made member variables, which would be more performant than local function variables or pooled memory in the long run (the expensive part is (re)allocating, the absolute amount of memory used is far less of an issue, so returning memory to the pool while these functions aren't running is likely a net loss performance-wise compared to always keeping the memory reserved for these maps)

src/openvic-simulation/economy/production/ArtisanalProducer.cpp
```diff
@@ -1,150 +1,157 @@
...
-						[this, &pop, &good_stockpile](const BuyResult buy_result) -> void {
-							if (buy_result.get_quantity_bought() > fixed_point_t::_0()) {
-								pop.add_artisan_inputs_expense(buy_result.get_money_spent());
-								good_stockpile += buy_result.get_quantity_bought();
-							}
-						}
-					});
+				const fixed_point_t optimal_quantity = demand[input_good_ptr] * max_possible_satisfaction_numerator / max_possible_satisfaction_denominator;
+				if (stockpile[input_good_ptr] >= optimal_quantity) {
+					goods_to_buy_and_max_price.erase(input_good_ptr);
```
Hop311, line 123: I'm not sure if it's safe to erase map elements while iterating through them 😕

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -90,23 +102,37 @@ namespace OpenVic {
...
+		#define NEED_MEMBERS(need_category)\
+			fixed_point_t need_category##_needs_acquired_quantity, need_category##_needs_desired_quantity;\
+			constexpr fixed_point_t get_##need_category##_needs_fulfilled() const {\
+				if (need_category##_needs_desired_quantity == fixed_point_t::_0()) {\
+					return fixed_point_t::_1();\
+				}\
+				return need_category##_needs_acquired_quantity / need_category##_needs_desired_quantity;\
+			}\
+			GoodDefinition::good_definition_map_t need_category##_needs {};\
+			ordered_map<GoodDefinition const*, bool> PROPERTY(need_category##_needs_fulfilled_goods);
```
Hop311, line 115:
```suggestion
			fixed_point_t need_category##_needs_acquired_quantity, need_category##_needs_desired_quantity; \
		public: \
			constexpr fixed_point_t get_##need_category##_needs_fulfilled() const { \
				if (need_category##_needs_desired_quantity == fixed_point_t::_0()) { \
					return fixed_point_t::_1(); \
				} \
				return need_category##_needs_acquired_quantity / need_category##_needs_desired_quantity; \
			} \
		private: \
			GoodDefinition::good_definition_map_t need_category##_needs {}; \
			ordered_map<GoodDefinition const*, bool> PROPERTY(need_category##_needs_fulfilled_goods);
```
`get_##need_category##_needs_fulfilled()` needs to be public so that the population menu UI code can access it (also better to have spaces before `\`s at the end of lines)

Hop311 (commented):
Great work! Various comments, also we still get a lot of log spam when passing time in-game:
- Mostly this kind of warning:
```
WARNING: Pop.cpp(274): Adding artisan_inputs_expense of 0 to pop. Context location: 1129 type: artisans culture: persian religion: shiite size: 1400
```
- But also a few of these kind of errors:
```
ERROR: Pop.cpp(274): Total cash became negative (-0.008636474609375) after adding artisan_inputs_expense of 0.019866943359375 to pop. Context location: 1055 type: artisans culture: north_caucasian religion: sunni size: 850
```

src/openvic-simulation/economy/production/ArtisanalProducer.cpp
wvpm: Out of scope for this PR.

src/openvic-simulation/pop/PopValuesFromProvince.cpp
```diff
@@ -0,0 +1,53 @@
...
+	fixed_point_t invention_needs_scalar = fixed_point_t::_1();
+	CountryInstance const* const owner_nullable = province.get_owner();
+	if (owner_nullable != nullptr) {
+		CountryInstance const& owner = *owner_nullable;
+		shared_base_needs_scalar *= fixed_point_t::_1() + owner.get_plurality() / 100;
+		invention_needs_scalar += owner.get_inventions_count() * defines.get_invention_impact_on_demand();
+	}
+
+	fixed_point_t shared_life_needs_scalar = shared_base_needs_scalar
+		* (fixed_point_t::_1() + province.get_modifier_effect_value(*strata_effects.get_life_needs()));
```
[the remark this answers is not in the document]
wvpm: Correct.

src/openvic-simulation/pop/PopValuesFromProvince.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "openvic-simulation/pop/PopType.hpp"
+#include "openvic-simulation/types/IndexedMap.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/utility/Getters.hpp"
+
+namespace OpenVic {
+	struct ModifierEffectCache;
+	struct PopManager;
```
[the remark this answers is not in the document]
wvpm: Leftovers. Thanks.

wvpm:
> Great work! Various comments, also we still get a lot of log spam when passing time in-game:
> 
>     * Mostly this kind of warning:
> 
> 
> ```
> WARNING: Pop.cpp(274): Adding artisan_inputs_expense of 0 to pop. Context location: 1129 type: artisans culture: persian religion: shiite size: 1400
> ```
> 
>     * But also a few of these kind of errors:
> 
> 
> ```
> ERROR: Pop.cpp(274): Total cash became negative (-0.008636474609375) after adding artisan_inputs_expense of 0.019866943359375 to pop. Context location: 1055 type: artisans culture: north_caucasian religion: sunni size: 850
> ```

Worrysome but out of scope for this PR.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -273,10 +282,156 @@ void Pop::pop_tick() {
...
+
+	PopType const& type_never_null = *type;
+	ProvinceInstance const& location_never_null = *location;
+	PopValuesFromProvince const& shared_values = location_never_null.get_shared_pop_values();
+	PopStrataValuesFromProvince const& shared_strata_values = shared_values.get_effects_per_strata()[type_never_null.get_strata()];
+	PopsDefines const& defines = shared_values.get_defines();
+	const fixed_point_t base_needs_scalar = (
+		fixed_point_t::_1() + 2 * consciousness / defines.get_pdef_base_con()
+	) * size;
+	constexpr int32_t size_denominator = 200000;
```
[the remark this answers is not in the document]
wvpm: Yes sadly.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -273,10 +282,156 @@ void Pop::pop_tick() {
...
+				const fixed_point_t max_money_to_spend = max_quantity_to_buy * good_instance.get_max_next_price();\
+				if (money_to_spend_per_good_draft[good_definition] >= max_money_to_spend) {\
+					continue;\
+				}\
+				fixed_point_t cash_available_for_good = cash_left_to_spend * good_instance.get_price_inverse() / need_category##_needs_price_inverse_sum;\
+				if (cash_available_for_good >= max_money_to_spend) {\
+					cash_left_to_spend -= max_money_to_spend;\
+					money_to_spend_per_good_draft[good_definition] = max_money_to_spend;\
+					need_category##_needs_price_inverse_sum -= good_instance.get_price_inverse();\
+					i = -1;\
```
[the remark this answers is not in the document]
wvpm: It processes all goods again but skips the goods that are maxed out already.
So we only redistribute the remaining cash over the remaining goods.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -273,10 +282,156 @@ void Pop::pop_tick() {
...
+				money_to_spend_per_good[good_definition] += money_to_spend;\
+			}\
+		}
+
+	DO_FOR_ALL_NEED_CATEGORIES(ALLOCATE_FOR_NEEDS)
+	#undef ALLOCATE_FOR_NEEDS
+
+	for (auto [good_definition, money_to_spend] : money_to_spend_per_good) {
+		const fixed_point_t max_quantity_to_buy = max_quantity_to_buy_per_good[good_definition];
+		market_instance.place_buy_up_to_order({
```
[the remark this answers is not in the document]
wvpm: The code filling max_quantity_to_buy_per_good only adds entries that have size > 0.
Still I'll add a check + error.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -273,10 +282,156 @@ void Pop::pop_tick() {
...
+						add_artisan_inputs_expense(expense);
+					}
+				}
+
+				#define CONSUME_NEED(need_category)\
+					if (quantity_left_to_consume <= fixed_point_t::_0()) {\
+						return;\
+					}\
+					if (need_category##_needs.contains(good_definition)) {\
+						const fixed_point_t desired_quantity = need_category##_needs[good_definition];\
```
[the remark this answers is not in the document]
wvpm: This seems like the [Dictionary.TryGet](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=net-9.0) I was looking for. I'll use that.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -273,10 +282,156 @@ void Pop::pop_tick() {
...
+			}
+		});
+	}
+
+	if (artisanal_produce_left_to_sell > fixed_point_t::_0()) {
+		market_instance.place_market_sell_order({
+			artisanal_producer_nullable->get_production_type().get_output_good(),
+			artisanal_produce_left_to_sell,
+			[this](const SellResult sell_result) -> void {
+				if (sell_result.get_money_gained() > fixed_point_t::_0()) {
```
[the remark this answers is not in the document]
wvpm: I don't understand your first point.
As for the > 0 checks, it's totally valid to gain 0 money from selling.
If I remove the check, we get a warning in the pop of "Adding artisanal_expense of 0 to pop."

src/openvic-simulation/economy/production/ArtisanalProducer.cpp
wvpm: Nice catch. A similar piece of code works fine as the loop uses a different collection.
Here it would probably screw things up.

src/openvic-simulation/pop/Pop.cpp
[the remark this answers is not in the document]
Hop311: The check:
```cpp
if (money_to_spend_per_good_draft[good_definition] >= max_money_to_spend) {
```
so if we can't guaruntee `max_money_to_spend` is always greater than 0 and every good's entry in that loop is later given a non-zero value, we could end up with entries with a value of 0

src/openvic-simulation/economy/production/ArtisanalProducer.cpp
Hop311: There's still a risk that the iterators get invalidated here, in general you can't trust any iterators to remain valid after adding or removing an element from a container unless the container specifically guarantees it. The safe approach is to use `erase_if`:
```suggestion
			erase_if(
				goods_to_buy_and_max_price,
				[
					this, &demand, max_possible_satisfaction_numerator, max_possible_satisfaction_denominator,
					&at_or_below_optimum
				](auto const& pair) {
					GoodDefinition const* const input_good_ptr = pair.first;
					const fixed_point_t optimal_quantity =
						demand[input_good_ptr] * max_possible_satisfaction_numerator / max_possible_satisfaction_denominator;
					if (stockpile[input_good_ptr] >= optimal_quantity) {
						at_or_below_optimum = false;
						return true;
					}
					return false;
				}
```

src/openvic-simulation/pop/Pop.hpp
Hop311: `get_##need_category##_needs_fulfilled()` still needs to be public, add `public:` above it and `private` below it in the macro

Hop311 (approved):

PR #298, opened 2025-02-08
Replace MSVC STL deque to support 512 block size

Spartan322:
This addresses https://github.com/microsoft/STL/issues/147#issuecomment-1090148740

### DIFF
--- a/src/openvic-simulation/utility/Deque.hpp
+++ b/src/openvic-simulation/utility/Deque.hpp
@@ -1,0 +1,1910 @@
+// deque standard header
+
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#pragma once
+#if __has_include(<__msvc_string_view.hpp>)
+
+#include <yvals_core.h>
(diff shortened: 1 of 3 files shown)


src/openvic-simulation/utility/Deque.hpp
```diff
@@ -0,0 +1,1910 @@
...
+
+		static constexpr int _Block_size = _Mydeque::_Block_size;
+
+	public:
+		using iterator_category = random_access_iterator_tag;
+
+		using value_type = typename _Mydeque::value_type;
+		using difference_type = typename _Mydeque::difference_type;
+		using pointer = typename _Mydeque::const_pointer;
+		using reference = const value_type&;
```
Hop311, line 40:
```suggestion
		using reference = value_type const&;
```
Non-standard `const&` and `const*` formatting here and throughout this file

Hop311 (approved):

src/openvic-simulation/utility/Deque.hpp
Spartan322: I'm trying to leave it as default to the MSVC STL as possible.

PR #299, opened 2025-02-09
Add proper clicksound Button support

Spartan322:


src/openvic-simulation/interface/UI.hpp
```diff
@@ -13,10 +14,14 @@ namespace OpenVic {
 
 		NamedInstanceRegistry<GUI::Scene, UIManager const&> IDENTIFIER_REGISTRY(scene);
 
+		DefinitionManager& PROPERTY_MOD(definition_manager);
```
Hop311, line 17: Why not just use `DefinitionManager const& PROPERTY(definition_manager);`, it's only used to get a `SoundEffect const*` so mutable access isn't required

src/openvic-simulation/utility/Getters.hpp
```diff
@@ -182,6 +197,8 @@ namespace OpenVic {
 	}
 }
 
+
+
```
Hop311, line 201: Are these extra newlines intentional? I've generally avoided having more than one empty line at a time in C++ files

src/openvic-simulation/utility/Getters.hpp
```diff
@@ -146,6 +146,21 @@ public: \
 	} \
 	ACCESS:
 
+/* Create const and non-const getters for a variable, applied to its name in its declaration, e
+ * for example: GameManager& PROPERTY_MOD(game_manager); */
+#define PROPERTY_MOD(NAME) PROPERTY_MOD_FULL(NAME, private)
+#define PROPERTY_MOD_FULL(NAME, ACCESS) \
```
Hop311, line 152: You could add a static assert to ensure this is being used on the right type (a non-const reference variable), similar to what I did for `PROPERTY_PTR`

src/openvic-simulation/DefinitionManager.hpp
```diff
@@ -19,10 +19,10 @@
 #include "openvic-simulation/pop/PopType.hpp"
 #include "openvic-simulation/research/ResearchManager.hpp"
 #include "openvic-simulation/scripts/ScriptManager.hpp"
+#include "openvic-simulation/interface/UI.hpp"
 
 namespace OpenVic {
-	struct DefinitionManager {
-	private:
+	class DefinitionManager {
```
Hop311, line 25: Class and struct (forward) declarations aren't compatible with MSVC, hence the link errors with the current commit. To fix this you need to do these things (or in each case you could do the opposite, replacing all declarations of the other variety):
- change the `struct Dataloader;` in `SoundEffect.hpp` to `class Dataloader;`
- change the `struct UIManager;` in `Dataloader.hpp` to `class UIManager;`
- change the `class DefinitionManager`s in `DefinitionManager.hpp` and `UI.hpp` to `struct DefinitionManager`

src/openvic-simulation/misc/SoundEffect.hpp
```diff
@@ -1,31 +1,34 @@
...
 	struct SoundEffect : HasIdentifier {
 		friend struct SoundEffectManager;
 
 	private:
-		std::string PROPERTY(file);
+		std::filesystem::path PROPERTY(file);
 		fixed_point_t PROPERTY(volume);
 
-		SoundEffect(std::string_view new_identifier, std::string_view new_file, fixed_point_t new_volume);
+		SoundEffect(std::string_view new_identifier, std::filesystem::path new_file, fixed_point_t new_volume);
```
Hop311, line 21: Should this use `std::filesystem::path const&` to avoid making unnecessary intermediate copies?

src/openvic-simulation/misc/SoundEffect.cpp
```diff
@@ -1,38 +1,48 @@
...
 	)(root);
 
 	if (sfx_identifier.empty()) {
 		Logger::error("Invalid sound identifier - empty!");
 		return false;
 	}
 	if (file.empty()) {
-		Logger::error("Invalid sound file name - empty!");
-		return false;
+		Logger::warning("Sound filename ", sfx_identifier, " was empty!");
```
Hop311, line 35: Maybe specify lookup failed rather than saying the filename was empty? If you cache `std::string_view val` from `file_callback` you could include that in this error message

src/openvic-simulation/DefinitionManager.hpp
Spartan322: I'm pretty certain that's not standard compliant, MSVC once again being super dumb.

src/openvic-simulation/misc/SoundEffect.cpp
Spartan322: The callback already reports that, there is technically the capability for it to be empty for other reasons.

src/openvic-simulation/misc/SoundEffect.hpp
Spartan322: What should be done is a move, a reference would still need to be copied otherwise it turns into a dangling reference.

src/openvic-simulation/DefinitionManager.hpp
Spartan322: Turns out its because they issue different mangled names to struct vs class, so the compiler treats it correctly by giving it the wrong name, but the linker has no idea what is going on. I really have to question why clang has a warning for this and MSVC doesn't say anything to warn you about this.

src/openvic-simulation/interface/UI.hpp
```diff
@@ -13,10 +14,14 @@ namespace OpenVic {
 
 		NamedInstanceRegistry<GUI::Scene, UIManager const&> IDENTIFIER_REGISTRY(scene);
 
+		DefinitionManager const& PROPERTY(definition_manager);
+
 		bool _load_font(ast::NodeCPtr node);
 		NodeTools::NodeCallback auto _load_fonts(std::string_view font_key);
 
 	public:
+		UIManager(DefinitionManager& manager);
```
Hop311, line 23:
```suggestion
		UIManager(DefinitionManager const& manager);
```
Same in `UI.cpp`

Hop311 (approved):

PR #300, opened 2025-02-10
Fix invalid clicksound pointer

Spartan322:


src/openvic-simulation/interface/GUI.hpp
```diff
@@ -146,7 +146,7 @@ namespace OpenVic::GUI {
 
 		fvec2_t PROPERTY(size);
 		fixed_point_t PROPERTY(rotation); /* In radians, usually one of 0, PI/2 or -PI/2. */
-		OpenVic::SoundEffect const* PROPERTY_REF(clicksound);
+		OpenVic::SoundEffect const* PROPERTY_REF(clicksound, nullptr);
```
Hop311, line 149: `PROPERTY_REF` generates `&` and `const&` getters, assuming we don't need to modify the address this variable points to and we don't need mutable access to the `SoundEffect` it points to, it should use `PROPERTY` rather than `PROPERTY_REF` or `PROPERTY_PTR`

Hop311 (approved):

PR #302, opened 2025-02-11
Rework military structs to have branch-agnostic bases and unique ids

Hop311:


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1397,6 +1449,19 @@ void CountryInstance::report_rgo_output(GoodDefinition const& good, const fixed_
 	//TODO record rgo output
 }
 
+CountryInstance::good_data_t& CountryInstance::get_good_data(GoodInstance const& good_instance) {
```
wvpm, line 1452: Again completely unrelated to the military rework. Avoid mixing PRs.
It looks like you have leftovers from some other branch mixed in.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -239,7 +241,7 @@ namespace OpenVic {
 		size_t PROPERTY(ship_count, 0);
 		fixed_point_t PROPERTY(total_consumed_ship_supply);
 		fixed_point_t PROPERTY(max_ship_supply);
-		fixed_point_t PROPERTY(leadership_point_stockpile);
+		fixed_point_t PROPERTY_RW(leadership_point_stockpile, 0);
```
wvpm, line 244: Why are some properties initialised with 0, I thought that was redundant with structs.

wvpm (commented):
Huge mixed bag of changes. I don't see the added value of the military change. Now there are switches everywhere :/

I recommend moving tech, economy and other non-military changes to different PRs.

src/openvic-simulation/military/UnitInstanceGroup.cpp
```diff
@@ -405,3 +403,109 @@ void UnitInstanceManager::tick() {
...
+		name,
+		branch,
+		creation_date,
+		personality,
+		background,
+		fixed_point_t::_0(), // 0 starting prestige
+		{} // No picture, will be set up by generate_leader
+	});
+
+	country.set_leadership_point_stockpile(country.get_leadership_point_stockpile() - leader_creation_cost);
```
wvpm, line 508: Move up to line 454 to keep it near the rest of the cost code.

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -83,8 +82,8 @@ namespace OpenVic {
 		Crime const* PROPERTY_RW(crime, nullptr);
 		ResourceGatheringOperation PROPERTY(rgo);
 		IdentifierRegistry<BuildingInstance> IDENTIFIER_REGISTRY(building);
-		ordered_set<ArmyInstance*> PROPERTY(armies);
-		ordered_set<NavyInstance*> PROPERTY(navies);
+		std::vector<ArmyInstance*> PROPERTY(armies);
```
wvpm, line 85: Consider merging them

src/openvic-simulation/country/CountryInstance.hpp
Hop311: Leftover from testing where I manually set it to 100

src/openvic-simulation/map/ProvinceInstance.hpp
Hop311: I'm going to leave them separate for now as I think it'll be useful to be able to iterate over the branches individually, but this rework does indeed allow us to merge them under a shared base pointer if we want

src/openvic-simulation/military/UnitInstanceGroup.hpp
```diff
@@ -140,37 +156,72 @@ namespace OpenVic {
...
+
+		LeaderInstance* get_leader_instance_by_unique_id(unique_id_t unique_id);
+		UnitInstance* get_unit_instance_by_unique_id(unique_id_t unique_id);
+		UnitInstanceGroup* get_unit_instance_group_by_unique_id(unique_id_t unique_id);
+
+		// Creates a new leader of the specified branch and adds it to the specified country. The leader's name and traits
+		// can be specified, but if they are not, the leader will be generated with a random name and traits. The country's
+		// leadership points will be checked and, if there are enough, have the leader creation cost subtracted from them.
+		// If the country does not have enough leadership points, the function will return false and no leader will be created.
+		bool create_leader(
```
wvpm, line 218: Move to CountryInstance or Leader?

src/openvic-simulation/military/UnitInstanceGroup.cpp
```diff
@@ -136,126 +155,143 @@ bool UnitInstanceGroup<Branch>::set_leader(_Leader* new_leader) {
...
-) {
-	unit_instance = &*get_unit_instances<Branch>().insert(
-		[&unit_deployment]() -> UnitInstanceBranched<Branch> {
+UnitInstanceBranched<Branch>& UnitInstanceManager::generate_unit_instance(UnitDeployment<Branch> const& unit_deployment) {
+	UnitInstanceBranched<Branch>& unit_instance = *get_unit_instances<Branch>().insert(
+		[this, &unit_deployment]() -> UnitInstanceBranched<Branch> {
 			if constexpr (Branch == LAND) {
 				return {
-					unit_deployment.get_name(), unit_deployment.get_type(),
+					unique_id_counter++,
```
wvpm, line 276: How does MP work here?

src/openvic-simulation/military/UnitInstanceGroup.cpp
```diff
@@ -405,3 +415,109 @@ void UnitInstanceManager::tick() {
...
+		];
+	}
+
+	if (background == nullptr && !leader_trait_manager.get_background_traits().empty()) {
+		background = leader_trait_manager.get_background_traits()[
+			item_selection_counter++ % leader_trait_manager.get_background_traits().size()
+		];
+	}
+
+	// TODO - should we abort is personality or background is still nullptr?
```
wvpm, line 509: Personality and background should fall back to empty values like Victoria 2 does.
I prefer setting the pointer to an empty instance here over nullptr as that would require all other code to do nullchecks.

src/openvic-simulation/military/UnitInstanceGroup.hpp
Hop311: It's here because leader instances are ultimately stored in `UnitInstanceManager`, along with units and unit groups, and it helps to keep this code in the same place as the starting/history leader generation code, with both making use of various manager references we store here for stuff like cultural leader picture lookup

src/openvic-simulation/military/UnitInstanceGroup.cpp
```diff
@@ -405,3 +415,109 @@ void UnitInstanceManager::tick() {
...
+	// Or even if name is empty? Should we add a default name or is empty fine?
```
wvpm, line 510: Preferably an empty name is treated as warning in the data loading and does not need to be handled here.
So returning an empty name seems fine to me.

src/openvic-simulation/military/UnitInstanceGroup.cpp
Hop311: All leaders/units/unit groups will be generated in lockstep across the host and clients, so in the exact same order with the exact same IDs

src/openvic-simulation/military/UnitInstanceGroup.cpp
Hop311: Yeah I've been thinking about how to handle all the `no*` things the base game uses, for now let's leave this as a nullptr and decide how to handle these `no*` cases across the board in a separate PR

src/openvic-simulation/military/UnitInstanceGroup.cpp
```diff
@@ -405,3 +415,109 @@ void UnitInstanceManager::tick() {
...
+
+	generate_leader(country, {
+		name,
+		branch,
+		creation_date,
+		personality,
+		background,
+		fixed_point_t::_0(), // 0 starting prestige
```
wvpm, line 518: Starting prestige is not 0 in Victoria 2.
It is a random value between 0 and defines.military.LEADER_MAX_RANDOM_PRESTIGE

wvpm (approved):

PR #309, opened 2025-02-13
rgo_output & rgo_throughput are country modifiers, not tech.

wvpm:
Tested in Victoria 2.


Hop311 (commented):
I tested in game and both `rgo_output` and `rgo_throughput` work in tech, they show up in tooltips and affect RGO production

wvpm:
Retested again and they do also work in tech. Hop311 is correct.

PR #310, opened 2025-02-14
Record production & consumption

wvpm:


src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -520,11 +526,8 @@ namespace OpenVic {
 		void tick(InstanceManager& instance_manager);
 
 		void report_pop_need_consumption(PopType const& pop_type, GoodDefinition const& good, const fixed_point_t quantity);
-		void report_artisan_input_consumption(ProductionType const& production_type, GoodDefinition const& good, const fixed_point_t quantity);
-		void report_artisan_output(ProductionType const& production_type, const fixed_point_t quantity);
-		void report_factory_input_consumption(ProductionType const& production_type, GoodDefinition const& good, const fixed_point_t quantity);
-		void report_factory_output(ProductionType const& production_type, const fixed_point_t quantity);
-		void report_rgo_output(GoodDefinition const& good, const fixed_point_t quantity);
+		void report_input_consumption(ProductionType const& production_type, GoodDefinition const& good, const fixed_point_t quantity);
+		void report_output(ProductionType const& production_type, const fixed_point_t quantity);
```
Hop311, line 530: These are not thread safe, causing the program to freeze or crash when they're simultaneously called from artisan or rgo ticks

wvpm: I've added locking and it works fine on my machine.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -209,6 +210,16 @@ namespace OpenVic {
 			fixed_point_t factory_needs;
 			fixed_point_t pop_needs;
 			fixed_point_t available_amount;
+
+			ordered_map<PopType const*, fixed_point_t> need_consumption_per_pop_type;
+			ordered_map<ProductionType const*, fixed_point_t> input_consumption_per_production_type;
+			ordered_map<ProductionType const*, fixed_point_t> production_per_production_type;
+
+			good_data_t() :
+				lock { std::make_unique<std::mutex>() }	{};
```
Hop311, line 219: You can move the implementation of this to the CPP file, also best to add back the `good_data_t(good_data_t&&) = default;` move constructor you had earlier

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1397,24 +1401,38 @@ void CountryInstance::tick(InstanceManager& instance_manager) {
...
-	//TODO
-}
-void CountryInstance::report_factory_output(ProductionType const& production_type, const fixed_point_t quantity) {
-	//TODO
-}
-void CountryInstance::report_rgo_output(GoodDefinition const& good, const fixed_point_t quantity) {
-	//TODO
+	good_data_t& good_data = get_good_data(good);
+	const std::lock_guard<std::mutex> lock_guard {*good_data.lock};
+	good_data.need_consumption_per_pop_type[&pop_type] += quantity;
```
Hop311, line 1425: I see that you put the lock guard after getting the good data, at first I was worried this could cause problems but after thinking some more I realise this is indeed a clever way to minimise the amount of work that is limited by the lock, allowing multiple calls to do the (thread safe) good data lookup while waiting for their turn to modify its data.

```suggestion
	good_data_t& good_data = get_good_data(good);
	const std::lock_guard<std::mutex> lock_guard { *good_data.lock };
	good_data.need_consumption_per_pop_type[&pop_type] += quantity;
```
Also not urgent but it would be nice to have spaces in the `{}` constructor

Hop311 (approved):

PR #312, opened 2025-02-14
Clean up

wvpm:


Spartan322 (commented):
This is a bad PR, the comments were put there for a reason, they are not to be removed unless you have fixed it so the formatter will leave these things alone.

PR #314, opened 2025-02-15
Extract trading from GoodInstance

wvpm:
Replacement for https://github.com/OpenVicProject/OpenVic-Simulation/pull/308


src/openvic-simulation/economy/trading/GoodMarket.hpp
```diff
@@ -0,0 +1,61 @@
...
+		std::vector<fixed_point_t> purchasing_power_per_order;
+
+		//only used during day tick (from actors placing order until execute_orders())
+		std::vector<GoodBuyUpToOrder> buy_up_to_orders;
+		std::vector<GoodMarketSellOrder> market_sell_orders;
+
+		protected:
+		bool PROPERTY(is_available);
+
+		private:
```
Hop311, line 36:
```suggestion
	protected:
		bool PROPERTY_ACCESS(is_available, protected);

	private:
```
- Indentation
- While it technically doesn't matter here as we switch back to `private` straight after, it's best to use `PROPERTY_ACCESS` to ensure the accessibility is still `protected` after declaring `is_available`. `PROPERTY` on its own changes the accessibility to `private`, and so can lead to accidental incorrect accesibility settings

Hop311 (dismissed):

src/openvic-simulation/economy/trading/GoodMarket.hpp
wvpm: Not a fan of PROPERTY_ACCESS since it does not alter the access for the property but only for what comes after it.
I couldn't find a better (and neither could gemini) so we're stuck with this it.

src/openvic-simulation/economy/GoodInstance.hpp
```diff
@@ -1,57 +1,20 @@
...
-		fixed_point_t PROPERTY(price_inverse);
-		fixed_point_t PROPERTY(price_change_yesterday);
-		fixed_point_t PROPERTY(max_next_price);
-		fixed_point_t PROPERTY(min_next_price);
-		bool PROPERTY(is_available);
-		fixed_point_t PROPERTY(total_demand_yesterday);
-		fixed_point_t PROPERTY(total_supply_yesterday);
-		fixed_point_t PROPERTY(quantity_traded_yesterday);
-		ValueHistory<fixed_point_t> PROPERTY(price_history);
+		GoodDefinition const& good_definition;
```
Hop311, line 15:
```suggestion
		GoodDefinition const& PROPERTY(good_definition);
```
This needs a getter, it's used in main repo UI code, as well as generally being part of the definition-instance architectural design.

Also could you add a newline between the constructor and `public` below here

Hop311 (approved):

PR #316, opened 2025-02-15
Fix pop need modifiers

wvpm:
- Assign properties instead of unused local variables.
- Multiply by `1 + modifier` instead of only the modifier.
- Correct local variable name
- Skip logic when max_quantity == 0
- Add space before \ (in macros)
- Reset country data, then do province ticks and then country ticks (otherwise country tick resets the data generated in province tick)
- Skip pop needs that aren't available yet.
- Renamed tick to country_tick so it's easier to find its uses.
- Added GoodDefinition to GoodMarket for easier debugging
- Extracted allocate_needs to a method instead of a macro for easier debugging.
- Update cash_left_to_spend even when a pop couldn't afford max price.
- In ArtisanalProducer::add_to_stockpile use .find() & .at() instead of indexing to allow parallel execution.

### DIFF
--- a/src/openvic-simulation/economy/trading/GoodMarket.hpp
+++ b/src/openvic-simulation/economy/trading/GoodMarket.hpp
@@ -17,7 +17,6 @@ namespace OpenVic {
 		std::unique_ptr<std::mutex> buy_lock;
 		std::unique_ptr<std::mutex> sell_lock;
 		GameRulesManager const& game_rules_manager;
-		const fixed_point_t base_price;
 		fixed_point_t absolute_maximum_price;
 		fixed_point_t absolute_minimum_price;
 
@@ -31,6 +30,7 @@ namespace OpenVic {
 		std::vector<GoodMarketSellOrder> market_sell_orders;
 
 	protected:
+		GoodDefinition const& PROPERTY_ACCESS(good_definition, protected);
 		bool PROPERTY_ACCESS(is_available, protected);
 
 	private:
@@ -46,7 +46,7 @@ namespace OpenVic {
 
 		void update_next_price_limits();
 	public:
-		GoodMarket(GameRulesManager const& new_game_rules_manager, fixed_point_t new_base_price, bool new_is_available);
+		GoodMarket(GameRulesManager const& new_game_rules_manager, GoodDefinition const& new_good_definition);
 		GoodMarket(GoodMarket&&) = default;
 
 		//thread safe
--- a/src/openvic-simulation/pop/Pop.cpp
+++ b/src/openvic-simulation/pop/Pop.cpp
@@ -216,7 +216,7 @@ std::stringstream Pop::get_pop_context_text() const {
 
 void Pop::reserve_needs_fulfilled_goods() {
 	PopType const& type_never_null = *type;
-	#define RESERVE_NEEDS(need_category)\
+	#define RESERVE_NEEDS(need_category) \
 		need_category##_needs_fulfilled_goods.reserve(type_never_null.get_##need_category##_needs().size());
 
 	DO_FOR_ALL_NEED_CATEGORIES(RESERVE_NEEDS)
@@ -225,9 +225,9 @@ void Pop::reserve_needs_fulfilled_goods() {
 
 void Pop::fill_needs_fulfilled_goods_with_false() {
 	PopType const& type_never_null = *type;
-	#define FILL_WITH_FALSE(need_category)\
+	#define FILL_WITH_FALSE(need_category) \
 		need_category##_needs_fulfilled_goods.clear(); \
-		for (auto [good, base_demand] : type_never_null.get_##need_category##_needs()) {\
+		for (auto [good, base_demand] : type_never_null.get_##need_category##_needs()) { \
 			need_category##_needs_fulfilled_goods.emplace(good, false); \
 		}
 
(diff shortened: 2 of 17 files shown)


src/openvic-simulation/pop/Pop.cpp
```diff
@@ -300,30 +300,36 @@ void Pop::pop_tick() {
...
-	CountryInstance* get_country_to_report_economy_nullable = location_never_null.get_country_to_report_economy();
-	#define FILL_NEEDS(need_category)\
+	CountryInstance* country_to_report_economy_nullable = location_never_null.get_country_to_report_economy();
+	#define FILL_NEEDS(need_category) \
 		need_category##_needs.clear(); \
 		const fixed_point_t need_category##_needs_scalar = base_needs_scalar * shared_strata_values.get_shared_##need_category##_needs_scalar(); \
 		fixed_point_t need_category##_needs_price_inverse_sum = fixed_point_t::_0(); \
 		need_category##_needs_acquired_quantity = need_category##_needs_desired_quantity = fixed_point_t::_0(); \
 		for (auto [good_definition, quantity] : type_never_null.get_##need_category##_needs()) { \
+			if (!market_instance.get_is_available(*good_definition)) { \
```
Hop311, line 310: Should we also check if the good `is_money`, similar to the `is_trading_good()` check?

Hop311 (dismissed):

src/openvic-simulation/pop/Pop.cpp
wvpm: Goods with `is_money` are valid goods for needs and can be bought. They do count towards need fulfillment just like any other.

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -5,6 +5,7 @@
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/economy/production/ArtisanalProducerFactoryPattern.hpp"
 #include "openvic-simulation/pop/PopType.hpp"
+#include "types/fixed_point/FixedPoint.hpp"
```
Hop311, line 8:
```suggestion
#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
```

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -253,29 +253,68 @@ void Pop::fill_needs_fulfilled_goods_with_false() {
...
 
-#define DEFINE_ADD_EXPENSE_FUNCTIONS(name)\
-	void Pop::add_##name(const fixed_point_t amount){\
-		if (OV_unlikely(amount == fixed_point_t::_0())) {\
-			Logger::warning("Adding ",#name, " of 0 to pop. Context", get_pop_context_text().str()); \
+#define DEFINE_ADD_EXPENSE_FUNCTIONS(name) \
+	void Pop::add_##name(const fixed_point_t amount){ \
+		if (OV_unlikely(amount == fixed_point_t::_0())) { \
+			if (size > 1024) { \
+				Logger::warning("Adding ",#name, " of 0 to pop. Context", get_pop_context_text().str()); \
```
Hop311, line 260:
```suggestion
				Logger::warning("Adding ", #name, " of 0 to pop. Context: ", get_pop_context_text().str()); \
```

src/openvic-simulation/economy/trading/GoodMarket.hpp
```diff
@@ -31,6 +30,7 @@ namespace OpenVic {
 		std::vector<GoodMarketSellOrder> market_sell_orders;
 
 	protected:
+		GoodDefinition const& PROPERTY_ACCESS(good_definition, protected);
```
Hop311, line 33: This can be private, `GoodInstance` only needs to access it via the `get_good_definition` getter function, and so it can be the first variable defined in this struct

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -253,29 +253,68 @@ void Pop::fill_needs_fulfilled_goods_with_false() {
...
-#define SET_ALL_INCOME_TO_ZERO(name)\
+#define SET_ALL_INCOME_TO_ZERO(name) \
 	name = fixed_point_t::_0();
 
+void Pop::allocate_for_needs(
+	GoodDefinition::good_definition_map_t const& scaled_needs,
+	fixed_point_t& price_inverse_sum,
+	fixed_point_t& cash_left_to_spend
+) {
+	GoodDefinition::good_definition_map_t money_to_spend_per_good_draft {};
```
Hop311, line 286: This could be a memory variable so its memory doesn't have to be re-allocated every time the function is called

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -135,6 +136,11 @@ namespace OpenVic {
 		std::stringstream get_pop_context_text() const;
 		void reserve_needs_fulfilled_goods();
 		void fill_needs_fulfilled_goods_with_false();
+		void allocate_for_needs(
+			GoodDefinition::good_definition_map_t const& scaled_needs,
+			fixed_point_t& price_inverse_sum,
+			fixed_point_t& cash_left_to_spend
+		);
```
Hop311, line 143:
```suggestion
		void allocate_for_needs(
			GoodDefinition::good_definition_map_t const& scaled_needs,
			fixed_point_t& price_inverse_sum,
			fixed_point_t& cash_left_to_spend
		) const;
```
This can be `const` as it only relies on `market_instance`, in fact it could even be a `static` function defined entirely within `Pop.cpp` if `MarketInstance const&` was passed to it as an argument (so not a `static` member function of `Pop`, but a standalone `static` function with the cpp file).

Hop311 (approved):

src/openvic-simulation/pop/Pop.cpp
wvpm: That's part of the TODO pool.
I couldn't add a comment earlier due it being part of a macro. I'll add it here.

src/openvic-simulation/pop/Pop.hpp
wvpm: No, `money_to_spend_per_good` is an instance variable and it needs to be mutable.

src/openvic-simulation/pop/Pop.hpp
wvpm: Removed. There was no need to add it. That was a mistake.

PR #319, opened 2025-02-16
Remove selected province, add player country+selected province mapmode args

Hop311:
Selected province should be handled at the GDExtension level, e.g. in multiplayer the host and each client will have a different selected province so it shouldn't be part of `MapInstance`

### DIFF
--- a/src/openvic-simulation/InstanceManager.cpp
+++ b/src/openvic-simulation/InstanceManager.cpp
@@ -225,9 +225,8 @@ bool InstanceManager::set_today_and_update(Date new_today) {
 	return true;
 }
 
-bool InstanceManager::expand_selected_province_building(size_t building_index) {
+bool InstanceManager::expand_province_building(ProvinceInstance* province, size_t building_index) {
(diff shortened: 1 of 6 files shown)


src/openvic-simulation/InstanceManager.cpp
```diff
@@ -225,9 +225,8 @@ bool InstanceManager::set_today_and_update(Date new_today) {
 	return true;
 }
 
-bool InstanceManager::expand_selected_province_building(size_t building_index) {
+bool InstanceManager::expand_province_building(ProvinceInstance* province, size_t building_index) {
```
wvpm, line 228: Consider using a reference as province may not be null.

wvpm (approved):

src/openvic-simulation/InstanceManager.cpp
Hop311: Yeah I was thinking about that but decided to minimise the number of changes rather than move the null check to the main repo, as this function is getting replaced by the game action system

PR #323, opened 2025-02-17
Fix strata order to: poor, middle, rich

Hop311:
- Needed for budget menu where strata UI elements are identified by index (0 = poor, 1 = middle, 2 = rich). Previous system indexed stratas based on their order of appearance in loaded pop type files, for vanilla: rich, middle, poor.
- This also adds a `default_strata`, following the base game's behaviour of defaulting to middle strata for invalid defines file strata values.
- This adds a vector of pop types to each strata so we can easily iterate over their pop types.
- This adds index variables to Strata and PopType, so they can be reliably identified by ints rather than strings.

### DIFF
--- a/src/openvic-simulation/pop/PopType.cpp
+++ b/src/openvic-simulation/pop/PopType.cpp
@@ -1,5 +1,8 @@
 #include "PopType.hpp"
 
+#include <array>
+#include <string_view>
+
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/country/CountryInstance.hpp"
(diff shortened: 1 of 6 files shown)


src/openvic-simulation/pop/PopType.cpp
```diff
@@ -197,9 +234,10 @@ bool PopManager::add_pop_type(
 		return false;
 	}
 
-	const bool ret = pop_types.add_item({
+	if (pop_types.add_item({
```
wvpm, line 237: Invert if so you return early.
It also saves indentation making the method easier to read.

wvpm (approved):

wvpm (approved):

PR #326, opened 2025-02-18
Performance: reuse maps & PopValuesFromProvince

wvpm:
Pops use several `GoodDefinition::good_definition_map_t` in their tick method. These maps can be cleared and reused by the next pop. `PopValuesFromProvince` similarly can be reused for multiple provinces as all data is overwritten in the `update_pop_values_from_province` method.

So to reuse resources when possible, `PopValuesFromProvince` is expanded to include several reusable maps.
The MapInstance now stores the instances of `PopValuesFromProvince` and creates threads explicitly to process provinces in parallel. This is done to limit the number of threads to the number of `PopValuesFromProvince`s.

`std::thread::hardware_concurrency()` is used to determine the amount of worker threads. Those threads are created in the map tick to process the provinces. The parent thread also processes provinces, so there is 1 more `PopValuesFromProvince`.

GoodMarket also has maps that can be reused. If this PR is approved, we can apply the same idea to MarketInstance.

### DIFF
--- a/src/openvic-simulation/map/MapInstance.cpp
+++ b/src/openvic-simulation/map/MapInstance.cpp
@@ -2,12 +2,14 @@
 
 #include "openvic-simulation/history/ProvinceHistory.hpp"
 #include "openvic-simulation/map/MapDefinition.hpp"
-#include "openvic-simulation/utility/CompilerFeatureTesting.hpp"
+#include "openvic-simulation/pop/PopValuesFromProvince.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
 
 using namespace OpenVic;
 
-MapInstance::MapInstance(MapDefinition const& new_map_definition) : map_definition { new_map_definition } {}
+MapInstance::MapInstance(MapDefinition const& new_map_definition)
+	: map_definition { new_map_definition }
+	{}
 
 ProvinceInstance& MapInstance::get_province_instance_from_definition(ProvinceDefinition const& province) {
 	return province_instances.get_items()[province.get_index() - 1];
@@ -43,7 +45,6 @@ bool MapInstance::setup(
 			if (province_instances.add_item({
 				market_instance,
 				modifier_effect_cache,
-				pop_defines,
 				province,
 				strata_keys,
 				pop_type_keys,
@@ -71,6 +72,24 @@ bool MapInstance::setup(
 			map_definition.get_province_definition_count(), ")!"
 		);
 		return false;
+	}
+
+	if (ret) {
+		const size_t max_worker_threads = std::thread::hardware_concurrency();
+		const size_t max_threads_including_parent = max_worker_threads + 1;
+		reusable_pop_values_collection.reserve(max_threads_including_parent);
+		threads.reserve(max_worker_threads);
+
+		for (size_t i = 0; i < max_worker_threads; i++) {
+			reusable_pop_values_collection.emplace_back(PopValuesFromProvince {
+				pop_defines, strata_keys
+			});
+			threads.emplace_back();
+		}
+
+		reusable_pop_values_collection.emplace_back(PopValuesFromProvince {
+			pop_defines, strata_keys
+		});
 	}
 
 	return ret;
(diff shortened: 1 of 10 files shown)


src/openvic-simulation/pop/Pop.hpp
```diff
@@ -104,17 +103,17 @@ namespace OpenVic {
...
+						return fixed_point_t::_1(); \
+					} \
+					return need_category##_needs_acquired_quantity / need_category##_needs_desired_quantity; \
 				} \
-				return need_category##_needs_acquired_quantity / need_category##_needs_desired_quantity; \
-			} \
 			private: \
-			GoodDefinition::good_definition_map_t need_category##_needs {}; \
-			ordered_map<GoodDefinition const*, bool> PROPERTY(need_category##_needs_fulfilled_goods);
+				GoodDefinition::good_definition_map_t PROPERTY(need_category##_needs); /* TODO pool? (if recalculating in UI is acceptable) */ \
```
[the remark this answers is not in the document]
Hop311: We can recalculate for the UI, it would be best if the code for recalculating them was part of the `Pop` struct so we don't have to update the UI separately whenever the logic changes

src/openvic-simulation/pop/PopValuesFromProvince.hpp
```diff
@@ -16,20 +18,23 @@ namespace OpenVic {
 		fixed_point_t PROPERTY(shared_everyday_needs_scalar);
 		fixed_point_t PROPERTY(shared_luxury_needs_scalar);
 	public:
-		void update_pop_strata_values_from_province(PopValuesFromProvince const& parent, Strata const& strata);
+		void update_pop_strata_values_from_province(PopsDefines const& defines, Strata const& strata, ProvinceInstance const& province);
 	};
 
 	struct PopValuesFromProvince {
 	private:
+		static constexpr size_t MAPS_FOR_POP = 4;
```
Hop311, line 26: This could be public, but it isn't a big deal as we can get it via `reusable_maps.size()`

src/openvic-simulation/pop/PopValuesFromProvince.hpp
```diff
@@ -16,20 +18,23 @@ namespace OpenVic {
...
 		PopsDefines const& PROPERTY(defines);
-		ProvinceInstance const* PROPERTY_RW(province, nullptr);
 		IndexedMap<Strata, PopStrataValuesFromProvince> PROPERTY(effects_per_strata);
 	public:
+	 	//public field as mutable references are required.
+		std::array<GoodDefinition::good_definition_map_t, MAPS_FOR_POP> reusable_maps {};
+
 		PopValuesFromProvince(
```
Hop311, line 33: Add a move constructor too, `PopValuesFromProvince(PopValuesFromProvince&&) = default;`, so we don't have to copy the maps in the array if/when we move this struct

src/openvic-simulation/map/MapInstance.cpp
```diff
@@ -156,27 +175,50 @@ void MapInstance::update_gamestate(const Date today, DefineManager const& define
 	state_manager.update_gamestate();
 }
 
+#define process_provinces_in_parallel(f) \
```
Hop311, line 178: This should be a function taking in a lambda function argument, rather than a macro. You could even make it a template to avoid unnecessary use of `std::function`, so (we don't need the `template` keyword if we use `<concept> auto` arguments):
```
void MapInstance::process_provinces_in_parallel(std::invocable<ProvinceInstance&> auto callback) {
	...
}
```
Just replace `f` with `callback(*it)`

src/openvic-simulation/pop/PopValuesFromProvince.hpp
wvpm: Why? Who else would need this?
Also the size is compiled into the type.

src/openvic-simulation/map/MapInstance.cpp
wvpm: Why should it be a function rather than a macro?
What are the pros and cons?

Pro for macro: no function jumping
Con for macro: hard to debug

Spartan322: Its not hard to debug with a concept, in fact macros are harder, breakpoints don't work at all for macros and with concepts if it violates the promise the concept mandates it tells you why it exactly failed.

Spartan322: I would also recommend avoiding using auto when we don't need a deduced type, it makes it harder to understand and review.

wvpm: Great point about debugging. I'll refactor it into a template.

Hop311 (approved):

PR #328, opened 2025-02-19
Thread safe pop tick and callbacks

wvpm:
Use atomic_fixed_point_t for += and -= opertions.
Use .at(key) instead of indexing to avoid try_emplace.


src/openvic-simulation/pop/Pop.hpp
```diff
@@ -97,21 +98,17 @@ namespace OpenVic {
 		fixed_point_map_t<Issue const*> PROPERTY(issue_distribution);
 		IndexedMap<CountryParty, fixed_point_t> PROPERTY(vote_distribution);
 
+		std::unique_ptr<std::mutex> cash_lock;
```
wvpm, line 101: Lock is required for the assertions and logs.
If they are removed, atomic_fixed_point_t should be used for expenses and cash as well.

Spartan322 (approved):

PR #329, opened 2025-02-19
Styling: remove trailing whitespace

wvpm:


Valerus9 (approved):
I reviewed all 6 lines of this commit and I believe that the 1 line change in ArtisanalProducer.cpp and 5 line change are indeed just trailing whitespace being removed and nothing more, nothing less.

PR #337, opened 2025-02-21
Round up income and expenses to handle inaccuracies

wvpm:
Round up income to epsilon when you have sold a good.
Round up expenses to epsilon when you have bought a good.
Round up money_to_spend when allocating to buy a good (so `money_to_spend >= max_next_price * max_quantity_to_buy`).

Round down quanity bought

### DIFF
--- a/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
+++ b/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
@@ -370,7 +370,10 @@ void ResourceGatheringOperation::pay_employees(
 
 		for (Pop* owner_pop_ptr : *owner_pops_cache_nullable) {
 			Pop& owner_pop = *owner_pop_ptr;
-			const fixed_point_t income_for_this_pop = revenue_left * (owner_share * owner_pop.get_size()) / total_owner_count_in_state_cache;
+			const fixed_point_t income_for_this_pop = std::max(
+				revenue_left * (owner_share * owner_pop.get_size()) / total_owner_count_in_state_cache,
+				fixed_point_t::epsilon() //revenue_left > 0 is already checked, so rounding up
+			);
 			owner_pop.add_rgo_owner_income(income_for_this_pop);
 			total_owner_income_cache += income_for_this_pop;
 		}
--- a/src/openvic-simulation/economy/trading/GoodMarket.cpp
+++ b/src/openvic-simulation/economy/trading/GoodMarket.cpp
@@ -2,6 +2,7 @@
 
 #include "openvic-simulation/economy/GoodDefinition.hpp"
 #include "openvic-simulation/misc/GameRulesManager.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
 #include "openvic-simulation/utility/CompilerFeatureTesting.hpp"
 
 using namespace OpenVic;
@@ -139,7 +140,13 @@ void GoodMarket::execute_orders() {
 			}
 
 			demand_sum += buy_up_to_order.get_max_quantity();
-			const fixed_point_t purchasing_power = purchasing_power_per_order[i] = buy_up_to_order.get_money_to_spend() / max_next_price;
+
+			if (buy_up_to_order.get_money_to_spend() <= fixed_point_t::_0()) {
+				purchasing_power_per_order[i] = quantity_bought_per_order[i] = fixed_point_t::_0();
+				continue;
+			}
+
+			fixed_point_t purchasing_power = purchasing_power_per_order[i] = buy_up_to_order.get_money_to_spend() / max_next_price;
 			if (purchasing_power >= buy_up_to_order.get_max_quantity()) {
 				quantity_bought_per_order[i] = buy_up_to_order.get_max_quantity();
 				max_quantity_to_buy_sum += buy_up_to_order.get_max_quantity();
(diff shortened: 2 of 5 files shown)


src/openvic-simulation/pop/Pop.cpp
```diff
@@ -357,6 +358,8 @@ void Pop::allocate_for_needs(
 			money_to_spend_per_good_draft[good_definition] = max_money_to_spend;
 			price_inverse_sum -= price_inverse;
 			i = -1; //Restart loop and skip maxed out needs. This is required to spread the remaining cash again.
+		} else if (cash_available_for_good / market_instance.get_min_next_price(*good_definition) < fixed_point_t::epsilon()) {
```
Hop311, line 361: So this is meant to represent the maximum amount of goods that could be bought, as any price over the min will result in an even smaller amount, right?

src/openvic-simulation/economy/trading/MarketInstance.cpp
```diff
@@ -21,6 +21,19 @@ fixed_point_t MarketInstance::get_max_next_price(GoodDefinition const& good_defi
...
+fixed_point_t MarketInstance::get_min_next_price(GoodDefinition const& good_definition) const {
+	return good_instance_manager.get_good_instance_from_definition(good_definition).get_min_next_price();
+}
+
+fixed_point_t MarketInstance::get_max_money_to_allocate_to_buy_quantity(GoodDefinition const& good_definition, const fixed_point_t quantity) const {
+	const fixed_point_t max_next_price = get_max_next_price(good_definition);
+	fixed_point_t costs = quantity * max_next_price;
+	if (costs / max_next_price < quantity) {
+		costs += fixed_point_t::epsilon(); //round up so money_to_spend >= max_next_price * max_quantity_to_buy
+	}
```
Hop311, line 33: I think this could be done more efficiently, without spending an entire division operation: `a * b` for fixed points has an error of the 16 bits shifted to the right, while division shifts it back up (with those bits now 0), so:
```
quantity * max_next_price = costs + err  (err < 2^-16)
costs / max_next_price = (quantity * max_next_price - err) / max_next_price = quantity - (err / max_next_price)
```
Therefore `costs / max_next_price < quantity` if `err > 0`, which it will be unless both `quantity` and `max_next_price` have at least 16 rightmost zeros between them. `fixed_point_t::epsilon()` is `parse_raw(1)`, so `2^-16` and therefore always greater than `err`.

We might as well always add `epsilon` to `costs` rather than doing the `costs < max_next_price < quantity` check which should almost always return true, except maybe for cases of `costs == 0` (similar to what you do in `GoodMarket.cpp`, in which case we must have `quantity == 0`.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -370,7 +370,10 @@ void ResourceGatheringOperation::pay_employees(
 
 		for (Pop* owner_pop_ptr : *owner_pops_cache_nullable) {
 			Pop& owner_pop = *owner_pop_ptr;
-			const fixed_point_t income_for_this_pop = revenue_left * (owner_share * owner_pop.get_size()) / total_owner_count_in_state_cache;
+			const fixed_point_t income_for_this_pop = std::max(
+				revenue_left * (owner_share * owner_pop.get_size()) / total_owner_count_in_state_cache,
```
Hop311, line 374: I'm not quite sure why the brackets are included here, is there any reason not to do:
```cpp
// fixed point * fixed point * int32_t / int32_t
revenue_left * owner_share * owner_pop.get_size() / total_owner_count_in_state_cache
```

src/openvic-simulation/pop/Pop.cpp
wvpm: Correct.

src/openvic-simulation/economy/trading/MarketInstance.cpp
wvpm: It will often be the case yeah, but not always. When `costs / max_next_price == quantity` we should not add epsilon.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: fixed point * int32_t is lossless
owner_pop.get_size() tends to be > 1 and thus increases the result and reduces rounding errors
revenue_left * owner_share is lossy

src/openvic-simulation/economy/trading/GoodMarket.cpp
```diff
@@ -248,11 +263,19 @@ void GoodMarket::execute_orders() {
 					buy_up_to_order.get_money_to_spend() / new_price
 				);
 
-				quantity_traded_yesterday += quantity_bought;
-				buy_up_to_order.get_after_trade()({
-					quantity_bought,
-					quantity_bought * new_price
-				});
+				const fixed_point_t money_spent = quantity_bought * new_price;
+				if (money_spent <= fixed_point_t::_0()) {
```
wvpm, line 267: random person suggested setting money_spent to epsilon in this case.
That's also a valid solution.

src/openvic-simulation/pop/Pop.cpp
wvpm: I made it a named variable.

src/openvic-simulation/economy/trading/MarketInstance.cpp
wvpm: I applied your suggestion.

src/openvic-simulation/economy/trading/GoodMarket.cpp
wvpm: Applied random persons suggestion.

Hop311 (approved):

PR #343, opened 2025-02-22
Fix artisan overspending

wvpm:
#337 rounded up expenses.
The artisanal optimal quantity was calculated as if we have endless precision and no rounding.
(Also optimal quantity wasn't used. Demand was used instead.)
The optimal quantity needs to take the minimum cost of epsilon into account. This is a flat cost and not proportional to the quantity bought.

Other changes:
- added force_tick_and_update to simulate several days in headless
- rounded up income for RGO workers (#337 should've included this...)
- minor spacing corrections in pop files

### DIFF
--- a/src/openvic-simulation/economy/production/ArtisanalProducer.cpp
+++ b/src/openvic-simulation/economy/production/ArtisanalProducer.cpp
@@ -1,4 +1,5 @@
 #include "ArtisanalProducer.hpp"
+#include <cstdint>
 
 #include "openvic-simulation/economy/production/ProductionType.hpp"
 #include "openvic-simulation/map/ProvinceInstance.hpp"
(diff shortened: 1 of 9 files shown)


src/openvic-simulation/economy/production/ArtisanalProducer.cpp
```diff
@@ -163,24 +163,33 @@ void ArtisanalProducer::artisan_tick(
...
 				max_quantity_to_buy_per_good[input_good_ptr] = max_quantity_to_buy;
 				pop.allocate_cash_for_artisanal_spending(money_to_spend);
 				pop_max_quantity_to_buy_per_good[input_good_ptr] += max_quantity_to_buy;
 				pop_money_to_spend_per_good[input_good_ptr] += money_to_spend;
+				debug_cash_left -= money_to_spend;
 			}
 		}
+
+		if (debug_cash_left < fixed_point_t::_0()) {
+			Logger::error("Artisan allocated more cash than the pop has. debug_cash_left: ", debug_cash_left);
```
wvpm, line 191: Is still triggered due to get_max_money_to_allocate_to_buy_quantity adding epsilon and optimal quantity calculation not taking that into account.

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -627,4 +627,4 @@ namespace OpenVic {
 	};
 
 	static_assert(sizeof(fixed_point_t) == fixed_point_t::SIZE, "fixed_point_t is not 8 bytes");
-}
+}
```
Hop311, line 630: In general I prefer having newlines at the end of files, there's no reason to remove this one especially given this PR doesn't change the file in any other way

Hop311 (approved):

PR #344, opened 2025-02-23
Reuse GoodMarket vectors

wvpm:
Similar to how provinces reuse PopValuesFromProvince, GoodMarket can reuse 2 vectors.


Hop311 (approved):
Great work!

PR #346, opened 2025-02-23
Make GameActionManager and its callback functions const

Hop311:


wvpm (approved):
If it compiles, ship it. LGTM

PR #348, opened 2025-02-24
Pool threads for performance

wvpm:


src/openvic-simulation/ThreadPool.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "openvic-simulation/types/Date.hpp"
+
+namespace OpenVic {
+	struct GoodInstance;
+	struct PopsDefines;
+	struct Strata;
+
+	class ThreadPool {	
+	private:
+		enum WORK_TYPE : uint8_t {
```
Spartan322, line 19: Enum names are types, treat them as types, also use enum class.

src/openvic-simulation/ThreadPool.hpp
```diff
@@ -0,0 +1,63 @@
...
+			NONE,
+			GOOD_EXECUTE_ORDERS,
+			PROVINCE_INITIALISE_FOR_NEW_GAME,
+			PROVINCE_TICK
+		};
+
+		void loop_until_cancelled(
+			WORK_TYPE& work_type,
```
Spartan322, line 27: Don't use references for enums, this is needlessly error prone, it just causes headaches.

src/openvic-simulation/ThreadPool.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include "openvic-simulation/pop/PopValuesFromProvince.hpp"
+#include "openvic-simulation/types/Date.hpp"
+
+namespace OpenVic {
+	struct GoodInstance;
+	struct PopsDefines;
+	struct Strata;
+
+	struct ThreadPool {
```
Hop311, line 17:
```suggestion
	struct ThreadPool {
```
Stray tab at the end of the line, there are a couple more + some stray spaces across this file and `ThreadPool.cpp`, could you search and remove them

src/openvic-simulation/ThreadPool.hpp
```diff
@@ -0,0 +1,63 @@
+#pragma once
```
Hop311, line 1: `ThreadPool..hpp` and `ThreadPool.cpp` could go in the `utility` or `misc` folder

src/openvic-simulation/InstanceManager.cpp
```diff
@@ -161,6 +168,15 @@ bool InstanceManager::setup() {
 		definition_manager.get_define_manager().get_economy_defines()
 	);
 
+	if (ret) {
```
Hop311, line 171: I'm not sure that we should gate thread pool initialisation behind `ret` being true here, the reason `ret` is false isn't necessarily game-ending and we could recover from it but not initialising the thread pool will definitely prevent the game from being playable.
We should think carefully about what errors are game-ending and in those cases gracefully exit, but beyond that we should press ahead and try salvage the game.

src/openvic-simulation/ThreadPool.hpp
```diff
@@ -0,0 +1,63 @@
...
+		
+	public:
+		ThreadPool() = default;
+		~ThreadPool();
+
+		void initialise(
+			PopsDefines const& pop_defines,
+			std::vector<Strata> const& strata_keys,
+			std::vector<GoodInstance>& goods,
+			std::vector<ProvinceInstance>& provinces
```
Hop311, line 56: It would be nicer to pass `std::span`s instead of mutable vector references here, just to make absolutely clear we're not modifying the vector/memory arrangement itself, just the individual items inside the vectors

src/openvic-simulation/ThreadPool.hpp
```diff
@@ -0,0 +1,63 @@
...
+		);
+
+		void process_good_execute_orders();
+		void process_province_ticks(const Date today);
+		void process_province_initialise_for_new_game(const Date today);
```
Hop311, line 61: I believe these `Date` arguments are leftovers from the previous way these functions were structured, I think it would be better now if `InstanceManager` updated `ThreadPool`'s current date at the beginning of each tick and all work orders followed that rather than needing to pass the date through these `process_*` functions

src/openvic-simulation/ThreadPool.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+void ThreadPool::process_work(const Date new_current_date, const work_t work_type) {
+	if (is_cancellation_requested) {
+		return;
+	}
+	{
+		std::unique_lock<std::mutex> lock {mutex};
+		if (is_cancellation_requested) {
+			return;
+		}
```
Hop311, line 77: Why are there two `is_cancellation_requested` checks? Is it to prevent a mutex dead-locking issue?

src/openvic-simulation/ThreadPool.cpp
```diff
@@ -0,0 +1,176 @@
...
+				&strata_keys,
+				goods_begin, goods_end,
+				provinces_begin, provinces_end
+			]() -> void {
+				loop_until_cancelled(
+					work_for_thread,
+					pop_defines,
+					strata_keys,
+					std::span<GoodInstance>{goods_begin, goods_end},
+					std::span<ProvinceInstance>{provinces_begin, provinces_end}
```
Hop311, line 156:
```suggestion
					std::span<GoodInstance> { goods_begin, goods_end },
					std::span<ProvinceInstance> { provinces_begin, provinces_end }
```
Could you also add spaces to the various mutex lock constructors

Hop311 (commented):
Great work, the performance gains you mentioned are very impressive! We should think about what else could be parallelised by this system, it should be fairly easy to add more work types

src/openvic-simulation/ThreadPool.cpp
wvpm: I removed it as is_cancellation_requested is only true when the ThreadPool is destructed and the method can't be called anymore then.

Hop311 (approved):

PR #349, opened 2025-02-25
Move std::bind use in GFXObject to lambda binding

Spartan322:
This follows the common rule that std::bind, being way too complex, difficult to optimize, and prone to hurting template instantion, and so should never be used.


Hop311 (approved):
Great work! We also use `std::bind_front` in `CountryInstance.cpp`, `Dataloader.cpp`, `MapDefinition.cpp` and `Culture.cpp`, and in the main repo we use `std::bind` + `std::bind_front` in `GameSingleton.cpp`, I assume we should switch all of those over to using lambda functions too?

Spartan322:
Doesn't matter for bind_front, bind however should be eliminated yes.

PR #350, opened 2025-02-25
Fix occurrences of undefined behavior

Spartan322:
Fix string_to_uint64 and string_to_int64 to act like std::from_chars
Add constexpr StringUtils::from_chars

### DIFF
--- a/src/openvic-simulation/types/fixed_point/FixedPoint.hpp
+++ b/src/openvic-simulation/types/fixed_point/FixedPoint.hpp
@@ -1,11 +1,13 @@
 #pragma once
 
+#include <charconv>
 #include <concepts>
 #include <cstdint>
 #include <cstdlib>
 #include <limits>
 #include <sstream>
 #include <string_view>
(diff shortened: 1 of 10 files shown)


src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -610,7 +612,11 @@ namespace OpenVic {
 
 	private:
 		static constexpr fixed_point_t parse_integer(char const* str, char const* const end, bool* successful) {
-			int64_t parsed_value = StringUtils::string_to_int64(str, end, successful, 10);
+			int64_t parsed_value;
```
Hop311, line 615: Should we default initialise this to 0, just to be sure it doesn't contain random data if parsing fails? (Ofc we should ignore the value in the case of failure, but safer to make it 0 so everything is consistent if it is used)

Hop311 (approved):

Hop311 (approved):

PR #352, opened 2025-02-25
Add country tax properties

Hop311:


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1007,6 +1011,14 @@ void CountryInstance::_update_budget(ModifierEffectCache const& modifier_effect_
 
 	// TODO - make sure we properly update everything dependent on these sliders' values,
 	// as they might change if their sliders' bounds shrink past their previous values.
+
+	const fixed_point_t tax_efficiency = country_defines.get_base_country_tax_efficiency() + get_modifier_effect_value(
```
wvpm, line 1015: Formatting...
When parts of an equation are equal, format them similarly.
So something like 
```cpp
	const fixed_point_t tax_efficiency = country_defines.get_base_country_tax_efficiency()
		+ get_modifier_effect_value(*modifier_effect_cache.get_tax_efficiency())
		+ get_modifier_effect_value(*modifier_effect_cache.get_tax_eff()) / 100;
```

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1007,6 +1011,14 @@ void CountryInstance::_update_budget(ModifierEffectCache const& modifier_effect_
...
+		*modifier_effect_cache.get_tax_efficiency()
+	) + get_modifier_effect_value(*modifier_effect_cache.get_tax_eff()) / 100;
+
+	for (auto const& [strata, tax_rate_slider_value] : tax_rate_slider_value_by_strata) {
+		effective_tax_rate_by_strata[strata] = tax_rate_slider_value.get_value() * tax_efficiency;
```
wvpm, line 1020: We're multiplying fixed point (slider value) by fixed point (tax eff) by fixed point (income)
This means we'll effectively round down taxes multiple times. The pops should be glad.

wvpm (dismissed):

src/openvic-simulation/country/CountryInstance.cpp
Hop311: Base 10 fixed points should fix most of these issues, as we'll eliminate most parsing/accuracy error in the initial values (mostly involving 1/100) which then means far fewer rightmost bits and therefore lower error from multiplication.

wvpm (approved):

PR #354, opened 2025-02-26
Record pop employment for RGOs and calculate unemployment fraction.

wvpm:


src/openvic-simulation/pop/PopType.hpp
```diff
@@ -76,7 +76,7 @@ namespace OpenVic {
 		const bool PROPERTY(can_invest);
 		const bool PROPERTY(factory);
 		const bool PROPERTY(can_work_factory);
-		const bool PROPERTY(unemployment);
+		const bool PROPERTY_CUSTOM_NAME(unemployment, can_be_unemployment);
```
Hop311, line 79: `can_be_unemployment` doesn't sound right, maybe `can_be_unemployed`?

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -99,7 +101,13 @@ namespace OpenVic {
 		fixed_point_map_t<Issue const*> PROPERTY(issue_distribution);
 		IndexedMap<CountryParty, fixed_point_t> PROPERTY(vote_distribution);
 
-		fixed_point_t PROPERTY(unemployment);
+		pop_size_t employed = 0;
+		constexpr fixed_point_t get_unemployment_fraction() {
+			if (!type->can_be_unemployment()) {
+				return fixed_point_t::_0();
+			}
+			return fixed_point_t { size - employed } / size;
```
Hop311, line 109:
```suggestion
			return fixed_point_t::parse(size - employed) / size;
```
Safer to use `fixed_point_t::parse` to avoid any risk of treating the input as a raw value (so not shifting it up)

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -353,6 +353,7 @@ void Pop::pop_tick_without_cleanup(PopValuesFromProvince& shared_values) {
 	DO_FOR_ALL_TYPES_OF_POP_EXPENSES(SET_TO_ZERO)
 	#undef SET_TO_ZERO
 	income = expenses = fixed_point_t::_0();
+	employed = 0;
```
Hop311, line 356: This shouldn't be necessary, it can be removed

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -543,6 +544,18 @@ void Pop::report_artisanal_produce(const fixed_point_t quantity) {
 	artisanal_produce_left_to_sell = quantity;
 }
 
+void Pop::hire(pop_size_t count) {
+	if (OV_unlikely(count)) {
```
Hop311, line 548:
```suggestion
	if (OV_unlikely(count == 0)) {
```
Right now this will fail if `count` is non-zero?

src/openvic-simulation/pop/Pop.hpp
```diff
@@ -99,7 +101,13 @@ namespace OpenVic {
 		fixed_point_map_t<Issue const*> PROPERTY(issue_distribution);
 		IndexedMap<CountryParty, fixed_point_t> PROPERTY(vote_distribution);
 
-		fixed_point_t PROPERTY(unemployment);
+		pop_size_t employed = 0;
+		constexpr fixed_point_t get_unemployment_fraction() {
```
Hop311, line 105:
```suggestion
	public:
		constexpr fixed_point_t get_unemployment_fraction() const {
```
This needs to be `public` and `const` so it can be accessed for the Population Menu.
We'll also need a function giving the absolute number of unemployed people in the pop for the notification on the Production section of the topbar, and for calculating the number of unemployed clerks and craftsmen for the Production Menu

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -208,8 +208,13 @@ void ResourceGatheringOperation::hire() {
 		for (Job const& job : jobs) {
 			if (job.get_pop_type() == &pop_type) {
 				const pop_size_t pop_size_to_hire = static_cast<pop_size_t>((proportion_to_hire * pop.get_size()).floor());
+				if (pop_size_to_hire <= 0) {
+					continue;
+				}
+
 				employee_count_per_type_cache[pop_type] += pop_size_to_hire;
 				employees.emplace_back(pop, pop_size_to_hire);
+				pop.hire(pop_size_to_hire);
```
Hop311, line 217: This RGO code isn't being called? And so `Pop::hire` is never being called either. Take a look in game and you'll see all RGOs with 0 workers

wvpm: https://github.com/OpenVicProject/OpenVic-Simulation/pull/357

Hop311 (approved):

PR #360, opened 2025-02-28
Implement taxation

wvpm:


src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -136,6 +136,10 @@ namespace OpenVic {
 			return province_definition;
 		}
 
+		constexpr CountryInstance const* get_country_to_report_economy() const {
+			return controller;
+		}
```
Hop311, line 141: Are we sure all economic stuff goes to the controller? E.g. when testing gold income, occupied gold RGOs still sent money to their owner rather than controller.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -240,8 +240,19 @@ void Pop::fill_needs_fulfilled_goods_with_false() {
...
 }
 
+void Pop::pay_income_tax(fixed_point_t& income) {
+	CountryInstance* const tax_collector_nullable = location->get_country_to_report_economy();
+	if (tax_collector_nullable == nullptr) {
+		return;
+	}
+	const fixed_point_t effective_tax_rate = tax_collector_nullable->get_effective_tax_rate_by_strata()[type->get_strata()];
+	const fixed_point_t tax = effective_tax_rate * income;
+	tax_collector_nullable->report_pop_income_tax(*type, tax, income);
```
Hop311, line 250: I think the args here are the wrong way around, `gross_income` comes before `paid_as_tax`

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1572,19 +1578,25 @@ void CountryInstance::good_data_t::clear_daily_recorded_data() {
 	production_per_production_type.clear();
 }
 
+void CountryInstance::report_pop_income_tax(PopType const& pop_type, const fixed_point_t gross_income, const fixed_point_t paid_as_tax) {
+	taxable_income_by_pop_type[pop_type] += gross_income;
```
Hop311, line 1582: Are you sure this doesn't need to be mutex protected? Is pop tax payment not parallelised at all?

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1485,6 +1486,10 @@ void CountryInstance::country_reset_before_tick() {
 	for (auto pair : goods_data) {
 		pair.second.clear_daily_recorded_data();
 	}
+
+	for (auto pair : taxable_income_by_pop_type) {
+		pair.second = fixed_point_t::_0();
+	}
```
Hop311, line 1492:
```suggestion
	taxable_income_by_pop_type.clear();
```
You could also do `taxable_income_by_pop_type.fill(fixed_point_t::_0());`, `clear` default-initialises every value while `fill` copies the arg into every value

src/openvic-simulation/country/CountryInstance.cpp
wvpm: It should be protected. I changed it.

src/openvic-simulation/pop/Pop.cpp
wvpm: Oops, you're right.

src/openvic-simulation/map/ProvinceInstance.hpp
wvpm: get_country_to_report_economy already existed, this is merely a const overload for it.

Hop311 (approved):

PR #362, opened 2025-02-28
Add game rule for economic country

wvpm:
In Victoria 2, the owner collects taxes and gold income even when provinces are occupied.
It makes more sense that the controller gets that. Alternatively occupied territories could be considered in disarray and neither the owner nor the controller are able to collect taxes.

All of this is now supported via game rules.

### DIFF
--- a/src/openvic-simulation/map/ProvinceInstance.cpp
+++ b/src/openvic-simulation/map/ProvinceInstance.cpp
@@ -10,6 +10,7 @@
 #include "openvic-simulation/map/Region.hpp"
 #include "openvic-simulation/map/TerrainType.hpp"
 #include "openvic-simulation/military/UnitInstanceGroup.hpp"
+#include "openvic-simulation/misc/GameRulesManager.hpp"
 #include "openvic-simulation/modifier/StaticModifierCache.hpp"
 #include "openvic-simulation/politics/Ideology.hpp"
 #include "openvic-simulation/pop/PopValuesFromProvince.hpp"
@@ -19,6 +20,7 @@ using namespace OpenVic;
 
 ProvinceInstance::ProvinceInstance(
 	MarketInstance& new_market_instance,
+	GameRulesManager const& new_game_rules_manager,
 	ModifierEffectCache const& new_modifier_effect_cache,
 	ProvinceDefinition const& new_province_definition,
 	decltype(population_by_strata)::keys_type const& strata_keys,
@@ -28,6 +30,7 @@ ProvinceInstance::ProvinceInstance(
 	HasIndex { new_province_definition.get_index() },
 	FlagStrings { "province" },
 	province_definition { new_province_definition },
+	game_rules_manager { new_game_rules_manager },
 	modifier_effect_cache { new_modifier_effect_cache },
 	terrain_type { new_province_definition.get_default_terrain_type() },
 	rgo { new_market_instance, pop_type_keys },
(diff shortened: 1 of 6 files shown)


src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -84,6 +87,14 @@ bool ProvinceInstance::set_owner(CountryInstance* new_owner) {
 		for (Pop& pop : pops) {
 			pop.update_location_based_attributes();
 		}
+
+		if (game_rules_manager.get_country_to_report_economy() == country_to_report_economy_t::Owner) {
+			country_to_report_economy = new_owner;
+		} else if (game_rules_manager.get_country_to_report_economy() == country_to_report_economy_t::NeitherWhenOccupied) {
+			country_to_report_economy = new_owner == controller
```
Hop311, line 94: An constexpr `is_occupied` helper function would be nice here and in `set_controller`

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -102,6 +113,14 @@ bool ProvinceInstance::set_controller(CountryInstance* new_controller) {
 		if (controller != nullptr) {
 			ret &= controller->add_controlled_province(*this);
 		}
+
+		if (game_rules_manager.get_country_to_report_economy() == country_to_report_economy_t::Controller) {
+			country_to_report_economy = new_controller;
+		} else if (game_rules_manager.get_country_to_report_economy() == country_to_report_economy_t::NeitherWhenOccupied) {
+			country_to_report_economy = new_controller == owner
```
Hop311, line 120: I've just realised that we probably want to make it so `set_controller(nullptr);` on a province with non-null owner actually changes ownership to the owner. It would be nice to incoporate that just above here, setting the controller to the owner if `new_controller`/`controller` after line 111 is nullptr, and then continuing with the `add_controlled_province` code

src/openvic-simulation/map/ProvinceInstance.hpp
```diff
@@ -64,6 +65,7 @@ namespace OpenVic {
 
 	private:
 		ProvinceDefinition const& PROPERTY(province_definition);
+		GameRulesManager const& PROPERTY(game_rules_manager);
```
Hop311, line 68: Not a big deal for now, but soon we should have a discussion about when to store references to managers in structs that use them vs when to get them as function arguments and/or via other managers. I'm thinking in particular because the condition scripts PR will make most update and tick functions take a `InstanceManager [const]&` argument for use with scripts, from which we can get basically everything in the game.

Hop311 (dismissed):

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -92,6 +103,10 @@ bool ProvinceInstance::set_owner(CountryInstance* new_owner) {
 bool ProvinceInstance::set_controller(CountryInstance* new_controller) {
 	bool ret = true;
 
+	if (new_controller == nullptr) {
+		new_controller = owner;
+	}
```
Hop311, line 108: It just ocurred to me that if we ever do actually want to set the controller to null, we'd have to change the owner first - if we do decide we may need to actually set the owner to null ever (which is technically possible in the base game, here Coventry is controlled by either null or something like `noCountry`), we'd need something like an extra bool arg to allow null without changing to owner, but IMO it would be simplest to ensure we never set a province's controller to nullptr when we actually mean owner. With that example in the base game, ENG regains control of the province on the next tick, so we should implement something similar and avoid changing controller on update_gamestate calls
image

Hop311 (approved):

PR #363, opened 2025-03-02
Government salaries and social income

wvpm:
Implements administration, education & military spending as well as pensions and unemployment subsidies.

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.cpp
+++ b/src/openvic-simulation/country/CountryInstance.cpp
@@ -2,6 +2,7 @@
 
 #include <cstdint>
 
+#include "openvic-simulation/country/SharedCountryValues.hpp"
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/defines/Define.hpp"
 #include "openvic-simulation/DefinitionManager.hpp"
@@ -14,9 +15,12 @@
 #include "openvic-simulation/modifier/ModifierEffectCache.hpp"
 #include "openvic-simulation/modifier/StaticModifierCache.hpp"
 #include "openvic-simulation/politics/Ideology.hpp"
+#include "openvic-simulation/pop/Pop.hpp"
+#include "openvic-simulation/pop/PopType.hpp"
 #include "openvic-simulation/research/Invention.hpp"
 #include "openvic-simulation/research/Technology.hpp"
 #include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/types/PopSize.hpp"
 #include "openvic-simulation/types/SliderValue.hpp"
 
 using namespace OpenVic;
@@ -41,6 +45,7 @@ CountryInstance::CountryInstance(
 	decltype(ship_type_unlock_levels)::keys_type const& ship_type_unlock_levels_keys,
 	decltype(tax_rate_slider_value_by_strata)::keys_type const& strata_keys,
 	GameRulesManager const& new_game_rules_manager,
+	SharedCountryValues const& new_shared_country_values,
 	GoodInstanceManager& new_good_instance_manager,
 	CountryDefines const& new_country_defines,
 	EconomyDefines const& new_economy_defines
@@ -49,6 +54,7 @@ CountryInstance::CountryInstance(
 	/* Main attributes */
 	country_definition { new_country_definition },
 	game_rules_manager { new_game_rules_manager },
+	shared_country_values { new_shared_country_values },
 	country_defines { new_country_defines },
 	colour { ERROR_COLOUR },
 
@@ -989,8 +995,9 @@ static inline constexpr fixed_point_t nonzero_or_one(fixed_point_t const& value)
 	return value == fixed_point_t::_0() ? fixed_point_t::_1() : value;
 }
 
-void CountryInstance::_update_budget(DefineManager const& define_manager, ModifierEffectCache const& modifier_effect_cache) {
-	CountryDefines const& country_defines = define_manager.get_country_defines();
+void CountryInstance::_update_budget() {
+	CountryDefines const& country_defines = shared_country_values.get_country_defines();
+	ModifierEffectCache const& modifier_effect_cache = shared_country_values.get_modifier_effect_cache();
 
 	const fixed_point_t min_tax = get_modifier_effect_value(*modifier_effect_cache.get_min_tax());
 	const fixed_point_t max_tax = nonzero_or_one(get_modifier_effect_value(*modifier_effect_cache.get_max_tax()));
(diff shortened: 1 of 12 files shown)


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1025,6 +1029,70 @@ void CountryInstance::_update_budget(DefineManager const& define_manager, Modifi
...
+	if ((pop_type.get_##need_category##_needs_income_types() & income_type) == income_type) { \
+		for (auto const& [good_definition_ptr, quantity] : pop_type.get_##need_category##_needs()) { \
+			GoodInstance const& good_instance = get_good_instance(*good_definition_ptr); \
+			total_need_costs += good_instance.get_price() * quantity; \
+		} \
+	}
+
+	DO_FOR_ALL_NEED_CATEGORIES(ADD_NEED_COSTS)
+#undef ADD_NEED_COSTS
+	const fixed_point_t low_administrative_efficiency_penalty = std::max(
```
wvpm, line 1049: See https://discord.com/channels/1063392556160909312/1063416834650554398/1344230596821651539

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1025,6 +1029,70 @@ void CountryInstance::_update_budget(DefineManager const& define_manager, Modifi
...
+}
+fixed_point_t CountryInstance::calculate_social_income_form_base(
+	PopType const& pop_type,
+	ModifierEffectCache const& modifier_effect_cache,
+	PopsDefines const& pop_defines
+) const {
+	fixed_point_t life_need_costs = fixed_point_t::_0();
+	for (auto const& [good_definition_ptr, quantity] : pop_type.get_life_needs()) {
+		GoodInstance const& good_instance = get_good_instance(*good_definition_ptr);
+		life_need_costs += good_instance.get_price() * quantity;
```
wvpm, line 1087: Caching per PopType makes sense but is not possible as PopType is a const definition.

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1017,13 +1023,80 @@ void CountryInstance::_update_budget(DefineManager const& define_manager, Modifi
...
+		= projected_unemployment_subsidies_spending_unscaled_by_slider
+		= fixed_point_t::_0();
+
+	for (auto const& [pop_type, size] : pop_type_distribution) {
+		projected_administration_spending_unscaled_by_slider += size * calculate_administration_salary_base(pop_type, shared_country_values);
+		projected_education_spending_unscaled_by_slider += size * calculate_education_salary_base(pop_type, shared_country_values);
+		projected_military_spending_unscaled_by_slider += size * calculate_military_salary_base(pop_type, shared_country_values);
+		projected_pensions_spending_unscaled_by_slider += size * calculate_pensions_base(pop_type, shared_country_values);
+		projected_unemployment_subsidies_spending_unscaled_by_slider += pop_type_unemployed_count[pop_type]
+			* calculate_unemployment_subsidies_base(pop_type, shared_country_values);
```
Hop311, line 1070: As mentioned in previous comments, the functions called here involve many repeated lookups/calculations which could be avoided with caching

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1552,15 +1625,73 @@ void CountryInstance::update_gamestate(InstanceManager& instance_manager) {
...
+					available_funds -= projected_social_spending;
+					actual_social_spending = projected_social_spending;
+
+					actual_military_spending = std::min(available_funds, projected_military_spending);
+				}
+			}
+		}
+	}
+
+	for (auto const& [good_instance, data] : goods_data) {
```
Hop311, line 1687:
```suggestion
	for (auto& data : goods_data.get_values()) {
```
You can iterate over an `IndexedMap`'s values directly

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1691,6 +1822,68 @@ void CountryInstance::report_output(ProductionType const& production_type, const
 	good_data.production_per_production_type[&production_type] += quantity;
 }
 
+void CountryInstance::request_salaries_and_welfare(Pop& pop, SharedCountryValues const& shared_country_values) const {
```
Hop311, line 1825: This function could also benefit from caching rather than always looking up/calculating new values in every function call

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -2009,14 +2215,21 @@ void CountryInstanceManager::update_gamestate(InstanceManager& instance_manager)
...
+	//TODO update and pay government salaries and social spending
 	for (CountryInstance& country : country_instances.get_items()) {
-		country.country_tick(instance_manager);
+		country.country_tick_after_map(instance_manager);
 	}
+	
+	shared_country_values.update_costs(instance_manager.get_good_instance_manager());
 }
+
+#undef DO_FOR_ALL_NEED_CATEGORIES
```
Hop311, line 2235: This and the `PopNeedsMacro.hpp` include at the top of the file can be removed

src/openvic-simulation/country/SharedCountryValues.hpp
```diff
@@ -0,0 +1,52 @@
...
+		DO_FOR_ALL_NEED_CATEGORIES(NEED_COST_FIELD)
+		#undef NEED_COST_FIELD
+
+		fixed_point_t PROPERTY(administration_salary_base);
+		fixed_point_t PROPERTY(education_salary_base);
+		fixed_point_t PROPERTY(military_salary_base);
+
+	public:
+		void update_costs(PopType const& pop_type, PopsDefines const& pop_defines, GoodInstanceManager const& good_instance_manager);
+		constexpr fixed_point_t get_social_income_form_base() const {
```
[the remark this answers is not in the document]
wvpm: It's a form (variant) of social income.
And base as in unscaled by slider, pop size or modifiers.

src/openvic-simulation/pop/PopType.cpp
```diff
@@ -13,6 +13,7 @@
 #include "openvic-simulation/politics/Rebel.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
 #include "openvic-simulation/utility/TslHelper.hpp"
+#include "dataloader/NodeTools.hpp"
```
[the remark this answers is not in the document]
wvpm: Idk how it got there. clangd is broken for me due to out of memory issues.
I didn't expect it to add includes. I typically check for this stuff just before merging.
I'll clean the imports.

src/openvic-simulation/country/SharedCountryValues.cpp
```diff
@@ -0,0 +1,53 @@
...
+		base_##need_category##_need_costs *= pop_defines.get_base_goods_demand(); \
+		if ((pop_type.get_##need_category##_needs_income_types() & ADMINISTRATION) == ADMINISTRATION) { \
+			administration_salary_base += base_##need_category##_need_costs; \
+		} \
+		if ((pop_type.get_##need_category##_needs_income_types() & EDUCATION) == EDUCATION) { \
+			education_salary_base += base_##need_category##_need_costs; \
+		} \
+		if ((pop_type.get_##need_category##_needs_income_types() & MILITARY) == MILITARY) { \
+			military_salary_base += base_##need_category##_need_costs; \
+		}
```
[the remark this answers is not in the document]
wvpm:
```cpp
	inline constexpr bool share_income_type(PopType::income_type_t lhs, PopType::income_type_t rhs) {
		return (lhs & rhs) != PopType::income_type_t::NO_INCOME_TYPE;
	}
```
Is not the same as `(lhs & rhs) == rhs`.
There is a potential for differences when rhs has multiple bits 1.
I could change `share_income_type`. I'm not sure if the current implementation was an oversight or desired behaviour.
That's why I wrote it like this. Also this is an instantly recognisable unambiguous simple pattern.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -131,21 +134,47 @@ namespace OpenVic {
...
 		moveable_atomic_fixed_point_t PROPERTY(cash_stockpile);
 		std::unique_ptr<std::mutex> taxable_income_mutex;
 		fixed_point_t PROPERTY(gold_income);
 		IndexedMap<PopType, fixed_point_t> PROPERTY(taxable_income_by_pop_type);
 		IndexedMap<Strata, fixed_point_t> PROPERTY(effective_tax_rate_by_strata);
 		IndexedMap<Strata, SliderValue> PROPERTY(tax_rate_slider_value_by_strata);
+
+		fixed_point_t PROPERTY(administrative_efficiency);
+		constexpr fixed_point_t get_corruption_cost_multiplier() const {
+			return 2 - administrative_efficiency;
```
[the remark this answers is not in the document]
wvpm: `administrative_efficiency` should be bound [0,1].
`get_corruption_cost_multiplier` shouldn't check for that.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -523,7 +552,50 @@ namespace OpenVic {
...
+			return shared_country_values.get_shared_pop_type_values()[pop_type].get_education_salary_base()
+				* get_corruption_cost_multiplier();
+		}
+		constexpr fixed_point_t calculate_military_salary_base(
+			PopType const& pop_type,
+			SharedCountryValues const& shared_country_values
+		) const {
+			return shared_country_values.get_shared_pop_type_values()[pop_type].get_military_salary_base()
+				* get_corruption_cost_multiplier();
+		}
```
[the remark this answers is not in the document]
wvpm: Good point. Your suggestion reduces look ups yeah.
I wrote it like this as I first wrote the 'interface' and then the implementation.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -588,8 +662,17 @@ namespace OpenVic {
...
+			CountryDefinitionManager const& new_country_definition_manager,
+			ModifierEffectCache const& new_modifier_effect_cache,
+			CountryDefines const& new_country_defines,
+			PopsDefines const& new_pop_defines,
+			std::vector<PopType> const& pop_type_keys
+		);
 
+		constexpr SharedCountryValues const& get_shared_country_values() const {
+			return shared_country_values;
+		}
```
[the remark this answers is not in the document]
wvpm: I wasn't sure `PROPERTY` returns a `const&` when I store the struct directly.
As clangd is broken for me, I couldn't verify.
From your comment I can conclude `PROPERTY` would work here and I'll use that.

src/openvic-simulation/country/CountryInstance.cpp
wvpm: Cool, let's do that then :)

src/openvic-simulation/country/CountryInstance.hpp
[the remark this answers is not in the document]
wvpm: *sigh*
I tested it and it is capped at 100% but has no lower limit. You can go negative. The results are hilarious.
Pensions take money from pops and unemployment benefits become an unemployment tax.
Tariffs are also flipped.

I'll add `administrative_efficiency = std::min(..., 1)`.
I'll also add a lower limit of 0 as game rule.

src/openvic-simulation/country/SharedCountryValues.cpp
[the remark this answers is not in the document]
Hop311: I understand that it's not always the same, but I mean in these specific cases it can be used as the right hand sides (`ADMINISTRATION`, `EDUCATION` and `MILITARY`) are all a single bit

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -1017,13 +1024,89 @@ void CountryInstance::_update_budget(DefineManager const& define_manager, Modifi
...
+		= projected_unemployment_subsidies_spending_unscaled_by_slider
+		= fixed_point_t::_0();
+
+	const fixed_point_t corruption_cost_multiplier = get_corruption_cost_multiplier();
+	for (auto const& [pop_type, size] : pop_type_distribution) {
+		SharedPopTypeValues const& pop_type_values = shared_country_values.get_shared_pop_type_values()[pop_type];
+		projected_administration_spending_unscaled_by_slider += size * calculate_administration_salary_base(pop_type_values, corruption_cost_multiplier);
+		projected_education_spending_unscaled_by_slider += size * calculate_education_salary_base(pop_type_values, corruption_cost_multiplier);
+		projected_military_spending_unscaled_by_slider += size * calculate_military_salary_base(pop_type_values, corruption_cost_multiplier);
+		projected_pensions_spending_unscaled_by_slider += size * calculate_pensions_base(modifier_effect_cache, pop_type_values);
```
Hop311, line 1078: A lot (if not all) of these involve multiplication by `corruption_cost_multiplier`, could we calculate `size * corruption_cost_multiplier` at the beginning of the loop to avoid repeating calculations? This may be premature/over-optimisation, but this code will run a lot so any bit of speedup here should be worth it

Hop311 (approved):
Looks good to me, I left a comment about further avoidance of repeated calculations, and there may be even more places this point applies.
On another note, I wonder if there's a cleaner way of providing access to `SharedCountryValues` which could avoid storing a copy of the same address in every `CountryInstance` instance.

src/openvic-simulation/country/CountryInstance.cpp
wvpm: Only the administration, education & military salaries are multiplied by `corruption_cost_multiplier`.
I think this approach makes more sense as the `calculate_military_salary_base` & `calculate_pensions_base` return a similar value that only has to be multiplied by size. Otherwise you have to be careful to multiply by size or size x corruption.

PR #368, opened 2025-03-10
Log valid keys in expect_dictionary_keys

wvpm:
Reworks the callback so the key map is passed along.
This allows us to log valid keys to aid modders instead of merely specifying their key is invalid.

### DIFF
--- a/src/openvic-simulation/dataloader/NodeTools.hpp
+++ b/src/openvic-simulation/dataloader/NodeTools.hpp
@@ -91,6 +91,20 @@ using namespace std::string_view_literals;
 		template<typename Fn>
 		concept KeyValueCallback = Callback<Fn, std::string_view, ast::NodeCPtr>;
 
+		template<typename Fn, typename Map>
+		concept MapKeyValueCallback = Callback<Fn, Map const&, std::string_view, ast::NodeCPtr>;
+		
+		template<typename Map>
+		constexpr static MapKeyValueCallback<Map> auto ignore_map(KeyValueCallback auto&& callback) {
+			return [default_callback = MOV(callback)](
+				Map const& key_map,
+				std::string_view key,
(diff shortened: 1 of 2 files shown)


src/openvic-simulation/utility/StringUtils.hpp
```diff
@@ -3,11 +3,20 @@
...
+#include <cstddef>
 #include <cstdint>
 #include <cstring>
+#include <iterator>
 #include <string>
 #include <string_view>
 #include <type_traits>
+#include <utility>
+
+#include <fmt/format.h>
```
Spartan322, line 15:
```suggestion
```

src/openvic-simulation/dataloader/NodeTools.hpp
```diff
@@ -358,6 +371,22 @@ namespace OpenVic {
...
+				},
+				FWD(args)...
+			);
+		}
+		template<IsOrderedMap Map, typename... Args>
+		NodeCallback auto expect_dictionary_key_map_and_default_map(
+			Map&& key_map, MapKeyValueCallback<Map> auto&& default_callback, Args&&... args
 		) {
 			add_key_map_entries(FWD(key_map), FWD(args)...);
 			return expect_dictionary_key_map_and_default(FWD(key_map), FWD(default_callback));
```
Hop311, line 392: Here and in similar cases below, IMO it's better to put the `MapKeyValueCallback` version before the `KeyValueCallback` version so all functions have their dependencies above them rather than below.

Also you could add a helper function for turning `KeyValueCallback`s into `MapKeyValueCallback`s rather than doing it manually 3 times in this file

Hop311 (commented):
The Linux jobs should pass once you've rebased onto master

Hop311 (approved):

PR #372, opened 2025-03-10
Add `fixed_point_t::from_chars` method

Spartan322:
Add `fixed_point_t::from_chars_with_plus` method
Add positive fixed_point_t parse assertion
Add negative fixed_point_t parse assertion

### DIFF
--- a/src/openvic-simulation/types/fixed_point/FixedPoint.hpp
+++ b/src/openvic-simulation/types/fixed_point/FixedPoint.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cctype>
 #include <charconv>
 #include <concepts>
 #include <cstdint>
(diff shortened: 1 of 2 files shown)


src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -306,67 +306,70 @@ namespace OpenVic {
...
 			}
 
 			if (dot_pointer + 1 < end) {
-				// Non-empty fractional part
-				bool frac_successful = false;
-				result += parse_fraction(dot_pointer + 1, end, &frac_successful);
-				if (!frac_successful && successful != nullptr) {
-					*successful = false;
-				}
+				// Non-empty fractoional part
```
Hop311, line 351:
```suggestion
				// Non-empty fractional part
```

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -306,67 +306,70 @@ namespace OpenVic {
...
 
 			fixed_point_t result = _0();
-			if (successful != nullptr) {
-				*successful = true;
+			std::from_chars_result from_chars = {};
+			if (dot_pointer != begin) {
+				// Non-empty integer part
+				fixed_point_t adder;
+				from_chars = from_chars_integer(begin, dot_pointer, adder);
+				result += adder;
```
Hop311, line 343: Why not use `result` directly and avoid the intermediate `adder` variable? Or at least set `result = adder;` rather than adding when we know `result` is 0?

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -306,67 +306,70 @@ namespace OpenVic {
...
 			}
 
-			char const* dot_pointer = str;
-			while (*dot_pointer != '.' && ++dot_pointer != end) {}
-
-			if (dot_pointer == str && dot_pointer + 1 == end) {
+			char const* dot_pointer = begin;
+			while (dot_pointer && *dot_pointer != '.' && ++dot_pointer != end) {}
+			if (dot_pointer == begin && dot_pointer + 1 == end) {
 				// Invalid: ".", "+." or "-."
```
Hop311, line 333: This mentions `"+."`, but we wouldn't actually catch that here as we only catch `-` early, relying on the integer parsing function to get any `+` prefix. How does the integer parsing function handle `"+"`? Should we have extra checks to account for it and avoid `"+."` being successfully parsed as 0?

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -306,67 +306,70 @@ namespace OpenVic {
...
+					return { begin, std::errc::invalid_argument };
 				}
 			}
 
-			char const* dot_pointer = str;
-			while (*dot_pointer != '.' && ++dot_pointer != end) {}
-
-			if (dot_pointer == str && dot_pointer + 1 == end) {
+			char const* dot_pointer = begin;
+			while (dot_pointer && *dot_pointer != '.' && ++dot_pointer != end) {}
```
Hop311, line 331: How could we ever have `dot_pointer == nullptr` here? We already know `dot_pointer < end` and we terminate the loop if/when we reach it, so surely we can't overflow to get `dot_pointer == nullptr`?

Hop311 (approved):
LGTM, only potential issue is we don't support inputs with a `+` prefix, e.g. `+1.23` - should we support this? Some number parsing functions do as an opposite to `-`, although ofc it's technically superfluous. It's only an issue in that `+1.23` return 0 and a parsing failure

Hop311 (approved):

PR #374, opened 2025-03-10
Remove prefix "struct " from `OpenVic::ast::get_type_name` on MSVC

Spartan322:
Ensures get_type_name is always consistent between platforms.


Hop311 (approved):
BTW `utility::type_name` still has the `"struct "` prefix

Spartan322:
That's expected.

PR #379, opened 2025-03-11
Allow enabling/disabling great and world wars

Hop311:


src/openvic-simulation/politics/PoliticsInstanceManager.cpp
```diff
@@ -44,3 +44,19 @@ void PoliticsInstanceManager::unlock_ideology(Ideology const& ideology) {
 		);
 	}
 }
+
+void PoliticsInstanceManager::set_great_wars_enabled(bool enabled) {
+	// if (enabled && !great_wars_enabled) {
+		// TODO - trigger "Great Wars discovered!" popup
+	// }
+
+	great_wars_enabled = enabled;
```
schombert, line 53: presumably you also want turning off great wars to turn off world wars to mirror the logic below

Hop311: No, this is following the base game's logic where the `world_wars_enabled = yes` effect also enables great wars, but `great_wars_enabled = yes` and `*_wars_enabled = no` only affect their respective varaibles.

wvpm (approved):

PR #382, opened 2025-03-12
Add canal enabling

Hop311:


src/openvic-simulation/map/MapInstance.cpp
```diff
@@ -16,6 +22,14 @@ ProvinceInstance const& MapInstance::get_province_instance_from_definition(Provi
 	return province_instances.get_items()[province.get_index() - 1];
 }
 
+void MapInstance::enable_canal(canal_index_t canal_index) {
+	enabled_canals.insert(canal_index);
```
schombert, line 26: don't forget to clear any cached pathfinding data

wvpm (approved):

PR #259, opened 2024-12-23
Add pre-commit configuration file

Spartan322:
Add codespell checks and corrections
Add static check github action

PR #384, opened 2025-03-12
Apply clang-tidy fixes

Spartan322:
- Depends on #259


src/openvic-simulation/interface/GFXSprite.cpp
```diff
@@ -104,7 +104,7 @@ bool PieChart::_fill_key_map(case_insensitive_key_map_t& key_map) {
 	return ret;
 }
 
-LineChart::LineChart() : size {}, linewidth {} {}
+LineChart::LineChart() : linewidth {} {}
```
Hop311, line 107: `linewidth` could be initialised to 0 in its `PROPERTY` declaration, and all the constructors in this file could be removed, either replaced with `default` or with nothing at all

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -824,7 +824,9 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
 			}
 
 			// no further points
-			if (merge) points.push_back(merge_location);
+			if (merge) {
+				points.push_back(merge_location);
+			}
```
Hop311, line 829: This and the other `if` in this file have been expanded in my canal enabling PR

src/openvic-simulation/modifier/ModifierEffectCache.cpp
```diff
@@ -1,251 +1,37 @@
...
-	life_needs { nullptr },
-	everyday_needs { nullptr },
-	luxury_needs { nullptr } {}
+ModifierEffectCache::building_type_effects_t::building_type_effects_t() {}
+ModifierEffectCache::good_effects_t::good_effects_t() {}
+ModifierEffectCache::unit_type_effects_t::unit_type_effects_t() {}
+ModifierEffectCache::regiment_type_effects_t::regiment_type_effects_t() {}
+ModifierEffectCache::ship_type_effects_t::ship_type_effects_t() {}
+ModifierEffectCache::unit_terrain_effects_t::unit_terrain_effects_t() {}
+ModifierEffectCache::strata_effects_t::strata_effects_t() {}
```
Hop311, line 17: Can these be moved to the header as `default` or just removed altogether?

src/openvic-simulation/modifier/ModifierEffectCache.cpp
```diff
@@ -1,251 +1,37 @@
...
-	good_effects { nullptr },
+	  regiment_type_effects { nullptr },
 
-	/* UnitType Effects */
-	army_base_effects {},
-	regiment_type_effects { nullptr },
-	navy_base_effects {},
-	ship_type_effects { nullptr },
-	unit_terrain_effects { nullptr },
+	  ship_type_effects { nullptr }, unit_terrain_effects { nullptr },
```
Hop311, line 28: The formatting here seems inconsistent, why are these on the same line? Also I've tried to avoid "tab + 2 spaces" formatting for constructor lists, instead adding 2 spaces before the `:` so that all the later lines only need a tab

src/openvic-simulation/scripts/ConditionalWeight.hpp
```diff
@@ -28,7 +29,7 @@ namespace OpenVic {
 	struct ConditionalWeight {
 
 	private:
-		fixed_point_t PROPERTY(base);
+		fixed_point_t PROPERTY(base, fixed_point_t::_0());
```
Hop311, line 32: Across many files in this PR, `fixed_point_t`s are unnecessarily initialised to 0. It's not the end of the world, but it would be cleaner to avoid it where possible

src/openvic-simulation/scripts/Script.hpp
```diff
@@ -7,13 +7,13 @@ namespace OpenVic {
...
 	private:
-		ast::NodeCPtr _root;
+		ast::NodeCPtr _root = nullptr;
 
 	protected:
 		virtual bool _parse_script(ast::NodeCPtr root, _Context... context) = 0;
 
 	public:
-		Script() : _root { nullptr } {}
+		Script() {}
```
Hop311, line 16: Why not `= default`?

src/openvic-simulation/types/fixed_point/FixedPointLUT_2_16_EXP_e.hpp
```diff
@@ -24,5 +31,5 @@ static constexpr std::array<int64_t, LUT_2_16_EXP_e_SIZE> LUT_2_16_EXP_e {
 	3578144,
 	195360063,
 	582360139072,
-	5174916558532155392
+	5174916558532153344
```
Hop311, line 34: Interesting that this changed, I guess we've found an example of why we don't use floats and platform specific implementations of stuff like `exp`. Not that big a deal here tho as this is by far the least likely value to be used in any exponential calculations we do, but still worth bearing in mind for any future re-generations of this file

src/openvic-simulation/defines/AIDefines.cpp
```diff
@@ -3,46 +3,7 @@
...
-	peace_ally_war_exhaustion_mult {},
-	peace_ally_war_direction_mult {},
-	peace_ally_force_balance_mult {},
-	aggression_base {},
-	aggression_unciv_bonus {},
-	fleet_size {},
-	min_fleets {},
-	max_fleets {},
-	time_before_disband {} {}
+AIDefines::AIDefines() : fleet_size {}, min_fleets {}, max_fleets {} {}
```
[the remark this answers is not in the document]
Spartan322: I missed these on running clang-tidy.

`= default` is deceptive because if all the members are trivially constructible and aren't set, the type becomes trivially constructible which while great performance also introduces UB if you forget to initialize the value much like an uninitialized int. As a result I'm a bit undecided if we should use `= default` for cases where we wouldn't benefit from trivial constructibility already.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -16,29 +16,15 @@
...
-	total_paid_employees_count_cache { 0 },
-	total_owner_income_cache { },
-	total_employee_income_cache { },
-	employee_count_per_type_cache { &pop_type_keys }
-{ }
+)
+	: market_instance { new_market_instance }, production_type_nullable { new_production_type_nullable },
+	  revenue_yesterday { new_revenue_yesterday }, output_quantity_yesterday { new_output_quantity_yesterday },
+	  unsold_quantity_yesterday { new_unsold_quantity_yesterday }, size_multiplier { new_size_multiplier },
+	  employees { std::move(new_employees) }, employee_count_per_type_cache { &pop_type_keys } {}
```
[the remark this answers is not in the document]
Spartan322: There isn't a single tool which rewrites source files that doesn't so something like this, personally I don't really care but I don't believe we'll ever be able to use any formatting or linting tool then.

src/openvic-simulation/interface/GFXObject.cpp
```diff
@@ -55,7 +54,7 @@ Actor::Attachment::Attachment(std::string_view new_actor_name, std::string_view
 Actor::Animation::Animation(std::string_view new_file, fixed_point_t new_scroll_time)
   : file { new_file }, scroll_time { new_scroll_time } {}
 
-Actor::Actor() : model_file {}, scale { 1 }, idle_animation {}, move_animation {}, attack_animation {} {}
+Actor::Actor() {}
```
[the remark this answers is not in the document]
Spartan322: It doesn't change anything to be default, and while it can't really happen as is, we don't want Actor to be trivial constructible which `= default` can do wherever possible. This PR is not intended to change the calling semantics, just cleanup things according to clang-tidy.

src/openvic-simulation/scripts/ConditionalWeight.hpp
Spartan322: fixed_point_t always initialized to 0

src/openvic-simulation/types/fixed_point/FixedPointLUT_2_16_EXP_e.hpp
Spartan322: It would be nice to get a consistent value we can use for this that won't change on each run of the script, I have literally no idea how we'd achieve that with python tho.

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -93,15 +52,11 @@ bool StaticModifierCache::load_static_modifiers(ModifierManager& modifier_manage
 		);
 	};
 
+	#define SET_COUNTRY_MODIFIER(PROP) set_static_country_modifier(PROP);
+	#define SET_COUNTRY_MODIFIER2(PROP, _) set_static_country_modifier(PROP);
```
wvpm, line 56: What is the difference?

wvpm (approved):
Generally speaking a positve change.
I left some minor comments.

src/openvic-simulation/modifier/StaticModifierCache.cpp
Spartan322: The amount of arguments, macros don't have any concept of function overloading.

src/openvic-simulation/modifier/StaticModifierCache.cpp
```diff
@@ -141,20 +86,13 @@ bool StaticModifierCache::load_static_modifiers(ModifierManager& modifier_manage
 	set_country_event_modifier(in_bankruptcy_id, default_icon);
 	set_country_event_modifier(generalised_debt_default_id, default_icon);
 
+	#define SET_PROVINCE_MODIFIER(PROP) set_static_province_modifier(PROP);
+	#define SET_PROVINCE_MODIFIER2(PROP, _) set_static_province_modifier(PROP);
```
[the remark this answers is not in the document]
Spartan322: The amount of arguments, macros don't have any concept of function overloading.

src/openvic-simulation/modifier/StaticModifierCache.cpp
wvpm: Ah I see now, there are modifiers that pass 2 arguments (via `COUNTRY_MODIFIER_LIST`) but the second argument is ignored. You could make `SET_COUNTRY_MODIFIER2` use `SET_COUNTRY_MODIFIER`. That would make it clearer.

wvpm (approved):

PR #392, opened 2025-03-14
Add stack string methods to basic_colour_t

Spartan322:
Add fmt formatter support to basic_colour_t
Optimize `operator<<(std::ostream&, basic_colour_t const&)`
Remove stringstream in `ProvinceDefinition::to_string`

### DIFF
--- a/src/openvic-simulation/types/Colour.hpp
+++ b/src/openvic-simulation/types/Colour.hpp
@@ -13,10 +13,13 @@
 #include <ostream>
 #include <span>
 #include <string>
+#include <string_view>
 #include <system_error>
 #include <tuple>
 #include <type_traits>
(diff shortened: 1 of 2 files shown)


wvpm (approved):

src/openvic-simulation/types/Colour.hpp
```diff
@@ -466,17 +469,71 @@ namespace OpenVic {
 			return result;
 		}
 
-		inline std::string to_hex_string(bool alpha = colour_traits::has_alpha) const {
+		struct stack_string {
```
schombert, line 472: why not `using stack_string = std::array<char, array_length>` instead?

Hop311 (approved):

PR #393, opened 2025-03-14
Add stack string method to fixed_point_t

Spartan322:
Add fmt formatter support to fixed_point_t
Optimize `operator<<(std::ostream&, fixed_point_t const&)`
Add unit tests for fixed_point_t string methods

### DIFF
--- a/src/openvic-simulation/types/fixed_point/FixedPoint.cpp
+++ b/src/openvic-simulation/types/fixed_point/FixedPoint.cpp
@@ -1,0 +1,21 @@
+#include "FixedPoint.hpp"
+
+#include <fmt/format.h>
+
+using namespace OpenVic;
+
+fixed_point_t::stack_string fixed_point_t::to_array(size_t decimal_places) const {
+	stack_string str {};
+	std::to_chars_result result = to_chars(str.array.data(), str.array.data() + str.array.size(), decimal_places);
+	str.string_size = result.ptr - str.data();
+	return str;
+}
+
+auto fmt::formatter<fixed_point_t>::format(fixed_point_t fp, format_context& ctx) const -> format_context::iterator {
+	fixed_point_t::stack_string result = fp.to_array();
+	if (OV_unlikely(result.empty())) {
+		return formatter<string_view>::format(string_view {}, ctx);
+	}
+
+	return formatter<string_view>::format(string_view { result.data(), result.size() }, ctx);
+}
--- a/src/openvic-simulation/types/fixed_point/FixedPoint.hpp
+++ b/src/openvic-simulation/types/fixed_point/FixedPoint.hpp
@@ -6,14 +6,16 @@
 #include <cstdint>
 #include <cstdlib>
 #include <limits>
-#include <sstream>
 #include <string_view>
 #include <system_error>
+
+#include <fmt/core.h>
 
 #include "openvic-simulation/utility/Getters.hpp"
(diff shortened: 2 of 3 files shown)


src/openvic-simulation/types/fixed_point/FixedPoint.cpp
```diff
@@ -0,0 +1,21 @@
...
+	return str;
+}
+
+auto fmt::formatter<fixed_point_t>::format(fixed_point_t fp, format_context& ctx) const -> format_context::iterator {
+	fixed_point_t::stack_string result = fp.to_array();
+	if (OV_unlikely(result.empty())) {
+		return formatter<string_view>::format(string_view {}, ctx);
+	}
+
+	return formatter<string_view>::format(string_view { result.data(), result.size() }, ctx);
```
schombert, line 20: You seem to be going to a lot of trouble to optimize this, but you don't need to make the extra stack_string copy of the value at all. You could instead define a range or iterator that supplies the numerical characters to the formatter one at a time as you generate them, removing the need for any intermediate copy completely. So instead of writing the characters one by one to result and then copying them to the final destination, you could just write them one by one to the destination directly.

Spartan322: stack_string never copies here, because of return value optimization, (aka copy eliason) which every C/C++ compiler does since the late 90s, there is never a copy, there is only ever one value. In this specific case RVO is always guaranteed to happen, and by C++20 it is required by the standard to be done anyway.

schombert: That isn't the copy I am talking about. If you had, say, a generator-type function (i.e. a coroutine-type function) for generating the characters of the number you could pass that to the formatter and then it could use that to write each of the characters directly into the final buffer as they are generated. Currently, you generate the characters and write them into `result` and then pass the contents of `result` to the formatter which will then copy them into the final buffer. That is the unnecessary copy I am talking about.

Spartan322: fmt doesn't expose the buffer to formatters and that would break fmt's formatting API. Aside from being in accordance to the documentation, the concern regarding copying to the buffer is required in either case, even if that was exposed it would still be required perform comparatively more expensive operations on the buffer because of all the specifier requirements, doing it that way would break them while retaining the same cost.

schombert: Yes, I am aware that formatter doesn't directly expose the final buffer. What I am describing is how you can send things to the formatter in a way that doesn't require the copy that you are currently doing. This is one of the things that the format library was written to support, although it does require more complicated code than what you are doing now.

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -252,48 +254,137 @@ namespace OpenVic {
...
-				val = val.get_frac();
-				if (val != fixed_point_t::_0()) {
-					stream << ".";
-					do {
-						val *= 10;
-						stream << static_cast<char>('0' + val.to_int64_t());
-						val = val.get_frac();
-					} while (val != fixed_point_t::_0());
+
+			if (is_integer()) {
```
Hop311, line 262: Shouldn't this be `is_negative()`? Currently positive integers are printing with a minus sign. I don't know if there are any other issues stemming from this

Hop311 (approved):

PR #395, opened 2025-03-15
Add land and sea PointMap to MapDefinition

Spartan322:
Add land and sea AStarPathing to MapInstance
Add `PointMap::try_add_point`
Add `PointMap::shrink_to_fit`
Fix PointMap proxy helper functions
Add pathfinding tests to headless simulation

### DIFF
--- a/src/openvic-simulation/map/MapDefinition.cpp
+++ b/src/openvic-simulation/map/MapDefinition.cpp
@@ -1,5 +1,6 @@
 #include "MapDefinition.hpp"
 
+#include <array>
 #include <cctype>
 #include <charconv>
 #include <cstdint>
@@ -9,12 +10,14 @@
 #include <range/v3/algorithm/all_of.hpp>
 
 #include "openvic-simulation/dataloader/NodeTools.hpp"
+#include "openvic-simulation/map/ProvinceDefinition.hpp"
 #include "openvic-simulation/modifier/ModifierManager.hpp"
 #include "openvic-simulation/types/Colour.hpp"
 #include "openvic-simulation/types/OrderedContainers.hpp"
 #include "openvic-simulation/types/Vector.hpp"
 #include "openvic-simulation/utility/BMP.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
+#include "openvic-simulation/utility/StringUtils.hpp"
 
 using namespace OpenVic;
 using namespace OpenVic::NodeTools;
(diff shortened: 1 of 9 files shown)


src/openvic-simulation/map/MapDefinition.hpp
```diff
@@ -65,6 +66,9 @@ namespace OpenVic {
 
 		ProvinceDefinition::index_t PROPERTY(max_provinces);
 
+		PointMap PROPERTY_REF(path_map_land);
+		PointMap PROPERTY_REF(path_map_sea);
```
Hop311, line 70: The sea/water point map (i.e. all provinces naval units can move to) should contain all water provinces (maybe not lakes if possible) and all land provinces with ports.
For the land point map, there are of course all regular land provinces, but land units can technically path across water if each province has a fleet with sufficient transport capacity. I expect this will be handled somewhat differently to regular movement as the troops need to be attached to the fleet they're embarked with, but it'll still end up with some kind of point map including the water provinces with sufficient transport capacity.
I see you've implemented something for this with default-disabled points for potential transport-able provinces, just adding this comment to state the requirements.
image

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -969,6 +1010,23 @@ bool MapDefinition::generate_and_load_province_adjacencies(std::vector<LineObjec
...
+		ProvinceDefinition const* def = get_province_definition_by_index(index);
+		if (!def->water) {
+			continue;
+		}
+		for (auto const& adj : def->get_adjacencies()) {
+			if (adj.get_type() == ProvinceDefinition::adjacency_t::type_t::COASTAL) {
+				goto SKIP;
+			}
+		}
+		path_map_land.remove_point(index);
```
Hop311, line 1024: Would an `erase_if` function for point maps be useful?

Hop311 (approved):

src/openvic-simulation/map/MapDefinition.hpp
Spartan322: It already handles land units moving onto coastal sea provinces, forgot about ships docking at land provinces.

Spartan322: Though now that I think about it, the land map probably should include all sea provinces and just have them disabled.

Hop311 (approved):

PR #398, opened 2025-03-15
Change `BuildingInstance::expansion_progress` to fixed_point_t

Spartan322:
Fix fixed_point_t binary operators
Add `fixed_point_t::to<T>` conversion function

### DIFF
--- a/src/openvic-simulation/economy/BuildingInstance.cpp
+++ b/src/openvic-simulation/economy/BuildingInstance.cpp
@@ -6,10 +6,7 @@ BuildingInstance::BuildingInstance(BuildingType const& new_building_type, level_
   : HasIdentifier { new_building_type },
 	building_type { new_building_type },
 	level { new_level },
-	expansion_state { ExpansionState::CannotExpand },
-	start_date {},
-	end_date {},
-	expansion_progress { 0.0f } {}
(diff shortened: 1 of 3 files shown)


src/openvic-simulation/economy/BuildingInstance.cpp
```diff
@@ -34,8 +31,7 @@ void BuildingInstance::update_gamestate(Date today) {
 		end_date = start_date + building_type.get_build_time();
 		break;
 	case ExpansionState::Expanding:
-		expansion_progress =
-			static_cast<float>((today - start_date).to_int()) / static_cast<float>((end_date - start_date).to_int());
+		expansion_progress = fixed_point_t { (today - start_date).to_int() } / (end_date - start_date).to_int();
```
Hop311, line 34: I wonder if we have sufficient protection from division by zero here, it seems to depend on `building_type.get_build_time()` being non-zero 😕

Hop311 (approved):

src/openvic-simulation/economy/BuildingInstance.cpp
Spartan322: idk, I was just maintaining the existing functionality.

PR #401, opened 2025-03-16
Add constexpr unit tests

Spartan322:
Add constexpr support for `basic_colour_t::to_chars`
Fix constexpr support for `basic_colour_t::parse_from_chars`
Add constexpr support for `Date::to_chars`
Add constexpr support to `Date::from_string`
Add constexpr support to vector abs method
Fix constexpr support `StringUtils::from_chars`
Add constexpr support for `fixed_point_t::to_chars`
Fix constexpr support for `fixed_point_t::parse`
Fix constexpr support for fixed_point_t constants
Add Approx constexpr support
Add constexpr to_chars in StringUtils.hpp
Add constexpr abs in Utility.hpp

### DIFF
--- a/src/openvic-simulation/types/Colour.hpp
+++ b/src/openvic-simulation/types/Colour.hpp
@@ -24,6 +24,7 @@
 #include <range/v3/algorithm/rotate.hpp>
 
 #include "openvic-simulation/utility/Utility.hpp"
+#include "openvic-simulation/utility/StringUtils.hpp"
 
 namespace OpenVic {
 	template<typename ValueT, typename IntT, bool HasAlpha = true>
@@ -222,8 +223,9 @@ namespace OpenVic {
 			};
 		}
 
-		static constexpr basic_colour_t
-		from_floats(float r, float g, float b, float a = colour_traits::alpha_to_float(max_value))
+		static constexpr basic_colour_t from_floats( //
+			float r, float g, float b, float a = colour_traits::alpha_to_float(max_value)
+		)
 		requires(colour_traits::has_alpha)
 		{
 			return {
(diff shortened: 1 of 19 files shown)


src/openvic-simulation/types/Colour.hpp
```diff
@@ -619,7 +621,7 @@ namespace OpenVic {
 
 		// See https://stackoverflow.com/a/69869976
 		// https://github.com/Myndex/max-contrast/
-		constexpr basic_colour_t contrast() const {
+		basic_colour_t contrast() const {
```
Hop311, line 624: Why is this no longer `constexpr`?

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -458,6 +491,9 @@ namespace OpenVic {
 		constexpr std::from_chars_result from_chars_with_plus(char const* begin, char const* end) {
 			if (begin && *begin == '+') {
 				begin++;
+				if (begin < end && *begin == '-'){
```
Hop311, line 494:
```suggestion
				if (begin < end && *begin == '-') {
```

src/openvic-simulation/utility/Utility.hpp
```diff
@@ -137,4 +139,13 @@ namespace OpenVic::utility {
 
 	template<typename T, typename T2>
 	concept not_same_as = !std::same_as<T, T2>;
+
+	template<typename T> requires std::integral<T> || std::floating_point<T>
+	[[nodiscard]] inline constexpr T abs(T num) {
+		if (std::is_constant_evaluated()) {
```
Hop311, line 145: Why not use `if constexpr` here and in other palces `std::is_constant_evaluated` is used?

Hop311 (approved):

src/openvic-simulation/types/Colour.hpp
Spartan322: Because it can't be constexpr, it uses a C standard library math function, they don't get constexpr support until C++23, while GCC's standard library adds constexpr support to them anyway, no other version does. Replacing the std::pow isn't trivial work.

src/openvic-simulation/utility/Utility.hpp
Spartan322: Because that always returns true, if constexpr is always `std::is_constant_evaluated`, that's why `if consteval` was added to the language.

Hop311 (approved):

PR #403, opened 2025-03-18
Optimize vector construction

Spartan322:


wvpm (approved):
Economy changes look fine.
The hardcoded reserve values are arbitrary.
Probably too large for most countries anyway.
If need be, reserve extra room for great powers as they tend to have more.

Spartan322:
The arbitrary values are intended to be a bit above the usual default values for loading vanilla files, I'll add a comment to clarify that.

Spartan322:
Changed the default values for reserve calls to generally be a power of 2 below the max so as to optimize the majority of cases, this should reduce the needed allocations overall while keeping memory usage lower, this means the heuristic should generally only reallocate once or twice in the max case. Only exceptions to that are cases would be when a vector member variable for the managers has an expected vanilla size.

src/openvic-simulation/history/DiplomaticHistory.cpp
```diff
@@ -209,29 +219,45 @@ bool DiplomaticHistoryManager::load_diplomacy_history_file(
...
+
+	thread_local std::vector<WarHistory::war_participant_t> defenders;
+	// Default max vanilla defenders is 1, 1 is the max I've seen in mods
+	// Eliminates reallocations
+	defenders.reserve(1);
+
+	thread_local std::vector<WarHistory::added_wargoal_t> wargoals;
+	// Default max vanilla wargoals is 2, 2 is the max I've seen in mods
+	// Eliminates reallocations
+	wargoals.reserve(1);
```
wvpm, line 246: Shouldn't this be 2 then (based on the comment)?

src/openvic-simulation/history/DiplomaticHistory.cpp
```diff
@@ -209,29 +219,45 @@ bool DiplomaticHistoryManager::load_diplomacy_history_file(
...
 
 bool DiplomaticHistoryManager::load_war_history_file(DefinitionManager const& definition_manager, ast::NodeCPtr root) {
 	std::string_view name {};
-	std::vector<WarHistory::war_participant_t> attackers {};
-	std::vector<WarHistory::war_participant_t> defenders {};
-	std::vector<WarHistory::added_wargoal_t> wargoals {};
+
+	// Ensures that if ever multithreaded, only one vector is used per thread
+	// Else acts like static
+	thread_local std::vector<WarHistory::war_participant_t> attackers;
```
wvpm, line 233: When/if it were used concurrently, you still need to add locking right?
The comment suggests it's ready to be used concurrently.

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -790,37 +792,42 @@ bool MapDefinition::load_map_images(fs::path const& province_path, fs::path cons
...
-	const auto next_segment = [&river_data, &rivers_bmp](auto&& self, ivec2_t last_segment_end, uint8_t last_segment_direction, river_t& river) {
+	static const auto next_segment = [&river_data, &rivers_bmp](auto&& self, ivec2_t last_segment_end, uint8_t last_segment_direction, river_t& river) {
 		size_t idx = last_segment_end.x + last_segment_end.y * rivers_bmp.get_width();
 
-		std::vector<ivec2_t> points;
+		// Ensures that if ever multithreaded, only one vector is used per thread
+		// Else acts like static
+		thread_local std::vector<ivec2_t> points;
+		// Default max vanilla river points is 444
+		points.reserve(256);
```
wvpm, line 802: Shouldn't this be 512 then (based on the comment)?

wvpm (approved):

src/openvic-simulation/history/DiplomaticHistory.cpp
Spartan322: Some cases are 1, a singular allocation is cheaper then the excess memory in a few cases, avoiding a call to shrink_to_fit is most optimal.

src/openvic-simulation/history/DiplomaticHistory.cpp
Spartan322: attackers is an independent variable for each thread, say you have 10 threads, you'll have 10 attackers, one in each thread, there's no need to lock

src/openvic-simulation/map/MapDefinition.cpp
Spartan322: This should prevent excessive memory allocations in most cases, in the max case it would only reallocate once and in the others it would maintain lower memory usage, MSVC's capacity tends to act a bit differently, but overall the heuristic for reallocation size should often result in 256 by the time points finish.

wvpm (approved):

PR #412, opened 2025-04-02
Make terrain defence and movement_cost properties

Hop311:


src/openvic-simulation/modifier/ModifierManager.cpp
```diff
@@ -457,13 +457,6 @@ bool ModifierManager::setup_modifier_effects() {
 	ret &= register_base_province_modifier_effect(
 		modifier_effect_cache.boost_strongest_party, "boost_strongest_party", false, PROPORTION_DECIMAL, {}, has_no_effect
 	);
-	// When applied to provinces (terrain), combat_width is a multiplicative proportional decimal value.
-	ret &= register_base_province_modifier_effect(
```
wvpm, line 461: Should have never been a base_province_modifier_effect.
It only works like this on terrain. Good change!

wvpm (approved):

PR #416, opened 2025-04-16
Add OpenVicProject/lauf@8f3b2dc

Spartan322:


Spartan322:
This is confirmed to work on Windows, Mac, and Linux (I cannot test it on a local Mac config, but it does work on Github's Mac runner) and via MSVC, Clang, and GCC.

PR #417, opened 2025-04-18
RGO minimum wage

wvpm:


src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -347,23 +347,38 @@ void ResourceGatheringOperation::pay_employees() {
 		return;
 	}
 
+	CountryInstance const* const country_to_report_economy_nullable = location.get_country_to_report_economy();
+	fixed_point_t total_minimum_wage = fixed_point_t::_0();
+	if (country_to_report_economy_nullable != nullptr) {
+		CountryInstance const& country_to_report_economy = *country_to_report_economy_nullable;
+		for (auto const& [pop_type, employees_of_type] : employee_count_per_type_cache) {
+			total_minimum_wage += employees_of_type * country_to_report_economy.calculate_minimum_wage_base(pop_type);
```
wvpm, line 355: Store per employee as you need to handle `revenue < total_minimum_wage` by paying pops according to their portion of total_minimum_wage insead of simply by pop size.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -347,49 +347,78 @@ void ResourceGatheringOperation::pay_employees() {
...
+					return;
+				}
+
+				if (employee_pop_type->get_is_slave()) {
+					continue;
+				}
+
+				const pop_size_t employee_size = employee.get_size();
+				const fixed_point_t income_for_this_pop = std::max(
+					revenue_left * employee_size / total_paid_employees_count_cache,
```
wvpm, line 413: Handle case where `revenue > total_minimum_wage` but `revenue_left * employee_size / total_paid_employees_count_cache < minimum_wage` for some employees.

src/openvic-simulation/pop/Pop.cpp
```diff
@@ -561,7 +561,7 @@ void Pop::after_buy(void* actor, BuyResult const& buy_result) {
 	#undef CONSUME_NEED
 }
 
-void Pop::after_sell(void* actor, SellResult const& sell_result) {
+void Pop::after_sell(void* actor, SellResult const& sell_result, std::vector<fixed_point_t>& reusable_vector) {
```
Hop311, line 564: Maybe add a comment explaining that `reusable_vector` is needed for the function signature to match `GoodMarketSellOrder::callback_t`, even though it's not actually used in this function

src/openvic-simulation/economy/trading/MarketSellOrder.hpp
```diff
@@ -25,8 +27,8 @@ namespace OpenVic {
 			after_trade { new_after_trade }
 			{}
 
-		constexpr void call_after_trade(SellResult const& sell_result) const {
-			after_trade(actor, sell_result);
+		constexpr void call_after_trade(SellResult const& sell_result, std::vector<fixed_point_t>& reusable_vector) const {
+			after_trade(actor, sell_result, reusable_vector);
```
Hop311, line 31: I assume we're always absolutely sure `after_trade` is not null, so we don't need a null-check to prevent crashes

src/openvic-simulation/economy/trading/MarketInstance.cpp
```diff
@@ -47,15 +47,20 @@ void MarketInstance::place_market_sell_order(MarketSellOrder&& market_sell_order
 	GoodDefinition const& good = market_sell_order.get_good();
 	if (OV_unlikely(market_sell_order.get_quantity() <= 0)) {
 		Logger::error("Received MarketSellOrder for ", good, " with quantity ", market_sell_order.get_quantity());
-		market_sell_order.call_after_trade(SellResult::no_sales_result());
+		std::vector<fixed_point_t> throw_away_vector {}; //TODO inject
```
Hop311, line 50: Would there be any point in making this `static`, so we don't have to construct a new one every time this code is executed? Although on the other hand, I'm not sure if that would create issues with memory location/requiring costly memory lookups. @Spartan322 may be able to help.

(same for the other `throw_away_vector` below)

src/openvic-simulation/economy/trading/GoodMarket.cpp
```diff
@@ -275,6 +275,8 @@ void GoodMarket::execute_orders(std::vector<fixed_point_t>& reusable_vector_0, s
 				}
 			}
 		}
+		reusable_vector_0.clear();
+		reusable_vector_1.clear();
```
Hop311, line 279: We clear `reusable_vector_1` here but then don't use it again, yet we do use `reusable_vector_0` later and don't clear it after that?
I may not understand the code properly, but it seems that either `reusable_vector_1` doesn't need to be cleared here, or `reusable_vector_0` should be cleared again after being used on line 301 below

src/openvic-simulation/economy/production/Employee.hpp
```diff
@@ -5,9 +5,15 @@
...
 	struct Employee {
 	private:
+		Pop& pop;
 		pop_size_t PROPERTY_RW(size);
+		fixed_point_t PROPERTY_RW(minimum_wage_cached);
 	public:
-		Pop& pop;
+		constexpr Pop& get_pop() {
+			return pop;
+		}
```
Hop311, line 14: You could wrap `pop`'s declaration in `PROPERTY_REF` to automatically generate both mutable and const `get_pop` reference getters

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -347,49 +347,107 @@ void ResourceGatheringOperation::pay_employees() {
...
+			);
+			const fixed_point_t owner_share = std::min(
+				fixed_point_t::_2() * total_owner_count_in_state_cache / total_worker_count_in_province_cache,
+				upper_limit
+			);
 
-	if (total_paid_employees_count_cache > 0) {
-		for (Employee& employee : employees) {
-			Pop& employee_pop = employee.pop;
+			for (Pop* owner_pop_ptr : *owner_pops_cache_nullable) {
```
Hop311, line 385: Are we certain `owner_pops_cache_nullable != nullptr`?

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -347,49 +347,107 @@ void ResourceGatheringOperation::pay_employees() {
...
+				Pop& owner_pop = *owner_pop_ptr;
+				const fixed_point_t income_for_this_pop = std::max(
+					revenue_left * (owner_share * owner_pop.get_size()) / total_owner_count_in_state_cache,
+					fixed_point_t::epsilon() //revenue > 0 is already checked, so rounding up
+				);
+				owner_pop.add_rgo_owner_income(income_for_this_pop);
+				total_owner_income_cache += income_for_this_pop;
+			}
+			revenue_left *= (fixed_point_t::_1() - owner_share);
```
Hop311, line 394: Minor style comment, but the brackets are unnecessary here. Also this could be rewritten as:
```cpp
revenue_left -= revenue_left * owner_share;
```
I'm not sure which is better performance-wise, if there even is any difference.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -347,49 +347,107 @@ void ResourceGatheringOperation::pay_employees() {
...
+				const fixed_point_t income_for_this_pop = std::max(
+					revenue_left * employee_size / count_workers_to_be_paid,
+					fixed_point_t::epsilon() //revenue > 0 is already checked, so rounding up
+				);
+
+				if (income_for_this_pop < minimum_wage) {
+					incomes[i] = minimum_wage;
+					revenue_left -= minimum_wage;
+					count_workers_to_be_paid -= employee_size;
+					i = -1; //Restart loop and skip minimum incomes. This is required to spread the remaining revenue again.
```
Hop311, line 434: I don't understand, `i -= 1` means we re-run the loop for the current `i` value, but then that'll hit the `continue` on line 421 as `incomes[i] == minimum_wage` - why not just leave `i` unchanged here and let it progress to the next employee as normal? Assuming `employee.get_minimum_wage_cached()` won't have changed by the time we call it again for the same `i` value?

Hop311 (approved):
Works both in headless SIM and passing time when run via the editor.
I'm not 100% sure how/why everything works, and I've left some comments about that + other minor details. Overall though I have no issues with merging 👍

src/openvic-simulation/economy/trading/GoodMarket.cpp
wvpm: The pattern is use, then clear.
When `reusable_vector_0` is passed to `call_after_trade`, that method will use and clear it.
The vector has to be cleared before passing it on. That's why it's cleared here.

src/openvic-simulation/economy/trading/MarketInstance.cpp
wvpm: I thought I'd replace it later but decided to inject it in this PR.
So I updated it.

src/openvic-simulation/economy/trading/MarketSellOrder.hpp
wvpm: With the current code, after_trade is never null.
I think it's stored as a function pointer and we don't do a null check in the constructor.
So in theory you could pass a null pointer.

Out of scope for this PR but it could be added.

src/openvic-simulation/pop/Pop.cpp
wvpm: Added it.

src/openvic-simulation/economy/trading/MarketSellOrder.hpp
Hop311: Yeah I understand, a null error could only happen due to a silly programmer mistake rather than a more serious runtime issue. Still, to catch such a silly mistake we could add `assert(after_trade != nullptr);` to the buy and sell order constructors, or even better make a compile time check such as providing the lambda function's type as a template parameter and initialising `after_trade` with it (although I'm not sure about how `static` member functions would fit into this).

src/openvic-simulation/utility/ThreadPool.cpp
```diff
@@ -47,12 +47,12 @@ void ThreadPool::loop_until_cancelled(
...
 			case work_t::PROVINCE_TICK:
 				for (ProvinceInstance& province : provinces_chunk) {
-					province.province_tick(current_date, reusable_pop_values);
+					province.province_tick(current_date, reusable_pop_values, reusable_vector_0);
 				}
 				break;
 			case work_t::PROVINCE_INITIALISE_FOR_NEW_GAME:
 				for (ProvinceInstance& province : provinces_chunk) {
-					province.initialise_for_new_game(current_date, reusable_pop_values);
+					province.initialise_for_new_game(current_date, reusable_pop_values, reusable_vector_0);
```
Hop311, line 55: Since we're reusing these vectors over and over again, it would be a good idea to choose which to use based on expected size, e.g. use 1 for all the big vector needs and 2 for the smaller ones. At least it would be good to put some thought in rather than always going with 1 first, then 2 if necessary.

Hop311 (approved):
LGTM, left some more comments, might do a more thorough look at the RGO code tomorrow but I'm fairly sure it's fine after the comments are addressed

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: `revenue_left -= total_owner_income_cache` seems better now that you focus my attention here.
I'll change it.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
wvpm: Yes, because it checks `total_owner_count_in_state_cache > 0` before the loop.

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -347,49 +350,107 @@ void ResourceGatheringOperation::pay_employees() {
...
-			PopType const* employee_pop_type = employee_pop.get_type();
-			if (employee_pop_type == nullptr) {
-				Logger::error("employee has nullptr pop_type.");
-				return;
+		if (total_paid_employees_count_cache == 0) {
+			//scenario slaves only
+			//Money is removed from system in Victoria 2.
+		} else {
+			std::vector<fixed_point_t>& incomes = reusable_vector;
+			incomes.resize(employees.size());
```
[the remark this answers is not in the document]
wvpm: I expect every using function to clear before exiting or passing the vector along if that function has modified the vector.

I debated this at the start with Gemini. I wondered whether I should expect clean or dirty shared objects.
Gemini convinced me to expect clean objects and always return them clean.

src/openvic-simulation/utility/ThreadPool.cpp
wvpm: Interesting idea and certainly worth investigating. Still out of scope for this PR.
We can do that when/if we notice performance issues.

src/openvic-simulation/economy/trading/MarketSellOrder.hpp
wvpm: Again, out of scope.

I wonder what you mean though. It already uses a custom lambda type `using callback_t = void (*)(actor_t, SellResult const&, std::vector<fixed_point_t>&);`

src/openvic-simulation/economy/trading/MarketSellOrder.hpp
```diff
@@ -8,7 +10,7 @@ namespace OpenVic {
 
 	struct GoodMarketSellOrder {
 		using actor_t = void*;
-		using callback_t = void (*)(actor_t, SellResult const&);
+		using callback_t = void (*)(actor_t, SellResult const&, std::vector<fixed_point_t>&);
```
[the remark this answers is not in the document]
wvpm: It must never be null, so reference makes sense. If a callback doesn't use it, that's fine.
As for the throw-away vectors, that is already resolved. Also throw away referrred to the parent method throwing away the vector after passing it to the callback. The name has no relation to what the callback does with it.
I'll leave it as is.

src/openvic-simulation/economy/trading/MarketSellOrder.hpp
Hop311: That's a function pointer type, I mean a template parameter for the type of the specific lambda used. But yeah, out of scope, I can explain exactly what I mean in a voice call sometime

src/openvic-simulation/economy/trading/MarketInstance.hpp
```diff
@@ -30,7 +30,7 @@ namespace OpenVic {
 		fixed_point_t get_max_money_to_allocate_to_buy_quantity(GoodDefinition const& good_definition, const fixed_point_t quantity) const;
 		fixed_point_t get_price_inverse(GoodDefinition const& good_definition) const;
 		void place_buy_up_to_order(BuyUpToOrder&& buy_up_to_order);
-		void place_market_sell_order(MarketSellOrder&& market_sell_order);
+		void place_market_sell_order(MarketSellOrder&& market_sell_order, std::vector<fixed_point_t>& reusable_vector);
 		void execute_orders();
```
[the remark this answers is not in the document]
Hop311: Maybe `execute_good_orders` and `execute_all_orders`? Ideally it would represent whatever your personal understanding of/intentions for the functions.

src/openvic-simulation/economy/trading/GoodMarket.cpp
```diff
@@ -291,15 +293,16 @@ void GoodMarket::execute_orders(std::vector<fixed_point_t>& reusable_vector_0, s
...
-			market_sell_order.call_after_trade({
-				quantity_sold,
-				money_gained
-			});
+			market_sell_order.call_after_trade(
+				{
+					quantity_sold,
+					money_gained
+				},
+				reusable_vector_0
```
Hop311, line 301: Is the `reusable_vector_0` cleared after being used here? It's cleared above, but the clears below have been removed

Hop311 (approved):
LGTM, only one potential issue but that may be my misunderstanding.

As for ensuring reusable vectors are always cleared after being used, could you move the clear calls to `ThreadPool` so they're always cleared after being used in any call? That would eliminate the need for devs to remember to always clear them directly atfter the point of use. (Although this may not be feasible if they're used multiple times in certain calls from `ThreadPool`...)

src/openvic-simulation/economy/trading/GoodMarket.cpp
wvpm: after_trade clears it if it uses it.

PR #421, opened 2025-05-12
Add foonathan/memory@79d054caaa491d9b6ed7cc65a3a84b495578e6c1

Spartan322:
Add memory tracking in debug


Hop311 (commented):
All seems to be working well on Windows, just a few comments:
- Will this have any meaningful impact on non-debug/release builds?
- Even for debug builds, do you expect any noticable performance impact? Some quick testing of mean tick duration showed a slight increase, but there was quite a lot of variation between runs anyway both with and without these changes, even with larger amounts of test ticks.
- There are still a few places using the old types without memory tracking, e.g. `std::vector<csv::LineObject>` (which could be changed with a Dataloader PR adding the allocator as a template arg?), and lots of `std::string` uses. Have you considered these exceptions and decided it isn't worth tracking their memory usage? What's the scope/rule for memory that should be tracked?

I would approve right now, but just to be safe I'd rather test this together with the corresponding main OpenVic repo changes first in case anything comes up that requires SIM tweaks.

Spartan322:
> * Will this have any meaningful impact on non-debug/release builds?

No, the memory tracking is disabled in release and everything else basically mimics the standard C++ allocator by default.

> * Even for debug builds, do you expect any noticable performance impact? Some quick testing of mean tick duration showed a slight increase, but there was quite a lot of variation between runs anyway both with and without these changes, even with larger amounts of test ticks.

None that I've seen, it probably technically does, but it would require nitpicking at benchmarks and it be so minuscule its almost certainly not worth considering, its functionally implemented the same as Godot, memory tracking.

> * There are still a few places using the old types without memory tracking, e.g. `std::vector<csv::LineObject>` (which could be changed with a Dataloader PR adding the allocator as a template arg?), and lots of `std::string` uses. Have you considered these exceptions and decided it isn't worth tracking their memory usage? What's the scope/rule for memory that should be tracked?

The rule was not changing anything int he dataloader, std::vector and memory::vector are not compatible cause they are different types, anywhere that the conversion wouldn't be fairly trivial for those reasons I skipped. Majority of those cases also aren't a big deal because the memory is small there and doesn't last long.

>  I would approve right now, but just to be safe I'd rather test this together with the corresponding main OpenVic repo changes first in case anything comes up that requires SIM tweaks.

It would require some OpenVic modifications anyway, so whatever is needed.

src/headless/main.cpp
```diff
@@ -155,17 +167,21 @@ static bool run_headless(fs::path const& root, std::vector<std::string>& mods, b
...
 
+	print_bytes("Instance Setup");
+
 	Logger::info("===== Starting game session... =====");
 	ret &= game_manager.start_game_session();
 
 	// This triggers a gamestate update
 	ret &= game_manager.update_clock();
 
+	print_bytes("Game Session Post-Start");
```
wvpm, line 178: Good idea to debug the memory per step in the sim loading.

src/openvic-simulation/console/ConsoleInstance.cpp
```diff
@@ -32,15 +34,15 @@ ConsoleInstance::ConsoleInstance(InstanceManager& instance_manager, write_func_t
...
+void ConsoleInstance::write(memory::string&& message) {
 	write_func(current_colour, std::move(message));
 }
 
 void ConsoleInstance::set_write_func(write_func_t&& new_write_func) {
 	write_func = std::move(new_write_func);
 }
 
-void ConsoleInstance::default_write_func(OpenVic::colour_t colour, std::string&& message) {
+void ConsoleInstance::default_write_func(OpenVic::colour_t colour, std::string_view message) {
```
wvpm, line 45: Are we sure this method is never called with a temporary value?

src/openvic-simulation/pathfinding/PointMap.hpp
```diff
@@ -21,7 +21,8 @@ namespace OpenVic {
 		using points_key_type = uint64_t;
 		using points_value_type = Point;
 		using points_pair_type = std::pair<points_key_type, points_value_type>;
-		using points_allocator_type = std::allocator<points_pair_type>;
+		using points_allocator_type =
```
wvpm, line 24: (Minor thing since it's a type alias)
Keep something after the `=`
For example
```cpp
using points_allocator_type = foonathan::memory::std_allocator<
points_pair_type,
memory::tracker<foonathan::memory::default_allocator>
>;
```

wvpm: Alternatively create an alias for `foonathan::memory::std_allocator`, how about `mem_allocator`?

src/openvic-simulation/politics/Issue.hpp
```diff
@@ -5,6 +5,8 @@
 #include "openvic-simulation/scripts/ConditionScript.hpp"
 #include "openvic-simulation/scripts/EffectScript.hpp"
 #include "openvic-simulation/types/IdentifierRegistry.hpp"
+#include "openvic-simulation/utility/Containers.hpp"
+#include "openvic-simulation/utility/Containers.hpp"
```
wvpm, line 9: Duplicate

src/openvic-simulation/pathfinding/PointMap.hpp
Spartan322: Type aliases aren't recognized as having indentation by any formatter.

wvpm (approved):

PR #424, opened 2025-05-17
Add commit info to GameManager

Spartan322:


SConstruct
```diff
@@ -29,6 +29,27 @@ SConscript("deps/SCsub", "env")
...
+        file.write(
+            """\
+#pragma once
+
+#include <cstdint>
+#include <string_view>
+
+namespace OpenVic {{
+	static constexpr std::string_view SIM_COMMIT_HASH = "{git_hash}";
+	static constexpr const uint64_t SIM_COMMIT_TIMESTAMP = {git_timestamp};
```
Hop311, line 44: It might be worth adding a suffix to the timestamp integer literal to ensure it's interpreted correctly (i.e. unsigned).
When testing I saw big enough integer literal values cause a warning:
```cpp
// Warning: Integer literal is too large to be represented in a signed integer type, interpreting as unsigned
static constexpr const uint64_t N = 18446744073709551615;
```

Hop311 (dismissed):
LGTM 👍 Tested on Windows

SConstruct
Spartan322: The number shouldn't be that large.

Hop311 (approved):

PR #427, opened 2025-05-28
Negative tariffs

wvpm:
Negative tariffs are handled with a 1 tick delay.
This allows the game to accurately project the cost and ensure the actual costs never exceed the projected costs (which might lead to bankruptcy otherwise).

In `Pop::after_buy` the pop keeps a running total of `yesterdays_import_value`.
This is cleared after calling `CountryInstance::request_salaries_and_welfare_and_import_subsidies` in `Pop::pop_tick`.

In the `update_gamestate` methods, `yesterdays_import_value` is propagated from province via state to country, just like total population.

In `CountryInstance::country_tick_before_map` the `projected_import_subsidies` are calculated and `actual_import_subsidies` is determined after handling budget cuts due to cash shortages.

`CountryInstance::request_salaries_and_welfare_and_import_subsidies` then calculates the subsidies per pop and adds them to the pop's cash (not income as it isn't taxed).

For alternative implementations and the discussion, see https://discord.com/channels/1063392556160909312/1064542494114725909/1376923223094067360
This PR implements option B.

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.cpp
+++ b/src/openvic-simulation/country/CountryInstance.cpp
@@ -66,7 +66,7 @@ CountryInstance::CountryInstance(
 	taxable_income_by_pop_type { &pop_type_keys },
 	effective_tax_rate_by_strata { &strata_keys },
 	tax_rate_slider_value_by_strata { &strata_keys },
-	import_value_mutex { std::make_unique<std::mutex>() },
+	actual_net_tariffs_mutex { std::make_unique<std::mutex>() },
 
 	/* Technology */
 	technology_unlock_levels { &technology_keys },
@@ -1088,6 +1088,9 @@ void CountryInstance::_update_budget() {
 	projected_military_spending_unscaled_by_slider /= Pop::size_denominator;
 	projected_pensions_spending_unscaled_by_slider /= Pop::size_denominator;
 	projected_unemployment_subsidies_spending_unscaled_by_slider /= Pop::size_denominator;
+	projected_import_subsidies = has_import_subsidies()
+		? -effective_tariff_rate * yesterdays_import_value
+		: fixed_point_t::_0();
 }
 
 fixed_point_t CountryInstance::calculate_pensions_base(
@@ -1156,6 +1159,7 @@ void CountryInstance::_update_politics() {
 
 void CountryInstance::_update_population() {
 	total_population = 0;
+	yesterdays_import_value = fixed_point_t::_0();
 	national_literacy = 0;
 	national_consciousness = 0;
 	national_militancy = 0;
@@ -1176,6 +1180,7 @@ void CountryInstance::_update_population() {
 
 	for (State const* state : states) {
 		total_population += state->get_total_population();
+		yesterdays_import_value += state->get_yesterdays_import_value();
 
 		// TODO - change casting if pop_size_t changes type
 		const fixed_point_t state_population = fixed_point_t::parse(state->get_total_population());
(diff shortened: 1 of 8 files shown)


schombert:
It sounds like you have already discussed this, but FYI that is not how V2 works -- a negative tariff is applied directly to the costs of goods as the pop buys them from the global market. Adding to pop cash has four potential knock-on effects, besides the immediate consequences to purchasing behavior. First, it doesn't cover factories and artisans importing good for production, although I imagine you plan to cover this in the same way. Secondly, it will not exactly match the next days purchasing, so some of that money may go towards domestic goods, or it may simply not cover all of the imports. Thirdly, by adding to pop cash it may have knock-on effects for pop savings, which in turn will affect the size of national loans that can be given by the national bank. Fourthly, it will have issues accounting for variable supply on the global market. As a one-step issue, that pops, etc would like to import a certain quantity doesn't mean that the market has enough available to satisfy that demand, especially when, with subsidies, the pops will be demanding more. This is solvable, but there is still the second-step issue. If the global supply is growing, then estimations of how much may be imported from the previous day will undershoot and the pops will be given less subsidy money than they ought. More troublesome, if the global supply is shrinking, then the pops may be given money to buy goods that simply aren't available, resulting in negative tariffs just acting as a general cash gift to pops.

wvpm:
> I don't doubt that it can be made to work, I was mostly pointing out the ways it would deviate from standard V2 behavior. I was under the impression that open vic was prioritizing replicating v2 as closely as possible. I can tell you from experience that even slightly changing how the economy works is going to change how militancy and ideology develop over the course of the game and will lead to complaints by the purists.

The 100% purists will probably continue to play Victoria 2.
OpenVic will be as close as possible to Victoria 2 and fully support mods without patches.

schombert:
Well, if the economy differs from the base game, some mods will experience militancy issues since they are balanced around the vanilla economy, requiring patches to rebalance them for your changes. Of course, I don't know how common negative tariffs are, so this may not be a huge issue in this particular case.

Spartan322:
> Well, if the economy differs from the base game, some mods will experience militancy issues since they are balanced around the vanilla economy, requiring patches to rebalance them for your changes.

If that were to actually be an issue, it can be reported as a vic2 discrepancy bug which will need to be fixed, however unless its complained about or reported its pointless to worry over theory of slight differences that produce the same effects, if balance becomes an issue we will rebalance the economy to correct it, but until its demonstrated that such balance is an issue, implementations are not intended to be 1-to-1, but to replicate it to a practical indecipherable degree to the players.

schombert:
I don't understand that approach in this case. If we were talking about something where we don't know exactly how V2 works, then I would agree with you: the best you can do is be as close as you can and then try to fix any discrepancies. However, how negative tariffs function are not one of those black boxes we can only guess at, so there doesn't seem to be a need to intentionally introduce a discrepancy only to maybe later go back and fix it. (as an aside, I suggest consulting with desolt if you need more details on the game internals) My experience simulating the V2 economy in PA is that the mods are extremely sensitive to small differences. Even when we were trying to replicate the base game economy back in 1.0, small differences always made the economic situation in mods behave differently over the course of the game, even when we managed to get the vanilla economy behaving basically the same. Adjusting things to match the vanilla economy only made mods behave worse, if anything. TGC in particular manipulates the cash pops have to make its economy work. If you implement changes of your own that also affect pop cash, as you are planning here, it seems likely that it will interact oddly with TGC.

Spartan322:
There's already a number of things which we know are distinct from vic2 in our implementation despite knowledge of how vanilla works, its not inherently true that every discrepancy produces a practical or even inherently mathematical difference, we already know we're gonna need to do balancing for existing systems anyway, getting something to test is better, whether something is a notable discrepancy can be crowdsourced, if someone can actually notice then its probably worth fixing. But I couldn't give less of a damn about a theoretical discrepancy, if I don't have a something empirical to work with it does not matter. This is the same exact approach as any performance issue or any theoretical bug, unless you can prove it as an issue then its at best pure bikeshedding.

That aside if necessary such functionality can be worked into to being unit testable so we can validate inputs and outs, its a lot more reliable to ensure cross-platform outcomes.

Spartan322 (approved):

PR #445, opened 2025-06-11
Move const reference vectors to span

Spartan322:
"Everywhere that you use std::vector<T> const& could you instead use std::span<T> as it reduces indirection and generalizes the use case instead of being dependent on the vector type being passed. It would allow any contiguous finite array of elements, dynamic or static, to be passed in and also doesn't depend on an allocator to describe the type."
https://github.com/OpenVicProject/OpenVic-Simulation/pull/291#issuecomment-2961121242

### DIFF
--- a/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
+++ b/src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
@@ -1,5 +1,6 @@
 #include "ResourceGatheringOperation.hpp"
 
+#include <span>
 #include <vector>
 
 #include "openvic-simulation/economy/production/Employee.hpp"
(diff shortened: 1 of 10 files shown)


src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
```diff
@@ -68,7 +69,7 @@ void ResourceGatheringOperation::initialise_rgo_size_multiplier() {
 	ProvinceInstance& location = *location_ptr;
 	ModifierEffectCache const& modifier_effect_cache = location.get_modifier_effect_cache();
 	ProductionType const& production_type = *production_type_nullable;
-	std::vector<Job> const& jobs = production_type.get_jobs();
+	std::span<const Job> jobs = production_type.get_jobs();
```
wvpm, line 72: Shouldn't these be `const std:span<const T>`?
They are entirely read-only.

src/openvic-simulation/map/ProvinceInstance.cpp
```diff
@@ -1,5 +1,9 @@
 #include "ProvinceInstance.hpp"
 
+#include <span>
+
+#include "openvic-simulation/DefinitionManager.hpp"
```
wvpm, line 5: This is already imported at line 9

wvpm (dismissed):

src/openvic-simulation/economy/production/ResourceGatheringOperation.cpp
Spartan322: spans can't be const, it does literally nothing.

wvpm (approved):

PR #447, opened 2025-06-13
Fix `Employee::get_pop() const` return type warning

Spartan322:


Valerus9 (approved):
All specified changes are actually implemented and no other unspecified change have been done.

Within my right as a developer, I approve these changes.

PR #448, opened 2025-06-14
Fix space/tab indentation mixing in MapDefinition

Spartan322:
Fix break/continue lacking braces in MapDefinition


Valerus9 (approved):
Format has been standardised and no other unspecified changes happend.

With my right as a developer, I approve these changes.

PR #449, opened 2025-06-14
Move even more const reference vectors to span

Spartan322:
- Continuation of #445

Add forwadable_span for span of forward types (thanks Apple)
Add OV_NO_UNIQUE_ADDRESS for msvc::no_unique_address support


wvpm (dismissed):

src/openvic-simulation/GameManager.cpp
```diff
@@ -21,7 +21,7 @@ bool GameManager::load_mod_descriptors(std::span<const std::string> descriptors)
 	return true;
 }
 
-bool GameManager::set_roots(Dataloader::path_vector_t const& roots, Dataloader::path_vector_t const& replace_paths) {
+bool GameManager::set_roots(std::span<const fs::path> roots, std::span<const fs::path> replace_paths) {
```
BrickPi, line 24: Should probably alias these to a `Dataloader::path_span_t` type seeing as we use it quite a bit.

BrickPi (approved):
lgtm 👏

PR #454, opened 2025-06-16
Add `PathingBase::get_iterator_by_id`

Spartan322:
Add `PathingBase::generate_path_result`
Add early exit for disabled from point in PathingBase

### DIFF
--- a/src/openvic-simulation/pathfinding/PathingBase.hpp
+++ b/src/openvic-simulation/pathfinding/PathingBase.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <functional>
+
 #include <fmt/core.h>
 
 #include "openvic-simulation/pathfinding/PointMap.hpp"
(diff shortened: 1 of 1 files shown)


src/openvic-simulation/pathfinding/PathingBase.hpp
```diff
@@ -100,138 +102,106 @@ namespace OpenVic {
...
-			search_iterator from_it = search.find(from_id);
-			if (from_it == search.end()) {
-				PointMap::points_value_type const* from_ptr = point_map->try_get_point(from_id);
-				if (from_ptr != nullptr) {
-					from_it = search.insert({ from_id, { .point = from_ptr } }).first;
+		inline search_iterator get_iterator_by_id(PointMap::points_key_type id) {
+			search_iterator it = search.find(id);
+			if (it == search.end()) {
+				PointMap::points_value_type const* ptr = point_map->try_get_point(id);
+				if (ptr != nullptr) {
```
wvpm, line 109: Wouldn't it be an error if ptr is null?
Why would a province id not be on the map?
And how can the pathfinding recover from this? (I think it just gives up)

src/openvic-simulation/pathfinding/PathingBase.hpp
```diff
@@ -100,138 +102,106 @@ namespace OpenVic {
...
 				// Use closest point instead.
 				to_it = last_closest_point;
 			}
 
-			search_iterator p = to_it;
-			size_t pc = 1; // Begin point
-			while (p != from_it) {
-				pc++;
-				p = p.value().prev_point;
+			return generate_path_result<ivec2_t>(from_it, to_it, [](search_iterator it) -> ivec2_t {
```
wvpm, line 172: Why not make this a static method instead of a lambda?

wvpm (approved):

src/openvic-simulation/pathfinding/PathingBase.hpp
Spartan322: Why? Its a one-off projection.

src/openvic-simulation/pathfinding/PathingBase.hpp
```diff
@@ -100,138 +102,106 @@ namespace OpenVic {
...
-			bool found_route = _solve(from_it, to_it, current_pass++, allow_partial_path);
-			if (!found_route) {
-				if (!allow_partial_path || last_closest_point == search.end()) {
-					return std::vector<ivec2_t>();
-				}
-
-				// Use closest point instead.
-				to_it = last_closest_point;
-			}
+			OV_ERR_FAIL_COND_V_MSG(it == search.end(), it, fmt::format("Can't get point by id: {} doesn't exist.", id));
```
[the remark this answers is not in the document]
Spartan322: No, maybe could move it after 111, moving it to 109 would miss failure cases causing bugs. But there's no point moving it cause it keeps returns together and reduces the indentation, the compiled result is exactly the same anyhow.

PR #457, opened 2025-06-16
Detect unused templates by the absence of output_goods instead of postfix

wvpm:
#291 added name based detection of templates to prevent erroring on unused templates.
The names are entirely arbitrary and should not be relied upon. Instead each production type requires an `output_goods` entry.
Templates don't have this and can be detected by the absence of `output_goods`.


src/openvic-simulation/economy/production/ProductionType.cpp
```diff
@@ -1,6 +1,8 @@
 #include "ProductionType.hpp"
 
 #include <openvic-dataloader/v2script/Parser.hpp>
+#include "dataloader/NodeTools.hpp"
+#include "utility/ErrorMacros.hpp"
```
Spartan322, line 5:
```suggestion

#include "openvic-simulation/dataloader/NodeTools.hpp"
#include "openvic-simulation/utility/ErrorMacros.hpp"
```

src/openvic-simulation/economy/production/ProductionType.cpp
```diff
@@ -207,34 +210,51 @@ bool ProductionTypeManager::load_production_types_file(
...
+	}
 
 	size_t expected_types = 0;
 
 	/* Pass #1: find and store template identifiers */
 	ordered_set<std::string_view> templates;
 	ordered_map<std::string_view, std::string_view> template_target_map;
 	bool ret = expect_dictionary(
-		[this, &expected_types, &templates, &template_target_map, &template_symbol](std::string_view key, ast::NodeCPtr value) -> bool {
+		[this, &expected_types, &templates, &template_target_map, &template_symbol,&output_goods_symbol]
```
BrickPi, line 224:
```suggestion
		[this, &expected_types, &templates, &template_target_map, &template_symbol, &output_goods_symbol]
```

src/openvic-simulation/economy/production/ProductionType.cpp
```diff
@@ -207,34 +210,51 @@ bool ProductionTypeManager::load_production_types_file(
...
-					return false;
-				}
+			bool has_found_template = false;
+			const bool is_parsing_template_success = expect_key(
+				template_symbol,
+				expect_identifier(assign_variable_callback(template_id)),
+				&has_found_template
+			)(value);
+
+
```
BrickPi, line 236:
```suggestion
```

BrickPi: Double newline.

src/openvic-simulation/economy/production/ProductionType.cpp
wvpm: Why break there?
To me that seems unnatural and arbitrary.

BrickPi: It shouldn't be a linebreak, I meant to add a space between `template_symbol` and `output_goods_symbol` as there isn't one in the current code.

PR #460, opened 2025-06-18
Add signal and signal_property based on sigslot

Spartan322:
Add cow_ptr for copy-on-write data
Add null_mutex
Add spin_mutex
Add signal tests

### DIFF
--- a/src/openvic-simulation/types/CowPtr.hpp
+++ b/src/openvic-simulation/types/CowPtr.hpp
@@ -1,0 +1,197 @@
+#pragma once
+
+#include <atomic>
+#include <concepts>
+#include <cstddef>
+#include <type_traits>
+#include <utility>
+
+#include "openvic-simulation/utility/Utility.hpp"
+
+namespace OpenVic {
+	/**
+	 * A Copy-On-Write pointer
+	 *
+	 * Any unmodified copy refers to the same pointer in memory
+	 * Any modified copy allocates a new memory pointer (given use_count() is not 1)
+	 */
+	template<typename T>
+	class cow_ptr {
+		struct payload;
+
+	public:
+		using element_type = T;
+		using reference = element_type&;
+		using const_reference = element_type const&;
+		using pointer = element_type*;
+		using const_pointer = element_type const*;
+
+		OV_ALWAYS_INLINE cow_ptr() : data(new payload) {}
+
+		template<typename U>
+		requires(!std::same_as<std::decay_t<U>, cow_ptr>)
+		OV_ALWAYS_INLINE explicit cow_ptr(U&& x) : data(new payload(std::forward<U>(x))) {}
+
+		template<typename... Args>
+		requires std::constructible_from<T, Args...>
+		OV_ALWAYS_INLINE cow_ptr(std::in_place_t, Args&&... args) : data(new payload(std::forward<Args>(args)...)) {}
+
+		template<typename U, typename... Args>
+		requires std::constructible_from<T, std::initializer_list<U>&, Args...>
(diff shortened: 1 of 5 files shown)


src/openvic-simulation/types/CowPtr.hpp
```diff
@@ -0,0 +1,145 @@
...
+#include <concepts>
+#include <cstddef>
+#include <type_traits>
+#include <utility>
+
+#include "openvic-simulation/utility/Utility.hpp"
+
+namespace OpenVic {
+	template<typename T>
+	class cow_ptr {
```
wvpm, line 13:
```suggestion
	class cow_ptr { //copy-on-write pointer
```
Idk if cow is a common term for cpp programmers.
For the rest of the world, it refers to the animal.

src/openvic-simulation/types/CowPtr.hpp
```diff
@@ -0,0 +1,145 @@
...
+		OV_ALWAYS_INLINE explicit operator bool() const {
+			return data == nullptr;
+		}
+
+		friend inline void swap(cow_ptr& x, cow_ptr& y) {
+			using std::swap;
+			swap(x.data, y.data);
+		}
+
+	private:
```
wvpm, line 89: Shouldn't private be before public?
See also https://github.com/OpenVicProject/OpenVic/blob/master/docs/styleguide-cpp.md#22-dos-and-donts

src/openvic-simulation/types/CowPtr.hpp
```diff
@@ -0,0 +1,145 @@
...
+		OV_ALWAYS_INLINE element_type const& operator*() const {
+			return read();
+		}
+
+		OV_ALWAYS_INLINE element_type const* operator->() const {
+			return get();
+		}
+
+		OV_ALWAYS_INLINE size_t use_count(std::memory_order memory_order = std::memory_order::relaxed) const {
+			return static_cast<bool>(*this) ? data->count.load(memory_order) : 0;
```
wvpm, line 77: Why cast `this*` to a `bool` instead of checking `!= nullptr`?

src/openvic-simulation/types/Signal.hpp
wvpm, line 1: Considering how massive this file is and the amount of imports, why not split it into a hpp and cpp file?

src/openvic-simulation/types/CowPtr.hpp
wvpm: Just because there is code that violates the styleguide doesn't mean your code should.
If you don't agree with the styleguide, just change it.

src/openvic-simulation/types/Signal.hpp
wvpm, line 1: This file is too large and convoluted to understand.
Types should be split into their own files and implementation should be moved to cpp.

I suppose you could treat it as a black box and just 'just trust me bro' the sigslot repo.
If that's the case, why not add a gitmodule?

Is there even custom code in here or is it all copied from another repo?

wvpm: Okay. Templating does restrict our options...
How about moving the openvic code to the top and then informing the reader via comment that the rest is copied from the sigslot?

Spartan322: It already has a comment that references the basis for the Signal header and includes the license.

src/openvic-simulation/types/CowPtr.hpp
Spartan322: I wouldn't mind it being fixed up to align with how we do things, but dealing with that over the code hasn't been much of a priority, and honestly I'm more concerned over us not using clang-format since the format mixing is a lot more disruptive then style has been.

PR #462, opened 2025-06-19
Add invalidation signals to pathfinding classes

Spartan322:
- Depends on #460

### DIFF
--- a/src/openvic-simulation/pathfinding/PathingBase.hpp
+++ b/src/openvic-simulation/pathfinding/PathingBase.hpp
@@ -5,6 +5,7 @@
 #include <fmt/core.h>
 
 #include "openvic-simulation/pathfinding/PointMap.hpp"
+#include "openvic-simulation/types/Signal.hpp"
 #include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
 #include "openvic-simulation/utility/ErrorMacros.hpp"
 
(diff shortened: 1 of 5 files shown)


src/openvic-simulation/pathfinding/PathingBase.hpp
```diff
@@ -67,13 +68,31 @@ namespace OpenVic {
 			search_iterator begin_point, search_iterator end_point, uint64_t pass, bool allow_partial_path
 		) = 0;
 
+		void _on_point_invalidated(search_key_type id) {
```
wvpm, line 71: Shouldn't this be in .cpp?

src/openvic-simulation/pathfinding/PointMap.cpp
```diff
@@ -44,7 +48,11 @@ bool PointMap::try_add_point(
 		return false;
 	}
 
+	size_t prev_capacity = get_point_capacity();
 	points[id] = Point { .position = position, .weight_scale = weight_scale, .enabled = true };
+	if (OV_unlikely(prev_capacity != get_point_capacity())) {
```
wvpm, line 53: When would this occur?

src/openvic-simulation/pathfinding/PathingBase.hpp
Spartan322: PathingBase doesn't have a source file, nor could it have one.

wvpm (approved):

wvpm (approved):

PR #463, opened 2025-06-19
Support multiple script nodes by combining them

wvpm:
In vanilla, event 16000 (A House Divided) has multiple `immediate` effect nodes. They need to be combined.
Third Age uses multiple `allow` nodes for a political reform. Victoria 2 simply combines them via AND.

To support this, scripts now store a vector of nodes instead of a single one.

### DIFF
--- a/src/openvic-simulation/scripts/Script.hpp
+++ b/src/openvic-simulation/scripts/Script.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <vector>
+
 #include "openvic-simulation/dataloader/NodeTools.hpp"
 
 namespace OpenVic {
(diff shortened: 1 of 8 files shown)


Spartan322 (approved):

src/openvic-simulation/scripts/Script.hpp
```diff
@@ -1,38 +1,46 @@
...
-					Logger::error("Null/missing script node!");
+					Logger::error("Missing script node!");
 				}
 				return can_be_null;
 			}
-			const bool ret = _parse_script(_root, context...);
-			_root = nullptr;
+
+			if (_root_nodes.size() > 1) {
+				Logger::warning("Multiple root nodes");
```
wvpm, line 38: How can we add context here?
I could expose a method to check for this and let the caller log a warning + context or pass context along.

Preferably the logger should have context.

In .NET you can use logging scopes to provide context.
https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#log-scopes

PR #467, opened 2025-06-21
Change fixed_point_t constants to constexpr variables

Spartan322:
Prefer integer fixed_point_t assignment
Prefer integer fixed_point_t return

### DIFF
--- a/src/openvic-simulation/economy/production/ArtisanalProducer.cpp
+++ b/src/openvic-simulation/economy/production/ArtisanalProducer.cpp
@@ -36,12 +36,12 @@ void ArtisanalProducer::artisan_tick(
 	GoodDefinition::good_definition_map_t& demand = reusable_map_1;
 
 	//throughput scalar, the minimum of stockpile / desired_quantity
-	fixed_point_t inputs_bought_fraction = fixed_point_t::_1(),
-		inputs_bought_numerator= fixed_point_t::_1(),
-		inputs_bought_denominator= fixed_point_t::_1();
+	fixed_point_t inputs_bought_fraction = 1,
+		inputs_bought_numerator= 1,
+		inputs_bought_denominator= 1;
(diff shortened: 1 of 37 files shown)


src/openvic-simulation/economy/production/ArtisanalProducer.cpp
```diff
@@ -118,32 +118,32 @@ void ArtisanalProducer::artisan_tick(
 	const fixed_point_t total_cash_to_spend = pop.get_cash().get_copy_of_value() / max_cost_multiplier;
 	MarketInstance const& market_instance = pop.get_market_instance();
 
-	if (total_cash_to_spend > fixed_point_t::_0() && !goods_to_buy_and_max_price.empty()) {
+	if (total_cash_to_spend > fixed_point_t::_0 && !goods_to_buy_and_max_price.empty()) {
 		//Figure out the optimal amount of goods to buy based on their price, stockpiled quantiy & demand
-		fixed_point_t max_possible_satisfaction_numerator= fixed_point_t::_1(),
-			max_possible_satisfaction_denominator= fixed_point_t::_1();
+		fixed_point_t max_possible_satisfaction_numerator= fixed_point_t::_1,
+			max_possible_satisfaction_denominator= 1;
```
Hop311, line 124: Why is there a mix of changes from `fixed_point_t::_1()` to `fixed_point_t::_1` and `1`, here and similarly throughout the PR?

Hop311 (approved):

src/openvic-simulation/economy/production/ArtisanalProducer.cpp
Spartan322: Assignment and returns should use the integer constructor because it should make no difference, the rest generally need to use the constant for type-safety (and potentially compiler optimization) reasons.

Hop311 (approved):
LGTM 👍

PR #468, opened 2025-06-23
Add country diplomacy getters to CountryInstance

Spartan322:
Implement `CountryInstance::is_at_war`
Fix OpinionType wrapping around in operator--
Fix double assignment in `CountryRelationManager::set_at_war_with`


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -191,14 +193,125 @@ bool CountryInstance::is_secondary_power() const {
...
+
+void CountryInstance::set_alliance_with(CountryInstance& country, bool alliance) {
+	country_relations_manager.set_country_alliance(this, &country, alliance);
+}
+
+bool CountryInstance::is_at_war_with(CountryInstance const& country) const {
+	return war_enemies.contains(&country);
+}
+
+void CountryInstance::set_at_war_with(CountryInstance& country, bool at_war) {
```
wvpm, line 223: How would this be used?
I don't see any link to the war instance.

PR #469, opened 2025-06-24
Add ArmyAStarPathing for ArmyInstance pathing

Spartan322:
Add exiled property to ArmyInstance

### DIFF
--- a/src/openvic-simulation/pathfinding/AStarPathing.cpp
+++ b/src/openvic-simulation/pathfinding/AStarPathing.cpp
@@ -8,8 +8,14 @@
 #include <range/v3/algorithm/find.hpp>
 #include <range/v3/algorithm/heap_algorithm.hpp>
 
+#include "openvic-simulation/country/CountryInstance.hpp"
+#include "openvic-simulation/map/MapDefinition.hpp"
+#include "openvic-simulation/map/MapInstance.hpp"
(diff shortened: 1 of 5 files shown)


src/openvic-simulation/pathfinding/AStarPathing.cpp
```diff
@@ -91,3 +97,34 @@ bool AStarPathing::_solve(search_iterator begin_point, search_iterator end_point
...
+
+ArmyAStarPathing::ArmyAStarPathing(MapInstance const& map)
+	: map_instance(map), AStarPathing(&map.get_map_definition().get_path_map_land()) {}
+
+bool ArmyAStarPathing::_is_point_enabled(search_const_iterator it) const {
+	if (!AStarPathing::_is_point_enabled(it)) {
+		return false;
+	}
+
+	if (OV_unlikely(army_instance == nullptr)) {
```
wvpm, line 109: Is this (and province null) a 'normal' situation or a corrupt state that you don't want to throw on?

Spartan322: The army instance being nullptr is a default state if one doesn't set the instance, the province being null shouldn't be able to happen in our logic, but it is possible to construction a PointMap that doesn't respect the province indexes which needs to be protected, this is why these are marked with OV_unlikely since they should be very uncommon cases.

PR #479, opened 2025-07-03
Add NavyAstarPathing

Spartan322:
Add `CountryInstance::can_navy_units_enter`
Rename `CountryInstance::can_units_enter` to `can_army_units_enter`

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.cpp
+++ b/src/openvic-simulation/country/CountryInstance.cpp
@@ -307,9 +307,14 @@ void CountryInstance::set_embassy_banned_from(CountryInstance& country, Date unt
 	country_relations_manager.set_embassy_banned_date(this, &country, until);
 }
 
-bool CountryInstance::can_units_enter(CountryInstance const& country) const {
+bool CountryInstance::can_army_units_enter(CountryInstance const& country) const {
 	// TODO: include war allies, puppets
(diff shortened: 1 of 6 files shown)


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -307,11 +307,16 @@ void CountryInstance::set_embassy_banned_from(CountryInstance& country, Date unt
...
 
-bool CountryInstance::can_units_enter(CountryInstance const& country) const {
+bool CountryInstance::can_army_units_enter(CountryInstance const& country) const {
 	// TODO: include war allies, puppets
 	return is_at_war_with(country) || has_military_access_to(country);
 }
 
+bool CountryInstance::can_navy_units_enter(CountryInstance const& country) const {
+	// TODO: include war allies, puppets
+	return this == &country || has_military_access_to(country);
```
wvpm, line 317: Move `this == &country` into `has_military_access_to` so it works for `can_army_units_enter` as well.

Spartan322: has_military_access_to is a diplomacy check, it shouldn't be used for that, see #480

wvpm: Okay, that also works.

wvpm (approved):

PR #486, opened 2025-07-07
Exclude colonies for admin eff

wvpm:
Colonial states are excluded for national administrative efficiency calculation.
Note that there still is a difference with Victoria 2. The difference is significantly reduced with this change though.


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -23,6 +23,7 @@
 #include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
 #include "openvic-simulation/types/PopSize.hpp"
 #include "openvic-simulation/types/SliderValue.hpp"
+#include "types/IndexedMap.hpp"
```
Spartan322, line 26:
```suggestion
#include "openvic-simulation/types/IndexedMap.hpp"
```

Spartan322:
Could the PR and commit message be `Remove colonies from admin efficiency` instead?

Spartan322 (approved):

PR #487, opened 2025-07-09
Round fixed_point_t to int away from zero

wvpm:
fixed_point_t only supported truncating to int32_t and int64_t.
Rounding before truncating could be done via explicit `floor()` or `ceil()`.

This PR just adds `to_int64_t_rounded()` which rounds away from zero.
There is also a `to_int32_t_rounded()` which merely casts the int64 to int32.


BrickPi (approved):
lgtm

PR #496, opened 2025-07-12
Use index from HasIndex in IndexedMap

wvpm:
IndexedMap searches for an item to get the index.
If the item inherits from HasIndex, it has a `get_index()` method which should be used instead.
This is faster than searching.

### DIFF
--- a/src/openvic-simulation/types/IndexedMap.hpp
+++ b/src/openvic-simulation/types/IndexedMap.hpp
@@ -1,9 +1,9 @@
 #pragma once
 
 #include <concepts>
-#include <vector>
 
 #include "openvic-simulation/types/fixed_point/FixedPointMap.hpp"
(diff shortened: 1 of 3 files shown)


src/openvic-simulation/types/IndexedMap.hpp
```diff
@@ -310,10 +310,14 @@ namespace OpenVic {
 		}
 
 		constexpr size_t get_index_from_item(key_ref_type key) const {
-			if (has_keys() && keys.data() <= &key && &key <= &keys.back()) {
-				return std::distance(keys.data(), &key);
+			if constexpr (InheritsFromHasIndex<Key>::value) {
```
Spartan322, line 313:
```suggestion
			if constexpr (utility::is_derived_from_specialization_of<Key, HasIndex>) {
```

PR #498, opened 2025-07-13
Use IndexedMap for HasIndex types

wvpm:
Depends on #496 
Thanks to #496 IndexedMap uses the index directly from HasIndex and doesn't have to search for keys.
This makes IndexedMap faster than `tsl::ordered_map`.

This PR makes all maps with key types that implement HasIndex use IndexedMap instead of `tsl::ordered_map`.
Note IndexedMap should be created with all possible keys. In most cases this comes from `IdentifierRegistry.hpp` `expect_item_decimal_indexed_map`. The ThreadPool gets the keys explicitly (via instance manager).

Dropped in favour of #499

### DIFF
--- a/src/openvic-simulation/pop/Pop.cpp
+++ b/src/openvic-simulation/pop/Pop.cpp
@@ -1,5 +1,3 @@
-#include <utility>
-
 #define KEEP_DO_FOR_ALL_TYPES_OF_INCOME
 #define KEEP_DO_FOR_ALL_TYPES_OF_EXPENSES
 #include "Pop.hpp"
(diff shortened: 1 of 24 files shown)


src/openvic-simulation/pop/Pop.cpp
```diff
@@ -567,28 +568,25 @@ void Pop::after_buy(void* actor, BuyResult const& buy_result) {
...
-				get_country_to_report_economy_nullable->report_pop_need_consumption(*pop.type, good_definition, consumed_quantity); \
-			} \
-			const fixed_point_t expense = fixed_point_t::mul_div( \
-				money_spent, \
-				consumed_quantity, \
-				quantity_bought \
-			); \
-			pop.add_##need_category##_needs_expense(expense); \
-		}
+		const fixed_point_t need_category##_desired_quantity = pop.need_category##_needs[good_definition]; \
```
wvpm, line 571: `pop.need_category##_needs` needs to be initialised with all good definitions as keys.

PR #504, opened 2025-07-15
Clearly distinguish between 0- & 1-based indexing

wvpm:
#496 uses the index from HasIndex to directly index arrays. This is great for performance but requires 0-based indexing.
Provinces use 1-based indexing for their 'external id' (data-loading, UI & console).

To remedy this, provinces now use 0-based indices for HasIndex and clearly marked 1-based indices for pathfinding, console, game actions and such.

### DIFF
--- a/src/openvic-simulation/console/ConsoleInstance.cpp
+++ b/src/openvic-simulation/console/ConsoleInstance.cpp
@@ -208,7 +208,7 @@ ProvinceInstance* ConsoleInstance::validate_province_index(std::string_view valu
 		return nullptr;
 	}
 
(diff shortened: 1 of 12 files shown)


wvpm:
> I feel like we should have instead do a separate name for index and position, index is 0-based, position is 1-based.

As mentioned in the discord, renaming the 1-based value to province id or something similar seems fine.
Currently the 1-based value is still heavily used for array indexing.

It should be refactored to use the 0-based index. That requires too much effort for a bug fix.
Let's fix the bug first and refactor it later.

Spartan322 (approved):

PR #510, opened 2025-07-29
Add sigslot to COPYRIGHT

Spartan322:


COPYRIGHT
```diff
@@ -38,6 +38,11 @@ Comment: plf::colony
 Copyright: 2019, mattreecebentley
 License: Zlib
 
+Files: src/openvic-simulation/types/Signal.hpp
+Comment: sigslot
+Copyright: 2017, Pierre-Antoine Lacaze
+License: Expat
```
wvpm, line 44: https://github.com/palacaze/sigslot?tab=License-1-ov-file
Says it uses the MIT License.
According to Gemini Expat is simply an alias for MIT license.
Not sure why Expat is chosen, but it seems fine then.

Spartan322: Expat is always MIT cause MIT can refer to at least two difference licenses.

PR #511, opened 2025-07-29
Expose vector properties via span

wvpm:
Instead `get_##NAME()` returning a `vector<T> const&` it now returns a `forwardable_span<const T>`.
This ensures the collection is const, but if T is a pointer to a non-const type, the instance it points to is non-const.
This is required for #508 as it requires non-const instances to get data that may update a cache.
If desired, the forwardable_span return could be exposed as non-const with the `vector<T> const&` remaining as const variant.
That would result in different return types between const and non-const `get_##NAME()`.

### DIFF
--- a/src/openvic-simulation/utility/Getters.hpp
+++ b/src/openvic-simulation/utility/Getters.hpp
@@ -255,3 +255,13 @@ public: \
 		return NAME; \
 	} \
 ACCESS:
+
+#define SPAN_PROPERTY(NAME) SPAN_PROPERTY_ACCESS(NAME, private)
+#define SPAN_PROPERTY_ACCESS(NAME, ACCESS) \
+	NAME; \
+\
+public: \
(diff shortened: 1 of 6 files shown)


Spartan322:
When its not a reference, we do `const T` because that's the type.

src/openvic-simulation/utility/Getters.hpp
```diff
@@ -255,3 +255,13 @@ public: \
 		return NAME; \
 	} \
 ACCESS:
+
+#define SPAN_PROPERTY(NAME) SPAN_PROPERTY_ACCESS(NAME, private)
+#define SPAN_PROPERTY_ACCESS(NAME, ACCESS) \
+	NAME; \
+\
+public: \
+	[[nodiscard]] constexpr auto get_##NAME() const -> OpenVic::utility::forwardable_span<decltype(NAME)::value_type const> { \
```
Spartan322, line 264:
```suggestion
	[[nodiscard]] constexpr auto get_##NAME() const -> OpenVic::utility::forwardable_span<const decltype(NAME)::value_type> { \
```

Spartan322 (approved):

PR #526, opened 2025-08-05
Instances are immovable, uncopyable and constructed in the manager constructor

wvpm:
Instances require pointer stability. This was handled via delayed setup methods.
Now it's handled by deleting the move and copy constructors.

`std::vector` (used in `IdentifierRegistry`) requires movable types when calling `reserve(size_t new_capacity)`.
To avoid this, a new FixedVector type is added.

FixedVector is designer for immovable uncopyable types.
It requires the capacity to be specified at construction time and cannot change its capacity.
All items must be emplaced.

Since `IdentifierRegistry` doesn't support FixedVector and all instance types have a `size_t get_index() const` method, IndexedFlatMap is used.

IndexedFlatMap now supports immovable and uncopyable types as well as movable and/or copyable types.
It does this by condtionally storing the items in an std::vector or a FixedVector.

All manager are refactored to create their instances in their constructor instead of a delayed setup method.


schombert:
In my experience, pointer stability is more easily achieved by storing std::unique_ptr s to the objects in your collections instead. While this does introduce an additional layer of indirection, it is also a lot less likely to be a source of bugs. Or you can just consistently store and pass the immutable index/handle to the object instead of a pointer to it, which solves the problem in the other direction, by not requiring address stability.

src/openvic-simulation/utility/memory_tracking/tracker.hpp
Spartan322, line 1: I don't see any reason not to put this in `memory_tracking/MemoryTracker` then.

wvpm: Good point. I've moved it. Note I couldn't get it to compile with MemoryTracker in OpenVic::memory, only in OpenVic::utility.

PR #527, opened 2025-08-07
Container documentation

wvpm:
Documentation describing the various C++ types used to store items.
Use this to pick between container types.

### DIFF
--- a/docs/Containers.md
+++ b/docs/Containers.md
@@ -1,0 +1,61 @@
+# Container types
+This file describes the various C++ types used to store items.
+Use this to pick between container types.
+
+## Sequence containers
+Sequence containers are array/list-like. They consist of a sequence of items.
+- array<T,SIZE> / C-array
+- vector<T>
+- cow_vector<T>
+- FixedVector<T>
+- colony<T>
+- deque<T>
+- stack<T>
+- queue<T>
+
+### array<T,SIZE> / C-array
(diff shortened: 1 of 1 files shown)


docs/Containers.md
```diff
@@ -0,0 +1,61 @@
...
+[Official documentation](https://en.cppreference.com/w/cpp/container/vector.html)
+Usage: `memory::vector<T> values;`
+Requires:
+	- T must be moveable (for reallocations).
+Capacity: expandable and shrinkable
+Size: expandable and shrinkable
+Note `vector<bool>` acts as a bit set, see [official documentation](https://en.cppreference.com/w/cpp/container/vector_bool.html).
+
+### cow_vector<T>
+Source: src\openvic-simulation\types\CowVector.hpp
```
schombert, line 38: Do you have tests for this type?
```template<typename T, typename Allocator>
	inline constexpr cow_vector<T, Allocator>::size_type cow_vector<T, Allocator>::payload::content_size =
		std::max<size_type>(1ul, (sizeof(payload) - sizeof(array)) / sizeof(T));
```
would cause the first member of the array to overlap the pointer to the array itself for some sizes of T. If you were trying to have it round up to the largest number of T sufficient to contain the non-T payload data, it would be ` = (sizeof(payload) + sizeof(T) - 1) / sizeof(T)`

PR #564, opened 2025-10-06
Make conditional std::source_location use unconditional

Spartan322:
While we'll be moving to dropping mac support, libc++ has supported source_location for a while now anyway, so there's not reason to keep the conditional around any longer since every major library implementation now handles source_location properly anyway.

### DIFF
--- a/src/openvic-simulation/utility/Logger.hpp
+++ b/src/openvic-simulation/utility/Logger.hpp
@@ -3,63 +3,20 @@
 #include <cstddef>
 #include <iostream>
 #include <mutex>
+#include <source_location>
 #include <string_view>
 #include <vector>
 #include <version>
 
 #include <function2/function2.hpp>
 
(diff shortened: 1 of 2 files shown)


src/openvic-simulation/utility/Logger.hpp
```diff
@@ -3,63 +3,23 @@
...
-	};
-#endif
-
 	class Logger final {
 		using log_func_t = fu2::function_view<void(memory::string&&)>;
 		using log_queue_t = memory::queue<memory::string>;
 
 	public:
-#ifdef __cpp_lib_source_location
 		using source_location = std::source_location;
```
Hop311, line 22: We don't need this alias anymore, we can just use `std::source_location` directly (although ofc leaving it in does no harm)

Hop311 (approved):

Hop311 (approved):

PR #566, opened 2025-10-07
Replace Logger with spdlog

Spartan322:
Draft until the tedious work of moving everything over to spdlog is done.

Add SPDLOG_LEVEL environment variable to globally disable log levels at compile-time

Replace LogScope with spdlog::scope
Add ovfmt::validate for format nullptr validation
Fix erroneous node pointer printing in Event.cpp
Fix erroneous node pointer printing in Condition.cpp


wvpm (dismissed):
Not sure what changed with the last commit.
Still looks good (I focused on economy files and such).

wvpm (approved):

wvpm (approved):

src/headless/main.cpp
```diff
@@ -1,11 +1,18 @@
...
+#include <memory>
 #include <random>
 #include <string>
 
 #include <fmt/base.h>
 #include <fmt/chrono.h>
 
+#include <spdlog/common.h>
+#include <spdlog/sinks/callback_sink.h>
+#include <spdlog/sinks/basic_file_sink.h>
```
Hop311: These two aren't in alphabetical order

src/headless/main.cpp
```diff
@@ -28,10 +35,17 @@ inline static void print_memory_usage(
 	std::string_view prefix, std::source_location const& location = std::source_location::current()
 ) {
 #ifdef DEBUG_ENABLED // memory tracking will return 0 without DEBUG_ENABLED
-	Logger::info<std::string_view&, const char(&)[16], uint64_t&&, const char(&)[7]>(
-		prefix, " Memory Usage: ", OpenVic::utility::MemoryTracker::get_memory_usage(), " Bytes", location
+#if SPDLOG_ACTIVE_LEVEL <= SPDLOG_LEVEL_INFO
+	spdlog::log(
+#ifndef SPDLOG_NO_SOURCE_LOC
+		spdlog::source_loc { location.file_name(), static_cast<int>(location.line()), location.function_name() },
+		#else
```
Hop311: Every other C++ `#else` in our code is at the beginning of its line. Although this one technically isn't necessary as the `spdlog::source_loc {` and `},` could go outside the `#ifndef` block and just automatically have no args when `SPDLOG_NO_SOURCE_LOC` is set.

src/headless/main.cpp
```diff
@@ -48,34 +62,34 @@ static void print_rgo(ProvinceInstance const& province) {
...
-			", production_type: ", production_type.get_identifier(),
-			", size_multiplier: ", rgo.get_size_multiplier().to_string(3),
-			", output_quantity_yesterday: ", rgo.get_output_quantity_yesterday().to_string(3),
-			", revenue_yesterday: ", rgo.get_revenue_yesterday().to_string(3),
-			", total owner income: ", rgo.get_total_owner_income_cache().to_string(3),
-			", total employee income: ", rgo.get_total_employee_income_cache().to_string(3),
-			"\n\temployees:"
-		);
 
+		std::string employees;
```
Hop311: Is there any reason for using `std::string` here instead of `memory::string`? Do we want to restrict use of `memory::string` to actual game stuff so our memory usage metrics don't include more indepdentant stuff?

src/headless/main.cpp
```diff
@@ -174,21 +217,28 @@ static bool run_headless(fs::path const& root, memory::vector<memory::string>& m
...
 	if (run_tests) {
-		Testing testing { game_manager.get_definition_manager() };
-		std::cout << std::endl << "Testing Loaded" << std::endl << std::endl;
+		// std::shared_ptr<spdlog::logger> testing_logger = spdlog::basic_logger_st(
+		// 	"testing",
+		// 	"testing/test_results/results.txt",
+		// 	true
+		// );
+		// testing_logger->set_pattern("%v");
+		std::shared_ptr<spdlog::logger> testing_logger = spdlog::default_logger();
```
Hop311: Why is that code commented out? I get identical test outputs when using it to the `spdlog::default_logger()` version
Also what's the difference between the `spdlog::default_logger_raw()` used forr the `SinkGuard` above and the `spdlog::default_logger()` used here?

src/headless/main.cpp
```diff
@@ -174,21 +217,28 @@ static bool run_headless(fs::path const& root, memory::vector<memory::string>& m
...
+		Testing testing { game_manager.get_definition_manager(), testing_logger };
+		SPDLOG_INFO("Testing Loaded");
 		testing.execute_all_scripts();
 		testing.report_results();
-		std::cout << "Testing Executed" << std::endl << std::endl;
+		SPDLOG_INFO("Testing Executed");
```
Hop311: This message will have one fewer newlines after it compared to before

src/headless/main.cpp
```diff
@@ -243,23 +294,23 @@ static bool run_headless(fs::path const& root, memory::vector<memory::string>& m
 
 		using test_time_units_t = std::chrono::milliseconds;
 
-		Logger::info("===== Land Pathfinding test... =====");
+		SPDLOG_INFO("===== Land Pathfinding test... =====");
 		test_duration_t duration = std::chrono::duration_cast<test_time_units_t>(run_pathing_test<TESTS>(
```
Hop311: Slightly unrelated to this PR, but here and for the sea pathing tests `test_duration_t` is `nanoseconds` while `test_time_units_t` is `milliseconds`, meaning the printed value is in nanoseconds but rounded to the nearest 1,000,000 (sometimes going to zero).

src/headless/main.cpp
```diff
@@ -387,16 +438,27 @@ int main(int argc, char const* argv[]) {
...
-	std::cout << "!!! HEADLESS SIMULATION END !!!" << std::endl;
+	SPDLOG_INFO("!!! HEADLESS SIMULATION END !!!");
 
-	std::cout << "\nLoad returned: " << (ret ? "SUCCESS" : "FAILURE") << std::endl;
+	spdlog::info("Load returned: {}", ret ? "SUCCESS" : "FAILURE");
 
-	std::cout << "\nLogger Summary: Info = " << Logger::get_info_count() << ", Warning = " << Logger::get_warning_count()
-		<< ", Error = " << Logger::get_error_count() << std::endl;
+	spdlog::info(
+		"Logger Summary: "
```
Hop311: Both of these log messages will have one fewer newline before them than before

src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -2550,3 +2542,7 @@ void CountryInstanceManager::country_manager_tick_after_map(InstanceManager& ins
 
 	shared_country_values.update_costs(instance_manager.get_good_instance_manager());
 }
+
+fmt::format_context::iterator fmt::formatter<CountryInstance>::format(const CountryInstance& instance, format_context& ctx) const {
+	return formatter<string_view>::format(instance.get_identifier(), ctx);
+}
```
Hop311: As mentioned earlier, this could be removed if `OpenVic::HasGetIdentifier` was used instead of `std::is_base_of_v<OpenVic::HasIdentifier, T>` in the other `fmt::formatter`

deps/SCsub
```diff
@@ -111,6 +111,55 @@ def build_memory(env):
...
+            "SPDLOG_LEVEL_DEBUG",
+            "SPDLOG_LEVEL_INFO",
+            "SPDLOG_LEVEL_WARN",
+            "SPDLOG_LEVEL_ERROR",
+            "SPDLOG_LEVEL_CRITICAL",
+            "SPDLOG_LEVEL_OFF"
+        ]:
+            print("SPDLOG_LEVEL can only be trace, debug, info, warn, error, critical, or off")
+            exit(255)
+    elif env["dev_build"]:
```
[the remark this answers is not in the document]
Spartan322: Scons options are guaranteed to exist in the env, its only before they're assigned or if they're conditionally assigned that they may not exist.

src/headless/main.cpp
```diff
@@ -120,51 +134,80 @@ static test_duration_t run_pathing_test(AStarPathing& pathing, std::optional<pat
...
 	return end_time - start_time;
 }
 
+struct SinkGuard {
+	spdlog::logger* logger;
+	spdlog::sink_ptr sink;
+
+	SinkGuard(spdlog::logger* l, spdlog::sink_ptr s) : logger(l), sink(s) {
+		logger->sinks().push_back(sink);
+	}
```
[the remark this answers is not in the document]
Spartan322: No, logger should never be nullptr, if it was it basically result in a crash either way, same with the sink. (well technically sink could be nullptr, its worthless to check it for null either way)

src/headless/main.cpp
Spartan322: > Why is that code commented out?

Preserved code for logging to a file to replicate what the testing used to do. But that's normally worthless, but it retains the capacity to replicate the old behavior.

> Also what's the difference between the spdlog::default_logger_raw() used forr the SinkGuard above and the spdlog::default_logger() used here?

The raw is a raw pointer where as the other is a shared_ptr, aside from being safe for multithreading it also is the standard way of dealing with the loggers, spdlog's interface expects you to use the `shared_ptr<logger>` type.

src/headless/main.cpp
Spartan322: There's no reason for it to have them, it just pollutes the log, it doesn't make anything easier to read or deal with. A logger can be written in such a way that this actually throws a wrench into what it writes.

src/openvic-simulation/GameManager.cpp
```diff
@@ -169,23 +171,23 @@ bool GameManager::load_definitions(Dataloader::localisation_callback_t localisat
...
 		gamestate_updated_callback
 	);
 
-	Logger::info("Setting up new game instance.");
+	SPDLOG_INFO("Setting up new game instance.");
 
 	bool ret = instance_manager->setup();
 
-	Logger::info("Loading bookmark \"", bookmark, "\" for new game instance.");
+	SPDLOG_INFO("Loading bookmark \"{}\" for new game instance.", fmt::validate(bookmark));
```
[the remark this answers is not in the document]
Spartan322: I don't recall if fmt even supports but that's foolish regardless, its unsafe and error-prone, validate ensures we never make the mistake of permitting a pointer into fmt which will hide user expectation. This is type safe and there is guaranteed no error in dev intent, the compiler will instead produce a compile-time error for being unspecific. Type safety trumps convenience especially for something thing small.

src/headless/main.cpp
```diff
@@ -1,11 +1,18 @@
 #include <chrono>
 #include <filesystem>
+#include <iostream>
```
Hop311: I believe this can be removed, unless stuff is arranged differently on Linux or Mac

src/openvic-simulation/dataloader/Vic2PathSearch.cpp
```diff
@@ -127,34 +162,35 @@ static fs::path _search_for_game_path(fs::path hint_path = {}) {
 			hint_path = home / ".local" / "share" / "Steam";
```
Hop311: Is `fs::read_symlink` above here guarunteed to return a valid directory path when given an input satifying `fs::is_symlink`? If not should we test `hint_path` before continuing on

Hop311 (commented):
Here are some more comments

src/openvic-simulation/dataloader/NodeTools.hpp
```diff
@@ -526,9 +533,10 @@ using namespace std::string_view_literals;
 				if (it != map.end()) {
 					return callback(it->second);
 				}
-				Logger::warn_or_error(warn, "\"", string, "\" is not a valid key. Valid keys: [",
-					StringUtils::string_join(map),
-					"]"
+				spdlog::log_s(
+					warn ? spdlog::level::warn : spdlog::level::err,
```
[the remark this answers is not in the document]
Spartan322: Its really not, it shows up sparingly and its not all that more convenient. That aside it was inherently a poor implementation, we shouldn't simply be writing convenience functions just to make something slightly easier. I deliberately did not bring that thing back and replaced it with log_s because its more clear about what its doing and only shows up a few times.

src/openvic-simulation/dataloader/Vic2PathSearch.cpp
```diff
@@ -111,8 +146,8 @@ static fs::path _search_for_game_path(fs::path hint_path = {}) {
 			static const fs::path prog_files_x86 = memory::string(std::getenv("ProgramFiles(x86)"));
 			hint_path = prog_files_x86 / "Steam";
```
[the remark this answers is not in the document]
Spartan322: Completely irrelevant, please stop nitpicking things that aren't applicable to the PR. If you have an issue please submit a PR or discuss it elsewhere. It wastes development time dealing with nitpicks that are completely irrelevant to the PR.

src/openvic-simulation/dataloader/Vic2PathSearch.cpp
```diff
@@ -201,16 +237,16 @@ static fs::path _search_for_game_path(fs::path hint_path = {}) {
 		if (!parser.parse()) {
 			// Could not find or load libraryfolders.vdf, report error as warning
 			if (!buffer.empty()) {
-				return empty_fail_result_callback(buffer);
+				return search_log(hint_empty, "{}", buffer).result();
```
Hop311: Could you just use `buffer` as the format string directly? Or is there a risk of stuff breaking e.g. if `buffer` contains instances of `{}`?

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -198,20 +205,26 @@ bool MapDefinition::add_special_adjacency(
 	case LAND:
 	case STRAIT:
 		if (from.is_water() || to.is_water()) {
-			Logger::error(adjacency_t::get_type_name(type), " adjacency from ", from, " to ", to, " has water endpoint(s)!");
+			spdlog::error_s(
+				"{} adjacency from {} to {}",
```
Hop311:
```suggestion
				"{} adjacency from {} to {} has water endpoint(s)!",
```
Missing part of the original message

src/openvic-simulation/map/MapDefinition.cpp
```diff
@@ -530,18 +550,19 @@ bool MapDefinition::load_province_definitions(std::span<const LineObject> lines)
...
+		spdlog::error_s("No header or entries in province definition file!");
 		ret = false;
 	} else {
 		LineObject const& header = lines.front();
 		if (!_validate_province_definitions_header(header)) {
-			Logger::error(
-				"Non-standard province definition file header - make sure this is not a province definition: ", header
+			spdlog::error_s(
+				"Non-standard province definition file header - make sure this is not a province definition: {}",
+				fmt::streamed(header) // TODO: remove streamed
```
Hop311: I think you can remove `fmt::streamed` from here and the other one below, as the SIM's OpenVic-Dataloader submodule has been updated to include the `fmt::formatter<ovdl::csv::LineObject>` specialisation.

src/openvic-simulation/types/HasIdentifier.hpp
```diff
@@ -104,3 +108,29 @@ namespace OpenVic {
 	template<typename T>
 	concept HasGetIdentifierAndGetColour = HasGetIdentifier<T> && HasGetColour<T>;
 }
+
+template<typename T>
+struct fmt::formatter<T, std::enable_if_t<std::is_base_of_v<OpenVic::HasIdentifier, T>, char>> : fmt::formatter<string_view> {
+	auto format(const T& has_id, format_context& ctx) const {
+		return formatter<string_view>::format(has_id.get_identifier(), ctx);
+	}
+};
```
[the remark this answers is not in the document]
wvpm: I like the idea. Identifier should have precedence over name, so you can drop the `requires (!OpenVic::HasGetName<T>)`.
That would first try the identifier, then name and then fall back to default.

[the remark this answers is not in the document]
Spartan322: This is still invalid for State because its does not validate HasGetIdentifier thanks to the fact it returns a string and not a string_view.

Hop311 (commented):
Hopefully these are still useful, dunno when I'll be able to finish the review

wvpm (approved):
It still LGTM

wvpm (approved):
approved again

PR #575, opened 2025-10-14
Add RingBuffer datatype

Spartan322:
- Partially supersedes #574 
- Replaces the RingBuffer in #428 

Should mostly behave as a drop-in replacement for `std::vector`, if you need stack allocation, use `foonathan::memory`.


wvpm (approved):
As discussed in the discord, we can use this as a base and add a wrapper type that restricts the direction of input, prevents reallocation and supports filling.

PR #581, opened 2025-10-17
Fixup `GameManager::load_mods`

Spartan322:
Rename `requested_mods` parameter to `mods_to_find`
Retype `mods_to_find` as `memory::vector const&`
Remove `roots` and `replace_paths` parameters

This reduces the capacity for misuse and cleans up the interface.

### DIFF
--- a/src/openvic-simulation/GameManager.cpp
+++ b/src/openvic-simulation/GameManager.cpp
@@ -1,11 +1,14 @@
 #include "GameManager.hpp"
 
 #include <chrono>
-
 #include <cstddef>
 #include <string_view>
 
+#include <range/v3/algorithm/contains.hpp>
+#include <range/v3/algorithm/find_if.hpp>
+
 #include "openvic-simulation/dataloader/Dataloader.hpp"
+#include "openvic-simulation/types/OrderedContainers.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
 
 using namespace OpenVic;
@@ -65,12 +68,8 @@ bool GameManager::load_mod_descriptors() {
 	return true;
 }
 
-bool GameManager::load_mods(
-	Dataloader::path_vector_t& roots,
-	Dataloader::path_vector_t& replace_paths,
-	utility::forwardable_span<const memory::string> requested_mods
-) {
-	if (requested_mods.empty()) {
+bool GameManager::load_mods(memory::vector<memory::string> const& mods_to_find) {
+	if (mods_to_find.empty()) {
 		return true;
 	}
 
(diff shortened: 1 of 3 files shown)


src/openvic-simulation/GameManager.hpp
```diff
@@ -55,11 +55,7 @@ namespace OpenVic {
 
 		bool load_mod_descriptors();
 
-		bool load_mods(
-			Dataloader::path_vector_t& roots,
-			Dataloader::path_vector_t& replace_paths,
-			utility::forwardable_span<const memory::string> requested_mods
-		);
+		bool load_mods(memory::vector<memory::string> const& mods_to_find);
```
Hop311: Why not leave this as a forwardable span? Is it due to the new find function use?

src/openvic-simulation/GameManager.cpp
```diff
@@ -118,27 +115,28 @@ bool GameManager::load_mods(
 		}
 	}
 
+	Dataloader::path_vector_t roots = dataloader.get_roots();
+	roots.reserve(load_list.size() + roots.size());
```
Hop311: This could be `reserve_more(roots, load_list.size())`, although it's not an important distinction. Also would it be worth making the `path_vector_t` alias use a `memory::vector`? Maybe even `memory::strirng` instead of `fs::path` if that's feasible and has advantages, or defining a `memory::path` struct as we use them very often during load

src/openvic-simulation/GameManager.cpp
```diff
@@ -118,27 +115,28 @@ bool GameManager::load_mods(
...
+
+	vector_ordered_set<fs::path> replace_paths;
```
Hop311: In the vector set template and other vector based tsl stuff, could we use `memory::vector` or at least `foonathan::memory::vector` instead of the base `std::vector`? I expect that would cover a lot of stuff not under the memory tracker yet, including all indexed flat maps/registries if I'm not mistaken

src/openvic-simulation/GameManager.hpp
Spartan322: forwardable_span foremost should only be for forward declared types, string can have that. That aside with how we use it, it still requires conversion anyway because it be a span of memory::string, so I decided might as well just make it convenient to write for a one-off so its done to make
```cpp
manager.load_mods({"HPM"});
```
possible.

src/openvic-simulation/GameManager.cpp
Spartan322: Its not possible to control the allocation of filesystem::path, also path_vector_t is already a memory::vector. Also std::vector doesn't have a reserve_more function. memory::vector is just a std::vector with a specialized allocator.

src/openvic-simulation/GameManager.cpp
Spartan322: That's what path_vector_t is.

PR #595, opened 2025-10-26
Reorganize generic concepts to `utility/Concepts.hpp`

Spartan322:
Standardize concept names as being snake_case

Since concepts are not types and should not be treated like types and can't be mistakenly used as variables, I decided to standardize the naming scheme of concepts to be snake_case, as well removed all `Is`/`is_` prefixes as they should be treated akin to types, the is prefixes should be reserved for type trait queries, which generally should only be defined to enable concepts. (as you cannot specialize a concept)


src/openvic-simulation/utility/Concepts.hpp
```diff
@@ -0,0 +1,203 @@
...
+	template<typename T>
+	concept either_char_type = any_of<T, char, wchar_t>;
+
+	template<typename T>
+	concept has_data = requires(T t) {
+		{ t.data() } -> std::convertible_to<typename T::value_type const*>;
+	};
+
+	template<typename T>
+	concept has_get_identifier = requires(T const& t) {
```
wvpm: Concepts ending with -able sound more natural to me.
What do you think of `identifiable`?

wvpm (approved):

PR #596, opened 2025-10-27
Expand CountryInstance's ADD_AND_REMOVE macros

Spartan322:
Remove has functions (already handled via PROPERTY getter)

### DIFF
--- a/src/openvic-simulation/country/CountryInstance.cpp
+++ b/src/openvic-simulation/country/CountryInstance.cpp
@@ -484,38 +484,85 @@ ReadOnlyClampedValue const& CountryInstance::get_tax_rate_slider_value_by_strata
 	return tax_rate_slider_value_by_strata.at(strata);
 }
 
-#define ADD_AND_REMOVE(item) \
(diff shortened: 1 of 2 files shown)


src/openvic-simulation/country/CountryInstance.cpp
```diff
@@ -484,38 +484,105 @@ ReadOnlyClampedValue const& CountryInstance::get_tax_rate_slider_value_by_strata
...
+bool CountryInstance::remove_owned_province(ProvinceInstance const& item_to_remove) {
+	OV_ERR_FAIL_COND_V_MSG(
+		owned_provinces.erase(&item_to_remove) == 0, false,
+		memory::fmt::format("owned_province \"{}\" not present in country {}", item_to_remove, *this)
+	);
+	return true;
+}
+
+bool CountryInstance::has_owned_province(ProvinceInstance const& owned_province) const {
+	return owned_provinces.contains(&owned_province);
```
schombert: presumably it is much faster to just check owned_province.owner == this, no?

PR #599, opened 2025-10-27
Remove influence_value_type macros

Spartan322:
Fix influence_value_type::value potentially being uninitialized


wvpm (approved):
I thought the macros were a clever solution to avoid code duplication but if you don't like them anymore, go ahead and expand the macros.

Note expanding the macros is unrelated to "Fix influence_value_type::value potentially being uninitialized" but I understand packing the changes is easier.

PR #609, opened 2025-10-31
Remove overhead of TLS on river point detection

xinitrcn1:
- `thread_local` not recommended, stack was preferred
- Cleans up the loop logic by chaining else/if (no codegen change, just a stylistic choice)
- Stylistic choice of making the loop more clearly bounded


xinitrcn1:
Done.

PR #672, opened 2025-11-27
Never break after operator

wvpm:
Never break after operators.
Operators are the most important signal of what the code does.
They should always be in sight and never at the end of a line like an afterthought.


zaaarf:
I agree with this, my only complaint is that this PR seems to do a number of things, not all of whom related to the stated task (ex. whatever is going on in main.cpp). Is there a reason it's like that?

wvpm:

PR #693, opened 2026-01-21
Move generic containers into core/containers

Spartan322:
- Incremental refactor to achieve #625 

Move tracked allocating containers to memory namespace

`HasIdentifier.hpp` will be split into distinct class headers after merge following #625, done to focus on changing the namespace and headers and minimize refactoring header inclusions in this PR.


wvpm (dismissed):
Prefixing commonly used types with `memory::` hurts code.
If you want to bring your own allocator, you'd be better off using `_detail::T<MyAllocator>`

wvpm:
We could revive this. I am strongly in favour of splitting up Containers.hpp as it's leaking imports everywhere.
I still oppose prefixing commonly used types with `memory::`, especially if there is a variant without that prefix.

wvpm:
Partially superseded by #731

PR #698, opened 2026-02-15
Condition Evaluation

DarkiBoi:


wvpm (changes requested):
The condition ids are not related to the types (province, pop, country, etc).
Please move them outside. The types are already massive and don't need tight coupling with conditions.
A condition can simply receive an instance and perform its checks from the outside.
All data required for conditions is exposed via public const fields or methods.

src/openvic-simulation/country/CountryInstance.hpp
```diff
@@ -77,6 +77,7 @@ namespace OpenVic {
 	struct TechnologySchool;
 	struct UnitInstanceGroup;
 	struct UnitTypeManager;
+	struct ConditionNode;
```
wvpm:
```suggestion
```
No longer needed

src/openvic-simulation/map/State.hpp
```diff
@@ -25,6 +25,7 @@ namespace OpenVic {
 	struct StateManager;
 	struct StateSet;
 	struct Strata;
+	struct ConditionNode;
```
wvpm:
```suggestion
```

src/openvic-simulation/population/Pop.cpp
```diff
@@ -42,6 +42,7 @@
 #include "openvic-simulation/types/TypedIndices.hpp"
 #include "openvic-simulation/utility/Containers.hpp"
 #include "openvic-simulation/utility/Logger.hpp"
+#include "openvic-simulation/core/Typedefs.hpp"
```
wvpm:
```suggestion
```
Not related to this PR

src/openvic-simulation/population/Pop.hpp
```diff
@@ -35,6 +35,7 @@ namespace OpenVic {
 	struct RebelType;
 	struct Religion;
 	struct SellResult;
+	struct ConditionNode;
```
wvpm:
```suggestion
```

src/openvic-simulation/scripts/Context.hpp
```diff
@@ -0,0 +1,65 @@
...
+		Context(State const* p, DefinitionManager const& dm, InstanceManager const& im)
+			: ptr(p), definition_manager(dm), instance_manager(im), this_scope(this) {}
+		Context(Pop const* p, DefinitionManager const& dm, InstanceManager const& im)
+			: ptr(p), definition_manager(dm), instance_manager(im), this_scope(this) {}
+
+		Context(
+			auto* p,
+			DefinitionManager const& dm,
+			InstanceManager const& im,
+			Context const* this_ctx,
```
wvpm: The naming convention inside this project is `new_[field_name]` for constructor arguments.
That also avoids a distinction between scope and context.

src/openvic-simulation/scripts/Context.hpp
```diff
@@ -0,0 +1,65 @@
...
+			Pop const*> ptr;
+
+		DefinitionManager const& definition_manager;
+		InstanceManager const& instance_manager;
+
+		Context const* this_scope = nullptr;
+		Context const* from_scope = nullptr;
+
+		Context(CountryInstance const* p, DefinitionManager const& dm, InstanceManager const& im)
+			: ptr(p), definition_manager(dm), instance_manager(im), this_scope(this) {}
```
wvpm: When using constructor overloads, forward the arguments to the main constructor (the line 39 one I think).
This ensures all constructors function the same and the overloads are merely for convenience.

src/openvic-simulation/scripts/Context.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace OpenVic {
+	struct CountryInstance;
+	struct ProvinceInstance;
+	struct State;
+	struct Pop;
+	struct DefinitionManager;
+	struct InstanceManager;
+
+	struct Context {
```
wvpm, line 24: The default access modifier of a struct is `public`.
In this project we avoid mutable publci fields.

You could make the fields private and expose their getters via the PROPERTY macro.
If the fields are only set in the constructor, you have to write `Context const* const`.
The `const*` makes it a pointer to a readonly Context. The second `const` makes the pointer itself readonly.

`DefinitionManager const& definition_manager;` & `InstanceManager const& instance_manager;` can safely be exposed as public fields as references are readonly.

src/openvic-simulation/scripts/Context.hpp
```diff
@@ -0,0 +1,65 @@
...
+		Context(CountryInstance const* p, DefinitionManager const& dm, InstanceManager const& im)
+			: ptr(p), definition_manager(dm), instance_manager(im), this_scope(this) {}
+		Context(ProvinceInstance const* p, DefinitionManager const& dm, InstanceManager const& im)
+			: ptr(p), definition_manager(dm), instance_manager(im), this_scope(this) {}
+		Context(State const* p, DefinitionManager const& dm, InstanceManager const& im)
+			: ptr(p), definition_manager(dm), instance_manager(im), this_scope(this) {}
+		Context(Pop const* p, DefinitionManager const& dm, InstanceManager const& im)
+			: ptr(p), definition_manager(dm), instance_manager(im), this_scope(this) {}
+
+		Context(
```
wvpm, line 40: Consider making this constructor private unless required otherwise

src/openvic-simulation/scripts/Context.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+#include <optional>
+#include <variant>
+#include <vector>
```
wvpm: Spartan322 added memory tracking. This requires using `memory::vector` instead of `std::vector`.
See also https://github.com/OpenVicProject/OpenVic-Simulation/pull/421
So use `#include "openvic-simulation/utility/Containers.hpp"` instead.

src/openvic-simulation/scripts/Context.hpp
```diff
@@ -0,0 +1,65 @@
...
+		std::variant<
+			CountryInstance const*,
+			ProvinceInstance const*,
+			State const*,
+			Pop const*> ptr;
+
+		DefinitionManager const& definition_manager;
+		InstanceManager const& instance_manager;
+
+		Context const* this_scope = nullptr;
```
wvpm: Can THIS ever be null?
How is accessing `context.this_scope` different from just using `context`?

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+#include "openvic-simulation/country/CountryInstance.hpp"
+#include "openvic-simulation/map/ProvinceInstance.hpp"
+#include "openvic-simulation/map/State.hpp"
+#include "openvic-simulation/population/Pop.hpp"
+#include "openvic-simulation/DefinitionManager.hpp"
+#include "openvic-simulation/InstanceManager.hpp"
+
+using namespace OpenVic;
+scope_type_t Context::get_scope_type() const {
```
wvpm: Why not determine this during construction and store it as a field?

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,131 @@
...
+		using T = std::decay_t<decltype(p)>;
+
+		if constexpr (std::is_same_v<T, CountryInstance const*>) {
+			return p->get_identifier();
+		}
+		else if constexpr (std::is_same_v<T, ProvinceInstance const*>) {
+			return p->get_identifier();
+		}
+		else if constexpr (std::is_same_v<T, State const*>) {
+			return p->get_identifier();
```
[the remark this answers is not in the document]
wvpm: @DarkiBoi  A state doesn't even have an identifier.
Consider skipping states for now.
I think we'll use Regions later on either as scope or merely for the identifier.

Terminology:
Region Victoria 2\map\regions.txt, for example Texas.
State a part of a region owned by a country with a colonial status. For example American colonial Texas.

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+		}
+		else if constexpr (std::is_same_v<T, Pop const*>) {
+			return "";
```
wvpm, line 16: For debug purposes, I suggest you at least write the pop type here.
Alternatively use `pop.id_in_province` & `pop.get_location()->index`.

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+	return std::visit(
+		[&](auto const* p) -> bool {
+			if (!p) {
+				return false;
+			}
+
+			using T = std::decay_t<decltype(*p)>;
+
+			if constexpr (std::is_same_v<T, CountryInstance>) {
+				// TODO: https://vic2.paradoxwikis.com/List_of_conditions#Country_Scope
```
wvpm: Please move the body of these if statements to separate methods.
```cpp
if constexpr (std::is_same_v<T, CountryInstance>) {
evaluate_country_leaf(node, *p);
}
```

```cpp
bool evaluate_country_leaf(ConditionNode const& node, CountryInstance const& country) const;
```

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+		}
+
+		return "";
+	}, ptr);
+}
+
+bool Context::evaluate_leaf(ConditionNode const& node) const {
+	std::string_view const id = node.get_condition()->get_identifier();
+
+	return std::visit(
```
wvpm, line 42: This isn't how `std::visit` is designed. See https://en.cppreference.com/w/cpp/utility/variant/visit
Either use `if constexpr (std::is_same_v<T, CountryInstance>)` or write overloads.

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+	else if (std::holds_alternative<ProvinceInstance const*>(ptr)) {
+		ProvinceInstance const* province = std::get<ProvinceInstance const*>(ptr);
+
+		// TODO
+	}
+
+	return result;
+}
+
+std::optional<Context> Context::get_redirect_context(std::string_view condition_id, scope_type_t target) const {
```
wvpm, line 227: `std::optional<Context>` owns the Context.
Both this_scope and from_scope already exist and are owned somewhere else.
Returning them here copies them. I don't think that's what you want.

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+		}
+		return std::nullopt;
+	}
+
+	if (std::holds_alternative<ProvinceInstance const*>(ptr)) {
+		ProvinceInstance const* province = std::get<ProvinceInstance const*>(ptr);
+
+		if (target == scope_type_t::COUNTRY) {
+			if (condition_id == "owner") {
+				auto const* owner = province->get_owner();
```
wvpm, line 247: In this project explicit types are preferred over auto types.
Just use `CountryInstance const* owner` here and for controller.

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+				if (id == "war_exhaustion") {
+					fixed_point_t expected = std::get<fixed_point_t>(node.get_value());
+					return p->get_war_exhaustion() >= expected;
+				}
+
+				spdlog::warn_s("Condition {} not implemented for Country scope", id);
+				return false;
+			}
+			else if constexpr (std::is_same_v<T, State>) {
+				// No state conditions according to wiki?
```
wvpm: See Victoria 2\common\production_types.txt
`trade_goods_in_state = coal`
`has_building = electric_gear_factory`

src/openvic-simulation/scripts/Context.cpp
```diff
@@ -0,0 +1,278 @@
...
+					fixed_point_t expected = std::get<fixed_point_t>(node.get_value());
+					return p->get_average_consciousness() >= expected;
+				}
+				if (id == "average_militancy") {
+					fixed_point_t expected = std::get<fixed_point_t>(node.get_value());
+					return p->get_average_militancy() >= expected;
+				}
+				if (id == "badboy") {
+					fixed_point_t expected_ratio = std::get<fixed_point_t>(node.get_value());
+					return p->get_infamy_untracked() >= expected_ratio * definition_manager.get_define_manager().get_country_defines().get_infamy_containment_limit();
```
wvpm: If you only need this define, you should request the CountryDefines in your constructor instead of searching for it yourself. Same for the CountryInstanceManager.

We can leave this as is and trim it later once other conditions are implemented.

DarkiBoi: Yeah i'd rather cast the net wide and make it easier to implement the other conditions, we can trim once thats done

PR #706, opened 2026-02-20
Refactor const properties to fields

wvpm:
Only use PROPERTY when the data must be publicly exposed and either may change or requires a chance in type (string -> string_view, vector -> span, etc).
For constant data, use const fields.
When a pointer must have a value, use a reference if possible, otherwise use `type_safe::object_ref` (follow up PR coming for that).


src/openvic-simulation/history/DiplomaticHistory.hpp
```diff
@@ -122,19 +20,41 @@ namespace OpenVic {
...
-
-		memory::vector<AllianceHistory const*> get_alliances(Date date) const;
-		memory::vector<ReparationsHistory const*> get_reparations(Date date) const;
-		memory::vector<SubjectHistory const*> get_subjects(Date date) const;
+		[[nodiscard]] constexpr bool is_locked() const {
+			return locked;
+		}
+
+		[[nodiscard]] memory::vector<AllianceHistory const*> get_alliances(Date date) const {
+			return filter_by_date<AllianceHistory>({ alliances }, date);
```
Spartan322:
```suggestion
			return filter_by_date<AllianceHistory>(alliances, date);
```

src/openvic-simulation/dataloader/ModManager.hpp
```diff
@@ -13,8 +13,11 @@ namespace OpenVic {
 	struct Mod : HasIdentifier {
 	private:
 		ModManager const& mod_manager;
+
+		//properties despite being const for string_view.
```
Spartan322:
```suggestion
		// properties despite being const for string_view.
```

src/openvic-simulation/utility/Getters.hpp
```diff
@@ -3,13 +3,16 @@
...
+#include <optional>
 #include <string>
 #include <string_view>
 #include <type_traits>
 #include <utility>
 
+#include "openvic-simulation/core/portable/ForwardableSpan.hpp" // IWYU pragma: keep for SPAN_PROPERTY
 #include "openvic-simulation/core/Typedefs.hpp" // IWYU pragma: keep
 #include "openvic-simulation/utility/Containers.hpp"
+#include "core/template/Concepts.hpp"
```
Spartan322:
```suggestion
#include "openvic-simulation/core/template/Concepts.hpp"
```

Spartan322 (approved):

PR #707, opened 2026-02-23
std::reference_wrapper for ModManager

wvpm:
`T*` is used throughout our codebase to store pointers to objects whenever `T&` isn't viable due to move semantics.
For example `vector_ordered_set<Mod const*>`. Technically these pointers could be null and to be safe you'd have to check it isn't null before dereferencing.

Often we know the pointer may never be null. `std::reference_wrapper<T>` is a great way to communicate that. It is implicitly convertible to `T&`. This allows us to do `for (Mod const& mod : loaded_mods)` (with `memory::vector<std::reference_wrapper<const Mod>> loaded_mods`).

This approach does require the type is hashable and equatable. For types derived from HasIndex, the index is used for this. For types that don't derive from HasIndex, but do derive from HasIdentifier, the identifier is used.

### DIFF
--- a/src/openvic-simulation/GameManager.cpp
+++ b/src/openvic-simulation/GameManager.cpp
@@ -1,6 +1,7 @@
 #include "GameManager.hpp"
 
 #include <chrono>
+#include <functional>
 #include <string_view>
 
 #include <range/v3/algorithm/contains.hpp>
@@ -76,7 +77,7 @@ bool GameManager::load_mods(memory::vector<memory::string> const& mods_to_find) 
 
 	bool ret = true;
 
-	vector_ordered_set<Mod const*> load_list;
+	vector_ordered_set<std::reference_wrapper<const Mod>> load_list;
 
 	/* Check loaded mod descriptors for requested mods, using either full name or user directory name
 	 * (Historical Project Mod 0.4.6 or HPM both valid, for example), and load them plus their dependencies.
@@ -94,8 +95,8 @@ bool GameManager::load_mods(memory::vector<memory::string> const& mods_to_find) 
 			continue;
 		}
 
-		Mod const* mod_ptr = &*it;
-		vector_ordered_set<Mod const*> dependencies = mod_ptr->generate_dependency_list(&ret);
+		Mod const& mod = *it;
+		vector_ordered_set<std::reference_wrapper<const Mod>> dependencies = mod.generate_dependency_list(&ret);
 		if(!ret) {
 			continue;
 		}
@@ -104,15 +105,15 @@ bool GameManager::load_mods(memory::vector<memory::string> const& mods_to_find) 
 		if (load_list.empty()) {
 			load_list = std::move(dependencies);
 		} else {
-			for (Mod const* dep : dependencies) {
+			for (Mod const& dep : dependencies) {
 				if (!load_list.contains(dep)) {
 					load_list.emplace(dep);
 				}
 			}
 		}
 
-		if (!load_list.contains(mod_ptr)) {
-			load_list.emplace(mod_ptr);
+		if (!load_list.contains(mod)) {
+			load_list.emplace(mod);
 		}
 	}
 
@@ -122,9 +123,9 @@ bool GameManager::load_mods(memory::vector<memory::string> const& mods_to_find) 
 	vector_ordered_set<fs::path> replace_paths;
 
 	/* Actually registers all roots and replace paths to be loaded by the game. */
-	for (Mod const* mod : load_list) {
-		roots.emplace_back(roots[0] / mod->get_dataloader_root_path());
-		for (std::string_view path : mod->get_replace_paths()) {
+	for (Mod const& mod : load_list) {
+		roots.emplace_back(roots[0] / mod.get_dataloader_root_path());
+		for (std::string_view path : mod.get_replace_paths()) {
 			replace_paths.emplace(path);
 		}
 	}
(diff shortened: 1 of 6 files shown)


src/openvic-simulation/types/HasIndex.hpp
```diff
@@ -26,4 +28,14 @@ namespace OpenVic {
 			return index == rhs.index;
 		}
 	};
+}
+
+namespace std {
+	template<typename T>
+	requires OpenVic::has_index<T>
```
Spartan322:
```suggestion
	template<OpenVic::has_index T>
```

src/openvic-simulation/types/HasIdentifier.hpp
```diff
@@ -95,4 +101,14 @@ struct fmt::formatter<T> : fmt::formatter<fmt::string_view> {
 	fmt::format_context::iterator format(T const& has_id, fmt::format_context& ctx) const {
 		return fmt::formatter<fmt::string_view>::format(has_id.get_name(), ctx);
 	}
-};
\ No newline at end of file
+};
+
+namespace std {
+	template<typename T>
+	requires OpenVic::has_get_identifier<T> && (!OpenVic::has_index<T>)
```
Spartan322:
```suggestion
	template<OpenVic::has_get_identifier T>
	requires (!OpenVic::has_index<T>)
```

src/openvic-simulation/types/HasIdentifier.hpp
```diff
@@ -30,6 +30,12 @@ namespace OpenVic {
 		HasIdentifier(HasIdentifier&&) = default;
 		HasIdentifier& operator=(HasIdentifier const&) = delete;
 		HasIdentifier& operator=(HasIdentifier&&) = delete;
+
+		template <typename T>
+        requires std::derived_from<T, HasIdentifier> && (!has_index<T>)
```
Spartan322:
```suggestion
		template <std::derived_from<HasIdentifier> T>
		requires (!has_index<T>)
```

Also check your editor settings, this is mixing spaces and tabs.

src/openvic-simulation/types/HasIdentifier.hpp
```diff
@@ -30,6 +30,12 @@ namespace OpenVic {
...
+        friend bool operator==(T const& lhs, T const& rhs) {
+            return lhs.get_identifier() == rhs.get_identifier();
+        }
```
Spartan322:
```suggestion
		friend bool operator==(T const& lhs, T const& rhs) {
			return lhs.get_identifier() == rhs.get_identifier();
		}
```

src/openvic-simulation/types/OrderedContainers.hpp
```diff
@@ -14,6 +16,19 @@
 #include <foonathan/memory/default_allocator.hpp>
 #include <foonathan/memory/std_allocator.hpp>
 
+//pre c++ 26
```
Spartan322: There's no reason to mark this with a comment, there's plenty of other cases where errors and warnings would show up that aren't marked if we switched, this would become an error in C++26 as well.
```suggestion
```

src/openvic-simulation/GameManager.cpp
```diff
@@ -94,8 +95,8 @@ bool GameManager::load_mods(memory::vector<memory::string> const& mods_to_find)
 			continue;
 		}
 
-		Mod const* mod_ptr = &*it;
-		vector_ordered_set<Mod const*> dependencies = mod_ptr->generate_dependency_list(&ret);
+		const Mod& mod = *it;
```
Spartan322: Aren't we supposed to be doing `Mod const&`, (despite the fact that it is deceptively frustrating imo, the const has nothing to do with the reference or pointer, its the object contained in it) every time I brought up the issue with this style I was opposed.

src/openvic-simulation/dataloader/ModManager.cpp
```diff
@@ -27,42 +27,48 @@ Mod::Mod(
...
 				);
 				return false;
 			} else {
 				current_recurse++;
 				ret &= self(self, dep, dep_list); /* recursively search for mod dependencies */
 			}
+
 			if (!dep_list.contains(dep)) {
-				dep_list.emplace(dep);
+				dep_list.insert(dep);
```
Spartan322: Why not use emplace?

src/openvic-simulation/GameManager.cpp
wvpm: I should've made it `Mod const&`. This was an oversight.
If you want to change the style, I'm open to that. This wasn't an attempt to change the style.

src/openvic-simulation/types/OrderedContainers.hpp
wvpm: Okay. I thought it would make it clear we can remove it once we upgrade to c++ 26 but I'll remove it then.

src/openvic-simulation/types/HasIdentifier.hpp
wvpm: Stuff with spaces is copied from Gemini output.
I'll fix it.

src/openvic-simulation/dataloader/ModManager.cpp
wvpm: Good catch, it's a leftover from trying to fix build errors before I realised HasIdentifier required an explicit == operator. I reverted it.

Spartan322 (approved):

PR #708, opened 2026-02-24
Refactor pop to use reference_wrapper

wvpm:
A pop always has a location. Pops must be movable, thus storing `ProvinceInstance&` isn't an option.
`std::reference_wrapper<ProvinceInstance>` keeps pops movable whilst avoiding redundant null checks.
Also adds support for `std::reference_wrapper<ProvinceInstance> PROPERTY(location);`
It generates `ProvinceInstance const& get_location() const`

### DIFF
--- a/src/openvic-simulation/population/Pop.hpp
+++ b/src/openvic-simulation/population/Pop.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstddef>
+#include <functional>
 
 #include "openvic-simulation/core/portable/ForwardableSpan.hpp"
 #include "openvic-simulation/economy/production/ArtisanalProducer.hpp"
@@ -40,7 +41,7 @@ namespace OpenVic {
 		friend PopManager;
 
 	protected:
-		PopType const* PROPERTY_ACCESS(type, protected);
+		std::reference_wrapper<const PopType> PROPERTY_ACCESS(type, protected);
 		pop_size_t PROPERTY_ACCESS(size, protected);
 		fixed_point_t PROPERTY_RW_ACCESS(militancy, protected);
 		fixed_point_t PROPERTY_RW_ACCESS(consciousness, protected);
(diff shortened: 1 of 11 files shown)


src/openvic-simulation/population/Pop.hpp
```diff
@@ -99,7 +98,12 @@ namespace OpenVic {
 		fixed_point_t cash_allocated_for_artisanal_spending = 0;
 		pop_size_t employed = 0;
 
-		ProvinceInstance* PROPERTY_PTR(location, nullptr);
+		std::reference_wrapper<ProvinceInstance> PROPERTY(location);
```
wvpm: Suggested change
      
    
    
      
          
            
            		std::reference_wrapper<ProvinceInstance> PROPERTY(location);
          
          
            
            		const std::reference_wrapper<ProvinceInstance> PROPERTY(location);
          
      
    
    
  

@Spartan322 We never reassign the location. The only reason to not be const is so it's movable in the plf::colony.
What do you think, const or not?

(pop.type may seem similar but is changed farmer <> labourer, although this could be refactored into creating entirely new pops)

Spartan322: I guess we could have it const I'm just not sure about it though.

src/openvic-simulation/population/Pop.hpp
```diff
@@ -99,7 +98,12 @@ namespace OpenVic {
 		fixed_point_t cash_allocated_for_artisanal_spending = 0;
 		pop_size_t employed = 0;
 
-		ProvinceInstance* PROPERTY_PTR(location, nullptr);
+		const std::reference_wrapper<ProvinceInstance> PROPERTY(location);
+	public:
+		[[nodiscard]] constexpr ProvinceInstance& get_location() {
+			return location;
+		}
+	private:
```
Spartan322, line 108:
```suggestion
		}

	private:
```

src/openvic-simulation/population/Pop.hpp
```diff
@@ -99,7 +98,12 @@ namespace OpenVic {
 		fixed_point_t cash_allocated_for_artisanal_spending = 0;
 		pop_size_t employed = 0;
 
-		ProvinceInstance* PROPERTY_PTR(location, nullptr);
+		const std::reference_wrapper<ProvinceInstance> PROPERTY(location);
+	public:
```
Spartan322, line 103:
```suggestion
		const std::reference_wrapper<ProvinceInstance> PROPERTY(location);

	public:
```

Spartan322 (approved):

PR #709, opened 2026-02-24
Refactor region to use reference_wrapper

wvpm:
Similar to #707 & #708 but for Region & State.

PR #710, opened 2026-02-24
Refactor military to use reference_wrapper

wvpm:
Similar to #707 , #708 & #709

### DIFF
--- a/src/openvic-simulation/military/UnitInstanceGroup.cpp
+++ b/src/openvic-simulation/military/UnitInstanceGroup.cpp
@@ -2,6 +2,8 @@
 
 #include <vector>
 
+#include <fmt/std.h>
+
 #include "openvic-simulation/country/CountryInstance.hpp"
 #include "openvic-simulation/defines/MilitaryDefines.hpp"
 #include "openvic-simulation/map/MapInstance.hpp"
@@ -12,7 +14,6 @@
 #include "openvic-simulation/population/PopType.hpp"
 #include "openvic-simulation/types/OrderedContainersMath.hpp"
 #include "openvic-simulation/utility/Containers.hpp"
-#include "openvic-simulation/core/FormatValidate.hpp"
 
 using namespace OpenVic;
 
@@ -21,10 +22,16 @@ using enum unit_branch_t;
 UnitInstanceGroup::UnitInstanceGroup(
 	unique_id_t new_unique_id,
 	unit_branch_t new_branch,
-	std::string_view new_name
+	std::string_view new_name,
+	CountryInstance& new_country,
+	ProvinceInstance& new_location
 ) : unique_id { new_unique_id },
 	branch { new_branch },
-	name { new_name } {}
+	name { new_name },
+	country { new_country },
+	location { new_location } {
+		new_country.add_unit_instance_group(*this);
+	}
 
 void UnitInstanceGroup::update_gamestate() {
 	total_organisation = 0;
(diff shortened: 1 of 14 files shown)


src/openvic-simulation/military/UnitInstanceGroup.hpp
```diff
@@ -196,7 +201,8 @@ namespace OpenVic {
 		bool generate_unit_instance_group(
 			MapInstance& map_instance, CountryInstance& country, UnitDeploymentGroup<Branch> const& unit_deployment_group
 		);
-		void generate_leader(CountryInstance& country, LeaderBase const& leader);
```
wvpm, line 199: @Spartan322 both `LeaderBase const&` and rvalues are passed to this.
Wouldn't the old code create dangling references?

src/openvic-simulation/military/UnitInstanceGroup.cpp
```diff
@@ -414,20 +408,26 @@ bool UnitInstanceManager::generate_unit_instance_group(
...
 	if (leader_instance.get_picture().empty() && country.get_primary_culture() != nullptr) {
 		leader_instance.set_picture(culture_manager.get_leader_picture_name(
-			country.get_primary_culture()->group.get_leader(), leader.branch
+			country.get_primary_culture()->group.get_leader(), leader_instance.branch
 		));
 	}
 }
 
+template void UnitInstanceManager::generate_leader(CountryInstance&, LeaderBase const&);
+template void UnitInstanceManager::generate_leader(CountryInstance&, LeaderBase&&);
```
Spartan322: This provides no advantage unless the header also provides an extern template for them.

src/openvic-simulation/military/UnitInstanceGroup.hpp
Spartan322: const value references will extend the lifetime, so long as the value was valid when passed, any temporary value passed is promoted to an lvalue whose lifetime is extended until the const value reference goes out of scope. Only const value references are capable of doing this, regular references will not do this. This is the entire reason why const value references can accept value semantics and you don't need to specify say `LeaderBase` or `LeaderBase&&` (granted the former wouldn't work here, but its hypothetical) for it to not be a build error. Regular references cannot accept temporary values nor other forms of value semantics.

This aside the scope of a temporary value only ends at the end of its expression, if you make a call to a function it will live until the end of the function call's expression, (granted if you used a reference for a temporary that's a build error anyway) the usual case where dangling references come in is when say you call a function which returns a reference (not a const value reference) to a variable or rvalue that was initialized on the stack within the function, thus it leaves scope of the function without an extended lifetime (cause it can't be) and the variable's/rvalue's lifetime existence is then destroyed leaving it dangling. The same thing can happen to pointers if they return internal stack addresses that are cleaned up on return, as example

```cpp
int& get_dangling_ref() {
	int i = 1;
	return i;
}

int const& get_extended_lifetime_ref() {
	int i = 1;
	return i;
}

int main() {
	int& dangling = get_dangling_ref(); // this is UB
	int const& extended_ref = get_extended_lifetime_ref(); // this is not UB
}
```

The reason you'd separately specify an rvalue over a const value reference is that you don't want to copy the values, you want to take the memory that currently exists and "move" it into a different place, which is cheaper then copying the whole thing, if you're dealing with a trivially constructible type on the stack though these operations are indistinguishable (the only difference is the previous stack value may now be in an invalid state) and it wouldn't matter whether it was a copy or move.

Spartan322 (approved):

PR #712, opened 2026-02-26
Reserve capacity for great & secondary powers based on defines

wvpm:
We know the maximum number of great powers and secondary powers thanks to the country defines.
So let's use those.

### DIFF
--- a/src/openvic-simulation/defines/CountryDefines.hpp
+++ b/src/openvic-simulation/defines/CountryDefines.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <cstddef>
+
 #include "openvic-simulation/dataloader/NodeTools.hpp"
 #include "openvic-simulation/types/Date.hpp"
 #include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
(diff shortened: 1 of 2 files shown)


src/openvic-simulation/defines/CountryDefines.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstddef>
```
Spartan322, line 3:
```suggestion
#include <cstddef>

```

src/openvic-simulation/defines/CountryDefines.hpp
```diff
@@ -39,6 +41,11 @@ namespace OpenVic {
 		Timespan PROPERTY(campaign_event_state_duration_modifier); // NOT USED
 		Timespan PROPERTY(campaign_duration);
 		size_t PROPERTY(secondary_power_rank, 0);
+	public:
+		constexpr size_t get_max_secondary_power_count() const {
+			return secondary_power_rank - great_power_rank;
+		}
+	private:
```
Spartan322, line 50:
```suggestion

	public:
		constexpr size_t get_max_secondary_power_count() const {
			return secondary_power_rank - great_power_rank;
		}

	private:
```

Spartan322 (approved):

PR #721, opened 2026-05-11
Minimise fixed point header

wvpm:
- Depends on #719, #720, #723 & #727

Avoids importing containers, logging and string headers.

### DIFF
--- a/src/openvic-simulation/core/Math.hpp
+++ b/src/openvic-simulation/core/Math.hpp
@@ -6,14 +6,15 @@
 #include <type_traits>
 
 #include "openvic-simulation/core/Typedefs.hpp"
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
 
 namespace OpenVic {
     template<typename T>
-    [[nodiscard]] OV_SPEED_INLINE constexpr T abs(T num);
+    [[nodiscard]] OV_SPEED_INLINE constexpr T abs(const T num);
 
 	template<typename T>
 	requires std::integral<T> || std::floating_point<T>
-	[[nodiscard]] OV_SPEED_INLINE constexpr T abs(T num) {
+	[[nodiscard]] OV_SPEED_INLINE constexpr T abs(const T num) {
 		if (std::is_constant_evaluated()) {
 			return num < 0 ? -num : num;
 		} else {
@@ -21,9 +22,14 @@ namespace OpenVic {
 		}
 	}
 
+    template<>
+	[[nodiscard]] OV_SPEED_INLINE constexpr fixed_point_t abs(const fixed_point_t  num) {
+		return num.abs();
+	}
+
     template<typename T>
-	requires (!(std::integral<T> || std::floating_point<T>))
-	[[nodiscard]] OV_SPEED_INLINE constexpr T abs(T num);
+	requires (!(std::integral<T> || std::floating_point<T> || std::is_same_v<T, fixed_point_t>))
+	[[nodiscard]] OV_SPEED_INLINE constexpr T abs(T const& num);
 
 	template<std::floating_point T>
 	OV_SPEED_INLINE constexpr int64_t round_to_int64(T num) {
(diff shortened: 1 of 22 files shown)


zaaarf:
so far so good

src/openvic-simulation/core/Math.hpp
```diff
@@ -6,24 +6,30 @@
...
 
 namespace OpenVic {
     template<typename T>
-    [[nodiscard]] OV_SPEED_INLINE constexpr T abs(T num);
+    [[nodiscard]] OV_SPEED_INLINE constexpr T abs(T const& num);
 
 	template<typename T>
 	requires std::integral<T> || std::floating_point<T>
-	[[nodiscard]] OV_SPEED_INLINE constexpr T abs(T num) {
+	[[nodiscard]] OV_SPEED_INLINE constexpr T abs(T const& num) {
```
Spartan322: There is no reason for this to be `const&`, best case scenario it has no advantage, worst case is you waste 7 bytes for something that could've been 1, value types where `sizeof(T)` is 8 or less should be value types unless they have non-trivial copy semantics.

src/openvic-simulation/core/Math.hpp
```diff
@@ -6,24 +6,30 @@
...
 		if (std::is_constant_evaluated()) {
 			return num < 0 ? -num : num;
 		} else {
 			return std::abs(num);
 		}
 	}
 
+    template<>
+	[[nodiscard]] OV_SPEED_INLINE constexpr fixed_point_t abs(fixed_point_t const& num) {
```
Spartan322: Same as [above](https://github.com/OpenVicProject/OpenVic-Simulation/pull/721/changes#r3237931042)

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -196,19 +150,19 @@ namespace OpenVic {
 			return value >> PRECISION;
 		}
 
+		#ifdef DEV_ENABLED
+			OV_SPEED_INLINE void warn_if_truncated() const;
+		#endif
```
Spartan322:
```suggestion
		OV_SPEED_INLINE void warn_if_truncated() const;
```

wvpm: Also in the .cpp then?

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
```diff
@@ -196,19 +150,19 @@ namespace OpenVic {
...
-			}
-#endif
+		OV_SPEED_INLINE explicit
+		#ifndef DEV_ENABLED
+			constexpr
+		#endif
+		operator T() const {
+			#ifdef DEV_ENABLED
+				warn_if_truncated();
+			#endif
```
Spartan322:
```suggestion
		OV_SPEED_INLINE explicit constexpr operator T() const {
#ifdef DEV_ENABLED
			if (!std::is_constant_evaluated()) {
				warn_if_truncated();
			}
#endif
```
Alternative we could make it so `warn_if_truncated` is also constexpr and in a constant evaluated context, if the truncation is found it makes a call to a function marked with `[[deprecated("Truncation will result in zero, may be a bug")]]`

src/openvic-simulation/core/Math.hpp
wvpm: Changed it to `const T`

src/openvic-simulation/types/fixed_point/FixedPoint.hpp
wvpm: The deprecated attribute requires more work and might be a good idea for a follow up.
For now I'll just use `!std::is_constant_evaluated()` from above

src/openvic-simulation/types/fixed_point/String.hpp
```diff
@@ -0,0 +1,232 @@
...
+#include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
+#include "openvic-simulation/types/StackString.hpp"
+#include "openvic-simulation/utility/Logger.hpp"
+
+namespace OpenVic::fp {
+	struct stack_string final : public StackString<25> {
+	protected:
+		using StackString::StackString;		
+	public:
+		constexpr char* first() {
```
Spartan322:
```suggestion
		constexpr char* front() {
```

Follows C++ STL types.

Spartan322 (approved):

PR #734, opened 2026-05-15
Optimise PopsAggregate by using IndexedFlatMaps

wvpm:
IndexedFlatMap is much faster than fixed_point_map_t as that uses hash-based lookups whilst the former uses indexing.
Since this code is called every gamestate update, it is a hot-path and thus performance matters here.

Note breaks godot, I'll fix that later.

### DIFF
--- a/src/openvic-simulation/InstanceManager.hpp
+++ b/src/openvic-simulation/InstanceManager.hpp
@@ -21,6 +21,7 @@
 #include "openvic-simulation/misc/SimulationClock.hpp"
 #include "openvic-simulation/politics/PoliticsInstanceManager.hpp"
 #include "openvic-simulation/population/PopDeps.hpp"
+#include "openvic-simulation/population/PopsAggregateDeps.hpp"
 #include "openvic-simulation/types/Date.hpp"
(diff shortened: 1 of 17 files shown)


src/openvic-simulation/InstanceManager.hpp
```diff
@@ -44,6 +45,7 @@ namespace OpenVic {
 		GoodInstanceManager PROPERTY_REF(good_instance_manager);
 		MarketInstance PROPERTY_REF(market_instance);
 		
+		PopsAggregateDeps pops_aggregate_deps; //before CountryInstanceDeps
```
Spartan322:
```suggestion
		PopsAggregateDeps pops_aggregate_deps; // Before CountryInstanceDeps
```
Is this a "Must be before CountryInstanceDeps", cause then the comment should probably say that.

wvpm: Yes. CountryInstanceDeps contains a reference to this. I suppose I could turn it into a copy.
I'll do that instead. It prevents this issue entirely.

Spartan322 (approved):

PR #739, opened 2026-05-17
Minimise allocations for ModifierSum

wvpm:
`CountryInstance::contribute_province_modifier_sum` is called for each province (expect 10s-100s per country). Every call to `ModifierSum::add_modifier_sum` called `reserve_more`, which is defined as:
```cpp
    constexpr void reserve_more(reservable auto& t, size_t size) {
        t.reserve(t.size() + size);
    }
```

This resulted in repeated allocations, possibly reallocating the country's modifiersum for each controlled province.
Instead we now first sum the extra size for all provinces and then resize (not reserve!) once.
We use resize instead of reserve as it uses the underlying growth algorithm that typically reserves extra room, avoiding reallocating because of 1 extra element.

Resize creates empty elements which are later replaced. The ModifierSum keeps track of the count of valid elements and uses this for its `size()`.

### DIFF
--- a/src/openvic-simulation/core/BulkInsertWrapper.hpp
+++ b/src/openvic-simulation/core/BulkInsertWrapper.hpp
@@ -1,0 +1,201 @@
+#pragma once
+
+#include <algorithm>
+#include <atomic>
+#include <cassert>
+#include <iterator>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+#include "openvic-simulation/core/Assert.hpp"
+
+namespace OpenVic {
+	// not thread safe
+	template <typename Container>
+	struct bulk_insert_wrapper {
+	public:
+		// Member types based on std::vector
+		using container_type = Container;
+		using value_type = typename container_type::value_type;
+		using allocator_type = typename container_type::allocator_type;
+		using size_type = typename container_type::size_type;
+		using difference_type = typename container_type::difference_type;
+		using reference = typename container_type::reference;
+		using const_reference = typename container_type::const_reference;
+		using pointer = typename container_type::pointer;
+		using const_pointer = typename container_type::const_pointer;
+		using iterator = typename container_type::iterator;
+		using const_iterator = typename container_type::const_iterator;
+		using reverse_iterator = typename container_type::reverse_iterator;
+		using const_reverse_iterator = typename container_type::const_reverse_iterator;
+
+		static_assert(std::is_default_constructible_v<value_type>);
+		static_assert(std::is_trivially_destructible_v<value_type>);
+
+	private:
+		Container container;
+		std::atomic<size_type> pending_extra_size {};
+
+		constexpr void flush_pending_room() {
+			if (pending_extra_size > size_type{}) {
+				size_type valid_size { size() };
+				container.resize(valid_size + pending_extra_size);
+				container.resize(valid_size);
+				pending_extra_size = size_type{};
+			}
+		}
+
+	public:
+		constexpr allocator_type get_allocator() const {
+			return container.get_allocator();
+		}
+
+		constexpr bulk_insert_wrapper() noexcept {};
+
+		// Forwarding constructor for custom allocators or initial capacities
+		template <typename... Args>
+		constexpr explicit bulk_insert_wrapper(Args&&... args)
+			: container(std::forward<Args>(args)...) {}
+
+		// thread safe
+		constexpr void make_room_for(const size_type count) noexcept {
+			pending_extra_size += count;
+		}
+
+		// Element access based on std::vector
+		constexpr reference operator[](const size_type pos) {
+			OV_HARDEN_ASSERT_ACCESS(pos, "operator[]");
+			return container[pos];
+		}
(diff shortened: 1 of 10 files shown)


schombert:
As a general note, calling reserve explicitly is typically only an optimization if you can find a one-time upper bound on the possible growth and thereby call reserve *once* instead of allowing the vector to grow naturally. However, if you call reserve repeatedly you run the risk of performing much worse than the vector's normal exponential growth. The growth factor of the vector is designed so that adding elements is amortized O(1). Repeated calls to reserve that grow the vector only by some amount that we can place an upper bound on (as presumably happens here, since the total number of possible modifiers is bounded) makes the entire process (cycles of reserve + element additions) amortized O(n). In your particular case, a little back of the napkin math will show that just allocating space for every country to have every possible modifier times the number of land provinces is within the memory budget, so you can use that as the value for a single reservation and be very, very unlikely to ever need to grow the vectors.

src/openvic-simulation/modifier/ModifierSum.hpp
```diff
@@ -117,17 +159,43 @@ namespace OpenVic {
...
+			const std::size_t new_valid_size = valid_size + modifier_sum.valid_size;
+			assert(new_valid_size <= modifiers.capacity());
+
+			// We could test that excluded_targets != ALL_TARGETS, but in practice it's always
+			// called with an explcit/hardcoded value and so won't ever exclude everything.
+			std::copy(
+				modifier_sum.modifiers.begin(),
+				modifier_sum.modifiers.begin() + modifier_sum.valid_size,
+				modifiers.begin() + valid_size
+			);
```
Spartan322: If including `<range/v3/algorithm/copy.hpp>`:

```suggestion
			ranges::copy(modifier_sum.modifiers, modifiers);
```

wvpm: That would be a bug as the size of modifiers includes the empty elements.
I left the arguments as is and replaced only `std::copy` with `ranges::copy`.

This is why I suggested creating a bulk insert wrapper around containers like modifiers.

src/openvic-simulation/utility/BulkInsertWrapper.hpp
```diff
@@ -0,0 +1,197 @@
...
+		USING_CONTAINER(difference_type);
+		USING_CONTAINER(reference);
+		USING_CONTAINER(const_reference);
+		USING_CONTAINER(pointer);
+		USING_CONTAINER(const_pointer);
+		USING_CONTAINER(iterator);
+		USING_CONTAINER(const_iterator);
+		USING_CONTAINER(reverse_iterator);
+		USING_CONTAINER(const_reverse_iterator);
+#undef USING_CONTAINER
```
Spartan322:
```suggestion
	public:
		using container_type = Container;
		using value_type = typename container_type::value_type;
		using allocator_type = typename container_type::allocator_type;
		using size_type = typename container_type::size_type;
		using difference_type = typename container_type::difference_type;
		using reference = typename container_type::reference;
		using const_reference = typename container_type::const_reference;
		using pointer = typename container_type::pointer;
		using const_pointer = typename container_type::const_pointer;
		using iterator = typename container_type::iterator;
		using const_iterator = typename container_type::const_iterator;
		using reverse_iterator = typename container_type::reverse_iterator;
		using const_reverse_iterator = typename container_type::const_reverse_iterator;
```

src/openvic-simulation/core/BulkInsertWrapper.hpp
```diff
@@ -0,0 +1,229 @@
...
+			}
+			++valid_size;
+
+		}
+		constexpr void push_back(value_type&& value) {
+			flush_pending_room();
+			if (has_invalid_element()) {
+				container[valid_size] = std::move(value);
+			} else {
+				container.push_back(std::forward(value));
```
Spartan322:
```suggestion
				container.push_back(std::move(value));
```
std::forward won't do anything here

wvpm:
I'll add unit tests and then we can merge.

src/openvic-simulation/core/BulkInsertWrapper.hpp
```diff
@@ -0,0 +1,244 @@
...
+			return container[pos];
+		}
+		constexpr const_reference operator[](const size_type pos) const {
+			OV_HARDEN_ASSERT_ACCESS(pos, "operator[]");
+			return container[pos];
+		}
+
+		constexpr reference front() {
+			OV_HARDEN_ASSERT_NONEMPTY("front");
+			return container.front();
```
Spartan322:
```suggestion
			return container[0];
```

src/openvic-simulation/core/BulkInsertWrapper.hpp
```diff
@@ -0,0 +1,244 @@
...
+		}
+		constexpr const_reference front() const {
+			OV_HARDEN_ASSERT_NONEMPTY("front");
+			return container.front();
```
Spartan322:
```suggestion
			return container[0];
```

wvpm:

PR #740, opened 2026-05-18
Accept truncations in pops smaller than 16

wvpm:
Truncations are bound to happen with fixed point math. At a certain pop size we just have to accept it to avoid spamming warnings.

### DIFF
--- a/src/openvic-simulation/population/Pop.cpp
+++ b/src/openvic-simulation/population/Pop.cpp
@@ -46,6 +46,8 @@
 
 using namespace OpenVic;
 
+constexpr pop_size_t TRUNCATION_ACCEPTABLE_BELOW_SIZE = 16;
+
 PopBase::PopBase(
 	PopType const& new_type, Culture const& new_culture, Religion const& new_religion, pop_size_t new_size,
 	fixed_point_t new_militancy, fixed_point_t new_consciousness, RebelType const* new_rebel_type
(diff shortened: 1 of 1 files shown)


src/openvic-simulation/population/Pop.cpp
```diff
@@ -295,7 +297,7 @@ void Pop::pay_income_tax(fixed_point_t& income) {
 
 template<bool IsTaxable>
 void Pop::add_artisanal_revenue(const fixed_point_t revenue) {
-	if (OV_unlikely(revenue == 0)) {
+	if (OV_unlikely(revenue == 0 && size >= TRUNCATION_ACCEPTABLE_BELOW_SIZE)) {
```
schombert: are you sure that you didn't mean `<=` ? (this applies below as well)
(and maybe `||` as well)

wvpm: No. Truncation of revenue in this case is acceptable for pops smaller than 16.
So we need to warn if size >= 16.
I updated the PR to create an inner if around the warning only.

src/openvic-simulation/population/Pop.cpp
```diff
@@ -46,6 +46,8 @@
 
 using namespace OpenVic;
 
+constexpr const pop_size_t TRUNCATION_ACCEPTABLE_BELOW_SIZE = 16;
```
Spartan322:
```suggestion
constexpr pop_size_t TRUNCATION_ACCEPTABLE_BELOW_SIZE = 16;
```
constexpr values are always const (and inline), you can't modify constexpr values cause they boil down to constant values, not variables, at compile time

Spartan322 (approved):

PR #741, opened 2026-05-19
Paralellise update gamestate & modifiers

wvpm:
Stacked on top of #739 

Note updating adjecencies inside the country is commented out.
This really should've never been placed there as recalculating it is expensive and can't be multithreaded easily.
Also the outcome rarely changes. It can only change when provinces change owner or our capital changes province.

### DIFF
--- a/src/openvic-simulation/utility/ThreadPool.cpp
+++ b/src/openvic-simulation/utility/ThreadPool.cpp
@@ -1,4 +1,5 @@
 #include "ThreadPool.hpp"
+#include "ThreadDeps.hpp"
 
 #include <array>
 #include <cstddef>
@@ -10,6 +11,7 @@
 #include "openvic-simulation/economy/GoodInstance.hpp"
 #include "openvic-simulation/economy/trading/GoodMarket.hpp"
 #include "openvic-simulation/map/ProvinceInstance.hpp"
+#include "openvic-simulation/population/PopValuesFromProvince.hpp"
 #include "openvic-simulation/types/fixed_point/FixedPoint.hpp"
 #include "openvic-simulation/types/TypedIndices.hpp"
 #include "openvic-simulation/types/TypedSpan.hpp"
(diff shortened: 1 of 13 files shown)


src/openvic-simulation/utility/ThreadPool.cpp
```diff
@@ -92,39 +85,68 @@ void ThreadPool::loop_until_cancelled(
...
 						);
 					}
-				}
-				break;
-			case work_t::COUNTRY_TICK_BEFORE_MAP:
-				for (WorkBundle& work_bundle : work_bundles) {
+					break;
+				case work_t::COUNTRY_TICK_BEFORE_MAP:
 					for (CountryInstance& country : work_bundle.countries_chunk) {
 						country.country_tick_before_map(
```
schombert, line 106: doesn't this call `manage_national_stockpile` which ultimately can call `add_buy_up_to_order` (for example), which is thread safe, but will make the order of the contents of `buy_up_to_orders` relative on the speed at which the threads in the pool finish (and their number), meaning that the order in which they are resolved in `GoodMarket::execute_orders` will vary based on timing, meaning that deterministic lockstep will not hold for multiplayer?

wvpm: The market doesn't care about ordering.

schombert: Ah, fascinating. I assumed that you were trying to replicate the v2 economy, where items are drained from the market in order (specifically in rank order) and nations later in the order simply can't get anything. Doing equal division is also what version 1.0 of PA did, and for the same reasons (it makes it possible to parallelize the economy). Since this was the one big difference between the PA economy and the V2 economy in 1.0, I imagine that your economy system will behave a lot like the 1.0 PA economy. So be warned that this creates some substantial differences in gameplay.

wvpm: The market still takes the country into account for internal market, sphere market and global market.
It's just that on the global market, there is no ranking. You get an amount proportional to the money you're willing to pay. This was implemented years ago and fixes the economy as lower ranking nations can also participate in the global market now.

schombert: I also prefer it that way. But it does make things work noticeably differently

>The market still takes the country into account for internal market, sphere market and global market.

Yes, this is also what we settled on for 1.0 PA

wvpm:

src/openvic-simulation/utility/ThreadPool.cpp
Spartan322: > The market still takes the country into account for internal market, sphere market and global market. It's just that on the global market, there is no ranking. You get an amount proportional to the money you're willing to pay. This was implemented years ago and fixes the economy as lower ranking nations can also participate in the global market now.

This shouldn't be the case, this is considered a problem, such behavior should only be implemented behind an opt-in.

PR #744, opened 2026-05-20
Move STL-like container headers to `core/stl/containers`

Spartan322:
- Partially supersedes #693 

Includes:
- CowPtr
- CowVector
- FixedVector
- RingBuffer
- StackString
- TypedSpan

Move allocator aware STL-like containers to `OpenVic::stl` namespace
Includes:
- CowPtr
- CowVector
- FixedVector
- RingBuffer

Move BasicIterator to `core/stl`
Move TslHelper.hpp to `core/stl` and rename to MutableIterator.hpp

Move memory container type aliases to `core/memory`

Add memory type alias for cow_ptr
Add memory type alias for cow_vector

Move FixedVector from namespace `OpenVic::_detail` to `OpenVic::stl`

Add stateful allocator support to FixedVector's constructors

Replace assert in TypedSpan with `OV_HARDEN_ASSERT_ACCESS`

Move STL-like container unit tests to `core/stl/containers`

PR #746, opened 2026-05-20
Add tunable_vector `std::vector` wrapper container

Spartan322:
- Depends on #744 

Add tunable_vector memory alias to `core/memory`

### DIFF
--- a/src/openvic-simulation/core/stl/containers/TunableVector.hpp
+++ b/src/openvic-simulation/core/stl/containers/TunableVector.hpp
@@ -1,0 +1,371 @@
+#pragma once
+
+#include <algorithm>
+#include <concepts>
+#include <cstddef>
+#include <initializer_list>
+#include <iterator>
+#include <memory>
+#include <ranges>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#include <range/v3/algorithm/move.hpp>
+
+#include "openvic-simulation/core/Assert.hpp"
+#include "openvic-simulation/core/Typedefs.hpp"
+
+namespace OpenVic::stl {
+	template<size_t Num, size_t Denom, size_t InitAllocSize = 0>
+	struct growth_traits {
+		static constexpr std::integral_constant<size_t, Num> numerator {};
+		static constexpr std::integral_constant<size_t, Denom> denominator {};
+		static constexpr std::integral_constant<size_t, InitAllocSize> initial_allocation_size {};
+	};
+
+	// Uses approximated golden ratio (1.617...)
+	using default_growth_traits = growth_traits<55, 34>;
+
+	namespace _detail {
+		template<typename T>
+		concept tunable_growth_trait = requires() {
+			{ T::numerator() } -> std::convertible_to<size_t>;
+			{ T::denominator() } -> std::convertible_to<size_t>;
+			{ T::initial_allocation_size() } -> std::convertible_to<size_t>;
+		};
+
+		template<typename Range, typename T>
+		concept _compatible_range =
+			std::ranges::input_range<Range> && std::convertible_to<std::ranges::range_reference_t<Range>, T>;
+	}
+
+	template<
+		typename T, _detail::tunable_growth_trait GrowthTrait = default_growth_traits, typename Allocator = std::allocator<T>>
+	class tunable_vector {
+	public:
+		using value_type = T;
+		using allocator_type = Allocator;
+		using container_type = std::vector<value_type, allocator_type>;
+		using size_type = typename container_type::size_type;
+		using difference_type = typename container_type::difference_type;
+		using reference = typename container_type::reference;
+		using const_reference = typename container_type::const_reference;
+		using pointer = typename container_type::pointer;
+		using const_pointer = typename container_type::const_pointer;
+		using iterator = typename container_type::iterator;
+		using const_iterator = typename container_type::const_iterator;
+		using reverse_iterator = typename container_type::reverse_iterator;
+		using const_reverse_iterator = typename container_type::const_reverse_iterator;
+		using growth_trait = GrowthTrait;
+
+		static_assert(growth_trait::denominator() != 0, "growth_trait's denominator cannot be 0");
+		static_assert(growth_trait::numerator() > growth_trait::denominator(), "growth_trait's divisor must be more than 1");
+
+		constexpr tunable_vector() = default;
+		explicit constexpr tunable_vector(allocator_type const& alloc) : _container(alloc) {}
+		explicit constexpr tunable_vector(size_type count, allocator_type const& alloc = allocator_type {})
+			: _container(count, alloc) {}
+		constexpr tunable_vector(size_type count, value_type const& value, allocator_type const& alloc = allocator_type {})
+			: _container(count, value, alloc) {}
+		template<typename InputIt>
+		constexpr tunable_vector(InputIt first, InputIt last, allocator_type const& alloc = allocator_type {})
+			: _container(first, last, alloc) {}
+
+		constexpr tunable_vector(container_type const& other) : _container { other } {}
+		constexpr tunable_vector(container_type&& other) : _container { std::move(other) } {}
+
+		constexpr tunable_vector(container_type const& other, allocator_type const& alloc) : _container { other, alloc } {}
+		constexpr tunable_vector(container_type&& other, allocator_type const& alloc)
+			: _container { std::move(other), alloc } {}
+
+		constexpr tunable_vector(tunable_vector const& other) : _container { other._container } {}
+		constexpr tunable_vector(tunable_vector&& other) : _container { std::move(other._container) } {}
+
+		constexpr tunable_vector(tunable_vector const& other, allocator_type const& alloc)
+			: _container { other._container, alloc } {}
+		constexpr tunable_vector(tunable_vector&& other, allocator_type const& alloc)
+			: _container { std::move(other._container), alloc } {}
+
+		template<typename OtherGrowth>
+		constexpr tunable_vector(tunable_vector<T, OtherGrowth, Allocator> const& other) : _container { other._container } {}
+		template<typename OtherGrowth>
+		constexpr tunable_vector(tunable_vector<T, OtherGrowth, Allocator>&& other)
+			: _container { std::move(other._container) } {}
+
+		template<typename OtherGrowth>
+		constexpr tunable_vector(tunable_vector<T, OtherGrowth, Allocator> const& other, allocator_type const& alloc)
+			: _container { other._container, alloc } {}
+		template<typename OtherGrowth>
+		constexpr tunable_vector(tunable_vector<T, OtherGrowth, Allocator>&& other, allocator_type const& alloc)
+			: _container { std::move(other._container), alloc } {}
+
+		constexpr tunable_vector(std::initializer_list<value_type> init, allocator_type const& alloc = allocator_type {})
+			: _container(init, alloc) {}
+
+		constexpr tunable_vector& operator=(container_type const& other) {
+			_container = other;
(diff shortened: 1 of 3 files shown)


wvpm:
> * Depends on [Move STL-like container headers to `core/stl/containers` #744](https://github.com/OpenVicProject/OpenVic-Simulation/pull/744)
> 
> 
> Add tunable_vector memory alias to `core/memory`

It shouldn't depend on that based on the code I read.
And it adds the entire type, not merely an alias.
Still nice draft.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,352 @@
...
+			{ T::initial_allocation_size } -> std::convertible_to<size_t>;
+		};
+
+		template<typename Range, typename T>
+		concept _compatible_range =
+			std::ranges::input_range<Range> && std::convertible_to<std::ranges::range_reference_t<Range>, T>;
+	}
+
+	template<
+		typename T, _detail::tunable_growth_trait GrowthTrait = growth_traits<2, 1>, typename Allocator = std::allocator<T>>
```
wvpm: Doubling creates memory fragmentation as the sum of your old memory would be 2^n -1 and you need 2^n.
The upper bound that avoids this is the golden ratio ~1.6 so 1.5 would be a better default

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,352 @@
...
+				reserve_minimum(capacity() + 1);
+			}
+			return _container.emplace_back(std::forward<Args>(args)...);
+		}
+
+		template<_detail::_compatible_range<T> RangeT>
+		constexpr void append_range(RangeT&& range) {
+			if constexpr (std::ranges::forward_range<RangeT> || std::ranges::sized_range<RangeT>) {
+				reserve_minimum(std::ranges::distance(range));
+				std::ranges::move(range, std::back_inserter(*this));
```
wvpm: Not ranges v3?
or std::uninitialized_copy + resize?

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,352 @@
...
+					tmp.emplace_back(*first);
+				}
+				std::ranges::subrange subrange(std::make_move_iterator(tmp.begin()), std::make_move_iterator(tmp.end()));
+				append_range(subrange);
+			}
+		}
+
+		constexpr void resize(size_type count) {
+			if (size() == capacity()) {
+				reserve_minimum(count);
```
wvpm: That's unexpected. resize normally does not trigger a reallocation unless you lack capacity

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,352 @@
...
+				(old_capacity != 0 && new_capacity / growth_trait::numerator < old_capacity / growth_trait::denominator)) {
+				return max;
+			}
+
+			if (old_capacity == size_type {} && new_capacity < growth_trait::initial_allocation_size) {
+				return growth_trait::initial_allocation_size;
+			}
+
+			if (value > new_capacity) {
+				return value;
```
wvpm, line 365: This seems wrong.
Say you reserve for value and fill it to or near value then add 1-3 elements, it would have to reallocate again.
It would make more sense if it:
- reserves for value + buffer size
- reserves for value x factor
- reserves for smallest factor^n that fits value (+ buffer)

Spartan322:
> > * Depends on [Move STL-like container headers to `core/stl/containers` #744](https://github.com/OpenVicProject/OpenVic-Simulation/pull/744)
> > 
> > Add tunable_vector memory alias to `core/memory`
> 
> It shouldn't depend on that based on the code I read. And it adds the entire type, not merely an alias. Still nice draft.

It doesn't rely on any code from there but I set it to depend specifically because it relies on the organization from that PR. Also the alias description is saying what its doing, it does add a memory alias for the tunable_vector type, that's that is going on in `core/memory`

src/openvic-simulation/core/stl/containers/TunableVector.hpp
Spartan322: I could write it as closer to the golden ratio, this was just what libstdc++ and libc++ does.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
Spartan322: Only specifically in the case that you reserve more than the next factor size, this is what all the standard libraries will do with resize.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
Spartan322: Can't do uninitialized_copy, would require handling allocations manually, which I could do, but I'd rather defer that for later unless requested.

wvpm: I thought there was a cross platform issue with ranges and that's why we generaly use ranges v3, is that no longer the case?

Spartan322: std::ranges doesn't generally have a cross-platform issue, (usually) the standard ranges has less features and tends to be implemented with less performance concerns.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,357 @@
...
+				for (; first != last; std::ranges::advance(first, 1)) {
+					tmp.emplace_back(*first);
+				}
+				std::ranges::subrange subrange(std::make_move_iterator(tmp.begin()), std::make_move_iterator(tmp.end()));
+				append_range(subrange);
+			}
+		}
+
+		constexpr void resize(size_type count) {
+			if (count > capacity()) {
```
wvpm: Redundant if, since reserve_minimum literally checks the same on line 308.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,357 @@
...
+	namespace _detail {
+		template<typename T>
+		concept tunable_growth_trait = requires() {
+			{ T::numerator } -> std::convertible_to<size_t>;
+			{ T::denominator } -> std::convertible_to<size_t>;
+			{ T::initial_allocation_size } -> std::convertible_to<size_t>;
+		};
+
+		template<typename Range, typename T>
+		concept _compatible_range =
```
wvpm, line 39: https://github.com/OpenVicProject/OpenVic-Simulation/pull/754/changes#diff-66b217c20a8a9c4dcc694b0c9515140cafa6b47f01a4aaa7127fba1c129ef66c

Does the same. This is a common concern.
We can leave it here for now and replace it with RangeConcepts.hpp once that's merged.
Alternatively, feel free to extract it now in case we need it for other PRs as well.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,357 @@
...
+
+		constexpr const_reverse_iterator crend() const {
+			return _container.crend();
+		}
+
+		constexpr bool empty() const {
+			return _container.empty();
+		}
+
+		constexpr size_type size() const {
```
wvpm: I recommend adding [[nodiscard]] attributes for these type of methods.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,357 @@
...
+		template<typename... Args>
+		constexpr reference emplace_back(Args&&... args) {
+			if (size() == capacity()) {
+				reserve_minimum(capacity() + 1);
+			}
+			return _container.emplace_back(std::forward<Args>(args)...);
+		}
+
+		template<_detail::_compatible_range<T> RangeT>
+		constexpr void append_range(RangeT&& range) {
```
wvpm, line 278: Whilst a useful method, it is not specific to tunable_vector.
Consider extracting it as a free function if it's used frequently.

If there are no uses, defer the implementation to future consumers.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
```diff
@@ -0,0 +1,357 @@
...
+		}
+
+		constexpr void resize(size_type count) {
+			if (count > capacity()) {
+				reserve_minimum(count);
+			}
+			_container.resize(count);
+		}
+
+		constexpr void reserve_minimum(size_type count) {
```
wvpm, line 315: For generic vector compliancy, add a `constexpr void reserve(size_type new_cap)` that picks reserve_minimum.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
wvpm: tunable_vector deviates from standard library implementation details on purpose.
This is part of the deviation and the scenario described before justifies the change.

wvpm (commented):
I'd like to see unit tests.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
Spartan322: vector implements it, just not in C++20, if we used C++23, the implementation could be more minimal and efficient, but since we don't this is the closest we can get outside of rewriting our own vector implementation

wvpm: Unless there is a real usecase I recommend against writing polyfills or backporting features.
Instead aim to upgrade to c++23.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
Spartan322: Well your first suggestion would break amortized O(1) growth completely.

As for the other two, I'm not so certain of that, changing it changes the characteristics of the growth, the point of this container is to make reserve by default naturally follow the growth formula, with the growth factor being defined by the traits. I don't want to diverge far from the vector implementation beyond the growth factor behavior because I know the characteristics of vector.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
Spartan322: Your use of bulk_inserter_wrapper was a real usecase, that aside we are not gonna be upgrading to C++23 probably for years, the compilers and standard library are not stable nor ready enough to put for production code, some consider C++20 to still be unstable, (thankful that's mostly modules) missing features have to be backported, there is no avoiding it.

src/openvic-simulation/core/stl/containers/TunableVector.hpp
wvpm: This case breaks your growth formula as you just accept whatever value is.
I think smallest factor^n that fits value is better here.

Spartan322: amortized O(1) is only a concern over time, one extra realloc isn't even a concern unless you're on an embedded device, (it be extremely hard to even figure out where it happened because of how memory pages work) calculating the smallest factor isn't a simple trivial solution either.

Spartan322:
See [`std::basic_string::resize_and_overwrite`](https://en.cppreference.com/cpp/string/basic_string/resize_and_overwrite) for a suggested addition. Recommend checking out [libstdc++'s string implementation](https://github.com/gcc-mirror/gcc/blob/ef63be5d8ae087f2805dd28d2bae9ac3eb5bd02c/libstdc%2B%2B-v3/include/bits/basic_string.tcc#L660) as an example.

PR #747, opened 2026-05-20
Main thread pick up work

wvpm:
Stacked on top of #741 
The ThreadPool spins up several worker threads and waits for them to finish.
Instead of waiting, it can join them and also work.
This reduces the thread count by 1 and saves some waiting.


wvpm:
performance gains were not enough to justify using the main thread.

PR #751, opened 2026-05-22
Fix width of strong typed integers

wvpm:
All of our indices fit within uint16_t. We want to avoid size_t as it is platform dependent and causes issues for macos in godot.


wvpm:
Shelved for now due to IndexedFlatMap issues and other PRs editing that type...

Spartan322:
I would suggest that it be best to simply be explicit with the godot bindings and not be clever with them, the only case where this turns into an issue despite that is assigning a size_to to a dictionary, which could be a justification but we should really not be relying on dictionaries being passed to godot anyway, its only being left around to avoid refactoring it twice as consequence of the potential ECS refactor.

PR #754, opened 2026-05-22
Split identifier registry

wvpm:
Stacked on top of #758

Goal: reduce headers spreading everywhere, specifically NodeTools.hpp included in IdentifierRegistry.hpp.
Plan: replace IdentifierRegistry.hpp with minimal headers for storage and free functions for searching and modifying.

### DIFF
--- a/src/openvic-simulation/types/registries/OwningRegistry_Search.hpp
+++ b/src/openvic-simulation/types/registries/OwningRegistry_Search.hpp
@@ -1,0 +1,126 @@
+#pragma once
+
+#include <algorithm>
+#include <ranges>
+#include <span>
+#include <string_view>
+
+#include "OwningRegistry.hpp"
+
+#include "openvic-simulation/core/stl/containers/TypedSpan.hpp"
+#include "openvic-simulation/core/template/Concepts.hpp"
+#include "openvic-simulation/utility/Logger.hpp"
+
+namespace OpenVic {
+	template<
+		typename ValueType,
+		typename SizeType,
+		typename Allocator
+	>
+	inline constexpr void assert_is_locked(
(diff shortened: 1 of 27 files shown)


Catylist0 (approved):
Looks fine, definitely agree with splitting out the headers and I don't see any issues with the implementation

src/openvic-simulation/types/registries/OwningRegistry_Search.hpp
```diff
@@ -0,0 +1,126 @@
...
+			spdlog::error_s("registry {} is used but not yet locked!", OpenVic::type_name<ValueType>());
+		}
+	}
+
+	template<
+		typename ValueType,
+		typename SizeType,
+		typename Allocator
+	> requires is_strongly_typed<SizeType>
+	[[nodiscard]] inline constexpr TypedSpan<SizeType, ValueType> get_values(OwningRegistry<ValueType, SizeType, Allocator>& registry) {
```
schombert, line 34: You might consider making these implicit conversion operators to avoid having to wrap their use as parameters in `get_values`, which adds noise. That said, it may be architecturally simpler to move the data into a new location to model locking rather than a flag and check, as it produces simpler code and doesn't require new types. (i.e. you would have a set of variables representing the unlocked state that the code responsible for loading references and would then move the data into new variables once loading is complete. This ensures that any loading functions called erroneously post load simply can't touch the data being used while the game runs. Functions that must be shared between the pre-load-completion and post-load logic can take a reference to the variable set itself so that they can be called on either data set.) A third option would be to create two view types to the same variable set and ensure that the view type used post load provides only const access to the data that must remain fixed. In this approach functions shared between pre-load-completion and post-load logic are accommodated by allowing the view with mutable access to be type cast to the view providing only const access.

PR #755, opened 2026-05-23
Minimise allocations for ModifierSum

wvpm:
Original PR: #739

`CountryInstance::contribute_province_modifier_sum` is called for each province (expect 10s-100s per country). Every call to `ModifierSum::add_modifier_sum` called `reserve_more`, which is defined as:
```cpp
    constexpr void reserve_more(reservable auto& t, size_t size) {
        t.reserve(t.size() + size);
    }
```

This resulted in repeated allocations, possibly reallocating the country's modifiersum for each controlled province.
Instead we now first sum the extra size for all provinces and then resize (not reserve!) once.
We use resize instead of reserve as it uses the underlying growth algorithm that typically reserves extra room, avoiding reallocating because of 1 extra element.

Resize creates empty elements which are later replaced. The ModifierSum keeps track of the count of valid elements and uses this for its `size()`.

### DIFF
--- a/src/openvic-simulation/core/stl/containers/BulkInsertWrapper.hpp
+++ b/src/openvic-simulation/core/stl/containers/BulkInsertWrapper.hpp
@@ -1,0 +1,201 @@
+#pragma once
+
+#include <algorithm>
+#include <atomic>
+#include <cassert>
+#include <iterator>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+#include "openvic-simulation/core/Assert.hpp"
+
+namespace OpenVic {
+	// not thread safe
+	template <typename Container>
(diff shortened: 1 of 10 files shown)


Spartan322:
First off squash the merge away, secondly I still firmly hold on not merging bulk_insert_wrapper as I think its completely unnecessary compared to #746 and maybe tracking inside ModifierSum if you really want to handle mult-threaded syncing.

wvpm:
Once #746 becomes available, I plan to do a follow up PR that makes bulk_insert_wrapper use tunable_vector internally.

Catylist0 (approved):
Approved as discussed in discord

Catylist0 (approved):
Fixing bugged review

PR #756, opened 2026-05-23
paralellise update gamestate

wvpm:
Original PR #741 
Stacked on top of #755 

Note updating adjecencies inside the country is commented out.
This really should've never been placed there as recalculating it is expensive and can't be multithreaded easily.
Also the outcome rarely changes. It can only change when provinces change owner or our capital changes province.

PR #757, opened 2026-05-23
Never break after operator

wvpm:
Original PR: #672
Stacked on top of #756 

Never break after operators.
Operators are the most important signal of what the code does.
They should always be in sight and never at the end of a line like an afterthought.

### DIFF
--- a/tests/src/SpyAllocator.hpp
+++ b/tests/src/SpyAllocator.hpp
@@ -1,0 +1,61 @@
+#include <cstddef>
+#include <memory>
+
+// Telemetry structure remains the same
+struct AllocationMetrics {
+    std::size_t total_allocated_bytes = 0;
+    std::size_t total_deallocated_bytes = 0;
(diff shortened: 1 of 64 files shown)


Spartan322 (changes requested):
Aside from the fact I will never permit merge commits inside an approved PR, I also won't approve anything in regards to "not breaking after operators", formatters will be required and I am not accepting any compromise on that but I also don't remotely agree with this claim regarding operators, in fact I find the styling here, aside from the lack of formatting, atrocious and would be absolutely horrified to see this continued any further in the codebase, I will never approve this.

tests/src/SpyAllocator.hpp
```diff
@@ -0,0 +1,61 @@
+#include <cstddef>
```
schombert, line 1: Some other changes, such as this file, seem to have accidentally ended up in the PR

PR #758, opened 2026-05-25
Prepare for IdentifierRegistry replacement

wvpm:
This PR is split off from #754 and prepares for it.
1. Store indices instead of pointers
2. Remove the combined unit_types registry and only use the separate registries


Catylist0 (approved):
These changes seem inoffensive unless I am missing something overt. Unless someone else pipes up with a specific concern I am happy to rubber-stamp this

PR #760, opened 2026-05-27
Add ECS

FakeByte:
ECS implementation, fully unit tested with RGO demo on how it is used.
The demo doesnt cache any data and runs multiple loops over data in multiple systems instead of combining systems to show how systems work together, its not meant to benchmark performance but usability.

Systems run in the following stages with stage 5 & 6 running 2 systems in parallel, while the other systems run alone in a stage but run multi threaded over the data:
```mermaid
flowchart TD
    classDef prov fill:#1f3a5f,stroke:#9ecbff,color:#eaf2ff;
    classDef pop  fill:#3a2f1f,stroke:#ffce9e,color:#fff3e6;
    S1["Stage 1<br/>RgoComputePopulationTotalsSystem<br/><i>over provinces</i>"]:::prov
    S2["Stage 2<br/>RgoHireSystem<br/><i>over provinces</i>"]:::prov
    S3["Stage 3<br/>RgoProduceAndPlaceOrderSystem<br/><i>over provinces</i>"]:::prov
    S4["Stage 4<br/>RgoResolveSellOrderAndOwnerShareSystem<br/><i>over provinces</i>"]:::prov
    S5a["Stage 5a<br/>RgoComputeOwnerIncomeSystem<br/><i>over provinces</i>"]:::prov
    S5b["Stage 5b<br/>RgoComputeEmployeeIncomeSystem<br/><i>over provinces</i>"]:::prov
    S6a["Stage 6a<br/>ApplyEmployeeIncomeToPopsSystem<br/><i>over pops</i>"]:::pop
    S6b["Stage 6b<br/>ApplyOwnerIncomeToPopsSystem<br/><i>over pops</i>"]:::pop
    S7["Stage 7<br/>AggregatePopIncomeSystem<br/><i>over pops</i>"]:::pop
    S1 --> S2 --> S3 --> S4
    S4 --> S5a
    S4 --> S5b
    S5b --> S6a
    S5a --> S6b
    S6a --> S7
    S6b --> S7
```

This chart shows which Systems read/write from which components and how the Systems interact with each other through components without having any dependencies on each other:
```mermaid
flowchart LR
    classDef sys  fill:#243b53,stroke:#9ecbff,color:#eaf2ff;
    classDef prov fill:#102a43,stroke:#5b8fc9,color:#d6e6ff;
    classDef pop  fill:#3a2f1f,stroke:#ffce9e,color:#fff3e6;
    classDef sing fill:#2b2b2b,stroke:#9e9e9e,color:#eee;
[...]

PR #781, opened 2026-07-04
implement minimal (simple) regiment creation

benjaminmcq:
this PR adds regiment creation
though resource allocation and recruitment times are not taken into account

TODO: 
add recruitment times
add resources needed to recruit


Spartan322:
Thank you for this PR, its appreciated to see someone outside our structure work , but be warned that as it stands we are working on adopting a ECS system as foundation to supplant our current structure which would make future contributions beyond the ECS outdated, if you'd like to know when that's complete so you can contribute to that I'd probably suggest subscribing to when #760 is merged or closed. (if its either merged or closed that will be when we have integrated an ECS system)

PR #786, opened 2026-07-10
Change LICENSE.md to MIT/Expat

Spartan322:
Update COPYRIGHT to remove GPL 3.0 license


COPYRIGHT
```diff
@@ -6,7 +6,7 @@ Source: https://github.com/OpenVicProject/OpenVic-Simulation
 Files: *
 Comment: OpenVic Simulation
 Copyright: 2023-present OpenVic Simulation contributors
-License: GPL-3.0-only
+License: Expat
```
FakeByte, line 9: Can you change it MIT License? I know the too are identical but with the license text in LICENSE.md saying MIT I think its nicer if the names we use match

LICENSE.md
```diff
@@ -1,674 +1,21 @@
...
-
-  The GNU General Public License does not permit incorporating your program
-into proprietary programs.  If your program is a subroutine library, you
-may consider it more useful to permit linking proprietary applications with
-the library.  If this is what you want to do, use the GNU Lesser General
-Public License instead of this License.  But first, please read
-<https://www.gnu.org/licenses/why-not-lgpl.html>.
+MIT License
+
+Copyright (c) 2023 OpenVicProject
```
FakeByte: This should be 2023 - present

FakeByte (changes requested):

COPYRIGHT
Spartan322: That's ill advised by the [debian COPYRIGHT format](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-specification) because calling a license MIT without a variant name is technically ambiguous and this is a technical file. (aside from it being kinda standard for the format) There is no one "MIT" which is the problem, whereas there is only one Expat which is MIT. Also calling it Expat means it matches Godot which means we don't copy the license text twice for OpenVicProject/OpenVic#711.

FakeByte: I didnt notice the change in https://github.com/OpenVicProject/OpenVic/pull/711, I think that should also be renamed to MIT License, thats how everyone calls it today, nobody these days calls it Expat license anymore, thats why there is no ambiguity.

From the debian docs you shared:
"For [SPDX](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#spdx) compatibility, versions with trailing dot-zeroes are considered to be equivalent to versions without (e.g., “2.0.0” is considered equal to “2.0” and “2”).

Currently, the full text of the licenses is only available in the [SPDX Open Source License Registry](https://spdx.org/licenses)."

This means even debian tries to stay SPDX compatible and instead of maintaining a full license list they refer to SPDX which is basically used by everyone. SPDX doesnt list an Expat license at all, it only list the MIT License with the SPDX identifier MIT. Thats what everyone uses, github auto checks the license file and will show our project as MIT License, our license itself calls itself MIT License, 

If you check the Open Source Initiative they also call it MIT License (and also refer to SPDX)
https://opensource.org/license/MIT

We should just follow modern standard and call it MIT

schombert:
Uh, if you are changing the license you probably need all prior contributors to sign off on it *or* you need to purge their contributions

FakeByte:
> Uh, if you are changing the license you probably need all prior contributors to sign off on it *or* you need to purge their contributions

We did contact everyone that committed code and requested permission prior to changing the license.

COPYRIGHT
Spartan322: This will need to be a separate PR then, Expat was already the standard used in the file and I don't want to change something that impacts the rest of it when its not related to the purpose.

Spartan322: See #790

FakeByte (approved):

PR #808, opened 2026-07-20
Replace SCons buildsystem with CMake

FakeByte:
Replaces scons with cmake

All third party submodules are removed from the repo and are now fetched with fetch content from cmake

The scripts repo is also fetched by fetch content, if run from the main repo then its only fetched a single time instead of having the submodule 4 times in the repo.


CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+        type_safe
+        Boost::int128
+        spdlog::spdlog
+        ov_plf_colony
+        ov_std_function
+        ov_xoshiro
+        ov_foonathan_memory
+)
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
benjaminmcq: i think platform detection as a whole should be replaced with something like:

```
if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
    set(IS_LINUX TRUE)
elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
    set(IS_WINDOWS TRUE)
elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
    set(IS_MACOS TRUE)
else()
    message(FATAL_ERROR "Unsupported platform: ${CMAKE_SYSTEM_NAME}.")
endif()
```
then platform specific behavior should be after...
i'm not sure how to put this as a suggestsion though so you can merge it straight up

Spartan322 (approved):

PR #810, opened 2026-07-22
Add unofficial BSD support

benjaminmcq:
- Depends on #808 

this PR makes it so that we link TBB if(UNIX AND NOT APPLE) instead of Linux for portability and semantic reasons.

it also adds BSD as a preset which is bsd-x64

unsure if i should make changes in the README because of "official support," but if it's needed it can probably be included when the README is updated or as a standalone fix.


CMakePresets.json
```diff
@@ -114,6 +128,21 @@
...
+        },
+        {
+            "name": "bsd-x64-debug",
+            "configurePreset": "bsd-x64",
+            "configuration": "Debug"
+        },
+        {
+            "name": "bsd-x64-relwithdebinfo",
+            "configurePreset": "bsd-x64",
+	    "configuration": "RelWithDebInfo"
```
Spartan322, line 144:
```suggestion
            "configuration": "RelWithDebInfo"
```

CMakeLists.txt
```diff
@@ -88,12 +86,29 @@ target_link_libraries(
...
+
+    if(NOT ICONV_INCLUDE_DIR OR NOT ICONV_LIBRARY)
+        message(FATAL_ERROR "iconv not found, try installing it")
+    endif()
+
+    include_directories(${ICONV_INCLUDE_DIR})
+    link_libraries(${ICONV_LIBRARY})
+endif()
+
+add_subdirectory(deps)
```
Spartan322, line 110: Why was deps moved down here?

CMakeLists.txt
```diff
@@ -88,12 +86,29 @@ target_link_libraries(
         ov_foonathan_memory
 )
 
-if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+# all unix non apple systems need tbb this way
```
Spartan322: Please follow the style FakeByte used for comments to keep consistency.

CMakeLists.txt
```diff
@@ -88,12 +86,29 @@ target_link_libraries(
...
+if(UNIX AND NOT APPLE)
     find_package(PkgConfig REQUIRED)
-    pkg_check_modules(TBB REQUIRED IMPORTED_TARGET tbb)
-    target_link_libraries(openvic-simulation PUBLIC PkgConfig::TBB)
+    pkg_check_modules(TBB_PC REQUIRED IMPORTED_TARGET GLOBAL tbb)
+    target_link_libraries(openvic-simulation PUBLIC PkgConfig::TBB_PC)
 endif()
 
+# if needed detect other platforms
+
```
Spartan322: Please follow the style FakeByte used for comments to keep consistency. Also remove the newline spacing.

Spartan322:
Also please change the commit message and PR to be specific about what this is doing, PRs and commit messages should describe what/why they're doing something, that's the style your commits should be doing, in this case it should say something like "Add unofficial BSD platform support" or something.

CMakeLists.txt
benjaminmcq: so the makefile it's jumping to knows about iconv

CMakeLists.txt
benjaminmcq: should be updated

CMakeLists.txt
benjaminmcq: should be updated

CMakePresets.json
benjaminmcq: i don't know why it isn't applying properly.
when i pull the changes down it doesn't have this problem either.

